PR #136, opened 2018-08-30
Config file for pep8speaks

### LINKED ISSUE #133: Integrate pep8speaks
Now that our repo is public, we could use `pep8speaks`: https://pep8speaks.com/

bourque:
This PR introduces a `.pep8speaks` config file for `pep8speaks`.  Closes #133.

### DIFF
--- a/.pep8speaks.yml
+++ b/.pep8speaks.yml
@@ -15,9 +15,8 @@ scanner:
     diff_only: False  # If True, errors caused by only the patch are shown
 
 pycodestyle:
-    max-line-length: 200  # Default is 79 in PEP8
+    max-line-length: 100  # Default is 79 in PEP8
     ignore:  # Errors and warnings to ignore
-        - W391
 
 only_mention_files_with_errors: True  # If False, a separate status comment for each file is made.
 descending_issues_order: False # If True, PEP8 issues in message will be displayed in descending order of line numbers in the file
--- a/.pep8speaks.yml
+++ b/.pep8speaks.yml
@@ -1,0 +1,23 @@
+# File : .pep8speaks.yml
+
+message:  # Customize the comment made by the bot
+    opened:  # Messages when a new PR is submitted
+        header: "Hello @{name}, Thank you for submitting the Pull Request !"
+                # The keyword {name} is converted into the author's username
+        footer: "If you have not done so, please consult the [`jwql` Style Guide](https://github.com/spacetelescope/jwql/blob/master/style_guide/style_guide.md)"
+                # The messages can be written as they would over GitHub
+    updated:  # Messages when new commits are added to the PR
+        header: "Hello @{name}, Thank you for updating !"
+        footer: ""  # Why to comment the link to the style guide everytime? :)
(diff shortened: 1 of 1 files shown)


bourque:
@SaOgaz Could you take a looksy and offer any opinions you have on this?

.pep8speaks.yml
```diff
@@ -0,0 +1,23 @@
...
+                # The keyword {name} is converted into the author's username
+        footer: "If you have not done so, please consult the [`jwql` Style Guide](https://github.com/spacetelescope/jwql/blob/master/style_guide/style_guide.md)"
+                # The messages can be written as they would over GitHub
+    updated:  # Messages when new commits are added to the PR
+        header: "Hello @{name}, Thank you for updating !"
+        footer: ""  # Why to comment the link to the style guide everytime? :)
+    no_errors: "Cheers ! There are no PEP8 issues in this Pull Request. :beers: "
+
+scanner:
+    diff_only: False  # If True, errors caused by only the patch are shown
```
SaOgaz: I'm guessing the plan is to change this to True once we've cleaned up the currently existing code?

.pep8speaks.yml
```diff
@@ -0,0 +1,23 @@
...
+
+pycodestyle:
+    max-line-length: 200  # Default is 79 in PEP8
```
SaOgaz: 200 seems a little long... I understand occasionally going over the line limit by 10 or so characters.  Or was there something already in our code you had in mind with this value?  I'd say something like 100 (120?) characters should be our max?

.pep8speaks.yml
```diff
@@ -0,0 +1,23 @@
...
+    ignore:  # Errors and warnings to ignore
+        - W391
```
SaOgaz: Can we put a comment here about what this warning is? (What is this warning? lol)

.pep8speaks.yml
```diff
@@ -0,0 +1,23 @@
...
+
+only_mention_files_with_errors: True  # If False, a separate status comment for each file is made.
+descending_issues_order: False # If True, PEP8 issues in message will be displayed in descending order of line numbers in the file
```
SaOgaz: This file is not pep 8 compliant, it's missing a blank line at the end 😆

.pep8speaks.yml
bourque: @SaOgaz Yeah I imagine we will eventually do that, but I think having it as `False` for now could help us catch things we have missed so far, as you elude to.

.pep8speaks.yml
bourque: @SaOgaz Good point.  I'll set it to something more reasonable.

.pep8speaks.yml
bourque: @SaOgaz :joy:  Well its not a python file so I think it is exempt!

.pep8speaks.yml
bourque: @SaOgaz This warning refers to not having a blank line at the end of the file, which on second thought, I decided that I didn't want to ignore afterall, so I removed it.

bourque:
@SaOgaz I think this is ready to jetty if you want to give it another review.

Issue #113, opened 2018-07-23
Explore using pyup


bourque:
`pyup` is a tool that automatically recognizes and opens pull reqeusts for dependency updates.  Perhaps this could be useful to us.

I think this will only work if our repo is public though.

https://pyup.io/

bourque:
I have configured `pyup` with this repo: https://pyup.io/account/repos/github/spacetelescope/jwql/

It looks like it is only picking up the dependencies listed under `pip:` in the `environment.yml` file though 🤔 Not sure how to get around that.

bourque:
I have reached out to the `pyup` developers, they said that this is expected behavior, and that I would need to list all of the dependencies under `pip` or have a separate `requirements` file.

PR #166, opened 2018-10-29
New requirements file for pyup

bourque:
This PR adds `.pyup-requirements`, which (hopefully) will get `pyup` to recognize all of our dependencies.  This (hopefully) solves the issues described in #113.


bourque:
@SaOgaz Could you take a look?  I don't think I will be able to tell if this works or not until these changes are in `master`, unfortunately.

PR #167, opened 2018-10-29
Simplified algorithm for determining if a log should be written to dev or not

### LINKED ISSUE #127: logging_functions should write logs from dev & test VMs to dev/
Currently `logging_functions.py` will write new log files to the `dev/` directory if the script being executed is not being executed from the admin account.  This should be extended to write to `dev/` when the script is run from personal accounts from the dev or test VMs. 

In other words, all logs should be written to `dev/` unless they are being run from the admin account on one of the production VMs.

bourque:
This PR modified `logging_functions.py` such that logs are written to `dev/` unless the user is the admin account and is in production mode.

Closes #127


bourque:
@catherine-martlin Could you review?

PR #173, opened 2018-11-01
Remove "Quicklook" reference from outward-facing content

bourque:
This PR removes references to the term "Quicklook" from the application, as we are trying to stray away from that term and promote our acronym `JWQL`.


bourque:
@laurenmarietta Since a good bit of this deals with the web app, could you review this?

PR #178, opened 2018-11-06
REST API for the web app

### LINKED ISSUE #131: Provide some RESTful services/API in web app
It would be useful to build in a RESTful API into our web application, so that users can easily get data programatically.  It could even be useful to us developers when we are in need of static data, and could cut down on the amount of code that we need to write in `views.py` and `data_containers.py`.

#### A simple example: Get the list of proposals for a given instrument:

in `urls.py`:

```python
urlpatterns = [
    ...
    path('api/<str:inst>/archive/proposals', views.proposals, name='proposals')
]
```

in `views.py`:

```python
from django.http import JsonResponse

def proposals(request, inst):
    """Return a list of proposals for the given instrument"""

    return JsonResponse({'proposal_list': [10000, 100001, 100002]})
```

What is returned when visiting `http://127.0.0.1:8000/jwql/api/FGS/archive/proposals`:

```
{"proposal_list": [10000, 100001, 100002]}
```

To get the data programmatically:

```python
import json
import urllib.request

url = urllib.request.urlopen('http://127.0.0.1:8000/jwql/api/FGS/archive/proposals')
data = json.loads(url.read().decode())
print(data)

# {'proposal_list': [10000, 100001, 100002]}
```

bourque:
This PR provides some RESTful API services for the web application.

Closes #131


bourque:
@laurenmarietta I believe this is ready to jetty.  Could you provide a review?  This one is pretty hefty so perhaps we should meet to discuss what I'm trying to do here.

jwql/website/apps/jwql/urls.py
```diff
@@ -39,18 +39,33 @@
...
-    path('<str:inst>/unlooked/', views.unlooked_images, name='unlooked'),
-    path('<str:inst>/<str:file_root>/', views.view_image, name='view_image'),
-    path('<str:inst>/<str:file>/hdr/', views.view_header, name='view_header'),
-    path('<str:inst>/archive/<str:proposal>', views.archive_thumbnails, name='archive_thumb'),
+    path('jwql/<str:inst>/', views.instrument, name='instrument'),
+    path('jwql/<str:inst>/archive/', views.archived_proposals, name='archive'),
+    path('jwql/<str:inst>/unlooked/', views.unlooked_images, name='unlooked'),
+    path('jwql/<str:inst>/<str:file_root>/', views.view_image, name='view_image'),
+    path('jwql/<str:inst>/<str:file>/hdr/', views.view_header, name='view_header'),
+    path('jwql/<str:inst>/archive/<str:proposal>', views.archive_thumbnails, name='archive_thumb'),
```
laurenmarietta: This is probably a question for another issue/PR, but do you think we should keep the "jwql" at the front of these URLs? It feels a little redundant, assuming the domain is ultimately going to be jwql.stsci.edu or something.

jwql/website/apps/jwql/urls.py
```diff
@@ -39,18 +39,33 @@
...
+    path('api/proposals/', api_views.all_proposals, name='all_proposals'),
+    path('api/<str:inst>/proposals/', api_views.instrument_proposals, name='instrument_proposals'),
+    re_path(r'^api/(?P<inst>({}))/preview_images/$'.format(api_instruments), api_views.preview_images_by_instrument, name='preview_images_by_instrument'),
```
laurenmarietta: I had to look up the magic that you used to do this, but it is SO cool!!! 
I'm thinking it might be worthwhile to implement the same kind of regex paths for the instrument variables in the non-API URLs, to make them more secure (prevent cross-site scripting)? (That could also be a separate PR if you think that's more appropriate.)

jwql/website/apps/jwql/api_views.py
```diff
@@ -0,0 +1,268 @@
...
+
+    Parameters
+    ----------
+    request : HttpRequest object
+        Incoming request from the webpage
+    instrument : str
+        The instrument of interest.  The name of the instrument must
+        mach one of the following: (``nircam``, ``NIRCam``, ``niriss``,
+        ``NIRISS``, ``nirspec``, ``NIRSpec``, ``miri``, ``MIRI``,
+        ``fgs``, ``FGS``).
```
laurenmarietta: Two small things: you specify the cases of the instruments in this docstring, but not in other views that also include an instrument argument (I think consistency would be better); and the docstring calls the argument `instrument` when it is just `inst`.

bourque:
@laurenmarietta Thanks for the comments!  I will look into those.  I also think it would be a good idea if I wrote some tests for these, which I will hopefully add to this PR if I have time before the end of the sprint.  If not, I will open a separate issue for it.

jwql/website/apps/jwql/data_containers.py
```diff
@@ -146,6 +166,53 @@ def get_filenames_by_instrument(instrument):
...
+    rootname : str
+        The rootname of interest (e.g. ``jw86600008001_02101_00007_guider2``).
+
+    Returns
+    -------
+    filenames : list
+        A list of filenames associated with the given ``rootname``.
+    """
+
+    proposal = rootname.split('_')[0].split('jw')[-1][0:5]
```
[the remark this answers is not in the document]
bourque: @laurenmarietta I believe `filename_parser` can already handle rootnames, since it does `filename = os.path.basename(filename)` and the `basename` of a `filename` just returns the `filename`:

```python
In [2]: os.path.basename('some_jwst_filename')
Out[2]: 'some_jwst_filename'
```

However, since this service can accept any shortened version of the rootname, `filename_parser` will crash if you were to just give it something like `jw86600008001_02101`.  So unless we make `filename_parser` more robust, I think we will have to explicitly scrape out he proposal number until then.

jwql/website/apps/jwql/data_containers.py
```diff
@@ -233,6 +300,124 @@ def get_image_info(file_root, rewrite):
...
+    """
+
+    service = "Mast.Jwst.Filtered.{}".format(instrument)
+    params = {"columns": "filename",
+              "filters": []}
+    response = Mast.service_request_async(service, params)
+    results = response[0].json()['data']
+
+    filenames = [result['filename'] for result in results]
+    proposals = list(set([filename[0:7].split('jw')[-1] for filename in filenames]))
```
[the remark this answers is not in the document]
bourque: @laurenmarietta We should be able to use `filename_parser` here, good call.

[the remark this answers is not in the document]
bourque: @laurenmarietta I was also able to use it in some other functions as well: `get_preview_images_by_instrument()` and `get_thumbnails_by_instrument()`

jwql/website/apps/jwql/urls.py
bourque: @laurenmarietta I agree that we should try to remove the "jwql", but yes let's tackle that in a separate PR.  It might be tricky.  I tried going down that road as part of this PR and ran into some issues.

jwql/website/apps/jwql/urls.py
bourque: @laurenmarietta Agreed!  Let's tackle that in our newly-conceived "make the URLs better" future PR.

bourque:
@laurenmarietta I think I've addressed all of your comments, let me know if there is anything else!

bourque:
@laurenmarietta By the way, I didn't have time to write any tests :( I'll have to do that in a future PR.

Issue #148, opened 2018-09-17
Add useful links to instrument-specific documentation on instrument landing pages


bourque:
Perhaps it would be convenient for our users if there were link(s) to JDocs and/or other useful instrument-specific information on each instrument landing page in the web application. Perhaps we could directly display any useful diagrams (e.g. detector layout) as well.

catherine-martlin:
https://jwst-docs.stsci.edu/display/HOM/JWST+User+Documentation+Home

PR #197, opened 2018-11-27
Added documentation button to instrument pages

catherine-martlin:
See issue #148 

I may need to change the link used for FGS - where is the specific documentation for that kept?

### DIFF
--- a/jwql/website/apps/jwql/templates/instrument.html
+++ b/jwql/website/apps/jwql/templates/instrument.html
@@ -34,6 +34,11 @@
 					{% endif %}
 				{% endfor %}
 			</div>
+			<div class="col-md-6 text-center">
+					<!-- Documentation -->
+					<h3>Documentation</h3>
+					<a class="btn btn-primary my-2" role="button" href='http://jwst-docs.stsci.edu/display/JTI/JWST+Observatory+and+Instrumentation+Documentation'>{{ inst }} Documentation</a></br>
+				</div>
 		</div>
 
 	</main>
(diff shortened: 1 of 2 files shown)


catherine-martlin:
Also, I made the link the same for all the instruments - the html links aren't a useful pattern so if we wanted to send people straight to that instrument documentation we would have to create either a set of conditions or a list somewhere for the template to know where to send people. It's a bit easier to just link all to the landing page where the information for each instrument is obvious (except FGS - were do I find that documentation?)

laurenmarietta:
Here is the JDox page for FGS - such a black sheep! https://jwst-docs.stsci.edu/display/JTI/Fine+Guidance+Sensor%2C+FGS?q=fgs

If only because FGS is different, it might be a good idea to do like you mentioned and pass the specific URLs for each instrument into the pages. I imagine the best way to go about doing this would be to add a `doc_url` key (or something like it) to the `context` dictionary that is passed out in `views.instrument()`. Something like this maybe:

```python
url_dict = {
    'fgs': 'https://jwst-docs.stsci.edu/display/JTI/Fine+Guidance+Sensor%2C+FGS?q=fgs',
    'miri': 'https://jwst-docs.stsci.edu/display/JTI/Mid-Infrared+Instrument%2C+MIRI',
    ...
}
doc_url = url_dict[inst]
context = {
    'doc_url': doc_url, 
    ...
}
```

Thoughts?

catherine-martlin:
That seems like the perfect way to do that - I'll give it a try and let you know if I have any questions.

catherine-martlin:
@laurenmarietta @bourque - I've updated the views.py and the instrument.html to allow for individual document links to the JDOCs for each instrument. Let me know if we want to change the format or add any specific pngs to pages on our web pages! I'm happy to do more additions if we can think of any good ones.

PR #176, opened 2018-11-06
Upgrade Bokeh to 1.0

bourque:
This PR contains changes required to update `bokeh` to version `1.0.1`

Issue #147, opened 2018-09-17
Build script execution status display on dashboard


bourque:
The dashboard on the web application should display the success/failure of the various `cron` jobs (e.g. `monitor_filesystem`) so that the JWQL team can easily identify scripts that are failing or have not executed when they should have.

Here is an example from what WFC3/Quicklook uses:

<img width="676" alt="screen shot 2018-09-17 at 16 05 56" src="https://user-images.githubusercontent.com/2250769/45647348-aa01be80-ba93-11e8-8ea1-2ffe314f4a8f.png">

Each row in the table corresponds to a particular `cron` job.  A user can click on the name of the monitor to display the contents of the corresponding log file.  The color indicates how recently the script executed successfully; bright green means that the script executed successfully that day, light green means the script executed successfully several days ago, red indicates that the script crashed.

bhilbert4:
So just to make sure I'm on the right track with this, I need to find the latest log file for a given monitor/script being run in the crontab (by looking into the proper subdirectory of our logging directory), check to see if it indicates success and at what time, and then this can go onto a webpage? Does that make sense, or is there a cron job log file somewhere that I should parse for this?

bourque:
@bhilbert4 That's correct.  Instead of worrying about the `crontab`, I would suggest to simply look at and parse what is in the `logs/` directory in our central storage area, as you elude to.  There is no `cron` log file.

bhilbert4:
@bourque I've been thinking about display options for this. The light green/dark green seems like it can be a little confusing given that different monitors will run at different cadences. I was thinking of trying something like:

green: latest execution was successful
yellow: based on the cadence of the log files, I expected to see a more recent run of this, but I don't
red: latest execution failed

What do you think? I don't have a good intuition for how HST QL teams used this table, and therefore what information is the most important to display.

bourque:
@bhilbert4 I like that idea a lot, and agree the various shades of green is not ideal for the reasons you mention.  I endorse that you implement it this way.

bhilbert4:
Also, my current function for parsing a log file looks for 'complete' (e.g. "filesystem monitor complete") in the final line of the log in order to determine if the job finished successfully. If 'complete' is not present in the final line, the job is flagged as failing. I'm sure this can be made a little more robust, but it does seem like it would be helpful to have some minimal logging guidelines for all monitors (e.g. at the end of the monitor, state that the monitor is complete or finished , or finished successfully).

bourque:
[...]

PR #202, opened 2018-11-29
Add cron job monitor table to web app

bhilbert4:
This PR makes several updates to the `jwql` package:

*  Introduces code to add a cron job monitor to the `jwql` package and accompanying table to the web app. 
* Updates all references to Bokeh library in html files to version 1.0.1
* Updates environment file and setup.py to use Bokeh 1.0.1
* Updates environment file to correctly install the latest version of the `jwst` package and its dependencies. (`jwst` requires a yet-to-be released version of `astropy`)

Currently, the new functionality is contained in `monitor_cron_jobs.py`, which is called within the `get_dashboard_components` method in `data_containers.py`, and therefore is run each time the dashboard page of the website is loaded.


bhilbert4:
This is still very much a work in progress. At the moment all of the plots on the Dashboard have gone missing, without any useful error statements. I'm hoping that perhaps someone will have suggestions for what may be causing this. @laurenmarietta @bourque 

I've modeled the functionality of the cron job monitor on that of the MAST monitor. Currently, the code is successfully creating and saving the new cron job table component files, but for some reason things are not being placed on the webpage correctly.

bhilbert4:
This PR is associated with #147

bhilbert4:
And once it's visible to everyone, I'm certainly open to suggestions on ways to make the table prettier.

bourque:
@bhilbert4 @laurenmarietta RE: Dashboard issues -- This behavior has also happening on the dev web server and I have yet to figure out why, so its not totally surprising that you're also coming across issues.  I wonder if its a `bokeh` version issue.  I can try to dig into it soon.

laurenmarietta:
I just commented on #176 (just saw this PR!) that I also think the dashboard issues might still be a Bokeh version problem. All the monitor plots are generated through the cron job on the VM, which still uses 0.12.5. 

Though I'm not sure why that would be happening on the dev server, too?

bhilbert4:
While testing yesterday, I manually updated the monitor plot files to reference version 1.0.1. I searched the web app for references to Bokeh and I think I got them all (probably...).

But yes, after the cron job ran last night and regenerated the files, they are back to 0.12.5. I just updated to 1.0.1 again, but we need to upgrade the version of Bokeh on the dev server, which is still running 0.12.5.

bhilbert4:
Ok @bourque, this is ready for review. 

Items to keep in mind:
* In my testing, none of the plots on the dashboard were showing up. This is related to Bokeh version
* I think I used `configure_logging` correctly for this case without a `if __name__ == '__main__'` case, but I'm not 100% positive.
* I followed examples in other parts of the code and used the boolean `production_mode` to determine where outputs should be placed.
* Should the div functions for coloring the table cells stay where they are, or move somewhere else within the app?
* The table right now isn't as pretty as I'd like it to be. Once you are able to see the table (I was using the `show` function during testing) let me know if you have any suggestions for making it nicer looking.

jwql/jwql_monitors/monitor_cron_jobs.py
```diff
@@ -0,0 +1,311 @@
...
+        log_path = os.path.join(log_path, 'dev')
+
+    # Set up a dictionary to keep track of results
+    logfile_status = {}
+
+    # Get a list of the directories under the main logging directory.
+    generator = os.walk(log_path, topdown=True)
+
+    # Loop over monitors (note that subsubdir should always be empty)
+    for subdir, subsubdir, filenames in generator:
```
bourque: @bhilbert4 I've noticed that if `log_path` is the one expected in `production_mode == True`, then this will include `dev/` logs, since `dev/` is a subdirectory.  You might want to add some code to make sure that subdirectory is not included in this `walk`.

jwql/jwql_monitors/monitor_cron_jobs.py
```diff
@@ -0,0 +1,311 @@
...
+        if len(filenames) > 0:
+            monitor_name = subdir.split('/')[-1]
+            log_file_list = [os.path.join(subdir, filename) for filename in filenames]
+
+            # Find the cadence of the monitor
+            delta_time, stdev_time = get_cadence(log_file_list)
+
+            # Identify the most recent log file
+            latest_log, latest_log_time = find_latest(log_file_list)
```
bourque: @bhilbert4 If you do `log_file_list = sorted(log_file_list)` then the `latest_log` should be as trivial as grabbing `log_file_list[-1]`.  Sorting it this way may also make `get_cadence()` a bit easier ?

jwql/jwql_monitors/monitor_cron_jobs.py
```diff
@@ -0,0 +1,311 @@
...
+    ----------
+    filename : str
+        Name of the log file to parse
+
+    Returns
+    -------
+    execution : str
+        'success' or 'failure'
+    """
+    with open(filename, 'r') as file_obj:
```
bourque: @bhilbert4 Nice use of `with open` :clap:

jwql/jwql_monitors/monitor_cron_jobs.py
```diff
@@ -0,0 +1,311 @@
...
+
+def create_table(status_dict):
+    """Create interactive Bokeh table containing the logfile status results.
+
+    Parameters
+    ----------
+    status_dict : dict
+        Nested dictionary with status results from all logfiles
+    """
+    # Rearrange the nested dictionary into a non-nested dict for the table
```
bourque: @bhilbert4 Another matter of preference thing, but another way to do this could be:

```python
filenames = [status_dict[monitor]['logname'] for monitor in status_dict]
dates = [datetime.fromtimestamp(status_dict[monitor]['latest_time']) for monitor in status_dict]
etc.
```

jwql/jwql_monitors/monitor_cron_jobs.py
```diff
@@ -0,0 +1,311 @@
...
+    dates = []
+    missings = []
+    results = []
+    for key in status_dict:
+        filenames.append(status_dict[key]['logname'])
+        dates.append(datetime.fromtimestamp(status_dict[key]['latest_time']))
+        missings.append(str(status_dict[key]['missing_file']))
+        results.append(status_dict[key]['status'])
+
+    # div to color the boxes in the status column
```
bourque: @bhilbert4 Neat!

jwql/website/apps/jwql/data_containers.py
```diff
@@ -88,7 +89,12 @@ def get_dashboard_components():
                  'filecount_type': 'Total File Counts by Type',
                  'size_type': 'Total File Sizes by Type',
                  'filecount': 'Total File Counts',
-                 'system_stats': 'System Statistics'}
+                 'system_stats': 'System Statistics',
+                 'monitor_cron_jobs': 'Cron Job Monitor',
+                 'cron_status_table': 'Cron Job Status'}
+
+    # Run the cron job monitor to produce an updated table
+    monitor_cron_jobs.status(production_mode=False)
```
bourque: @bhilbert4 Interesting, I wouldn't have thought to do this.  I think this is a good idea, as long as `monitor_cron_jobs.status()` doesn't take too long to run and significantly slows down the dashboard from loading on the web app.  I assume that the function is quite fast right now, but once we have a lot of log files this may eventually become a problem.  Either way, its an issue for sometime down the road I think. 

The alternative would be to remove this call completely and just rely on `cron` running `monitor_cron_jobs` (how meta) on a daily or perhaps hourly basis.  In that case, it would be possible for the table to be outdated, though hopefully not by much.

jwql/jwql_monitors/monitor_cron_jobs.py
```diff
@@ -0,0 +1,311 @@
+#! /usr/bin/env python
+
+"""This module monitors the status of JWQL monitors run via cron job via log files. Basic
```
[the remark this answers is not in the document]
bhilbert4: Rearranged.

jwql/jwql_monitors/monitor_cron_jobs.py
```diff
@@ -0,0 +1,311 @@
...
+        on the timestamps of the existing files. 'status' is a string that is
+        either 'success' or 'failure'.
+    """
+    # Begin logging
+    configure_logging('monitor_cron_jobs', production_mode=production_mode)
+    logging.info("Beginning cron job status monitor")
+
+    # Get main logfile path
+    settings = get_config()
+    log_path = settings['log_dir']
```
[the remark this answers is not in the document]
bhilbert4: Condensed into a single line.

jwql/jwql_monitors/monitor_cron_jobs.py
bhilbert4: Wouldn't that be slower, since it would involve 4 for loops instead of 1?

jwql/jwql_monitors/monitor_cron_jobs.py
bhilbert4: Interesting. I hadn't thought about that. Is it guaranteed that log file names will be such that `sorted` will always work? I guess the way `make_log_file` is currently set up does guarantee that.

I replaced the line in `find_latest` that was using the timestamps with your fix above.

jwql/jwql_monitors/monitor_cron_jobs.py
bourque: Probably yes, but since we are talking about a small number of items in each list (probably tens of monitors when this project is fully built) I doubt that speed will be an issue at this point in the code.  Like I said, its a matter of preference in this case (slightly faster speed vs 4 lines of code instead of 9). 🤷‍♂️

bourque: It would be interesting to time profile the use case, I should do that ...

jwql/jwql_monitors/monitor_cron_jobs.py
```diff
@@ -0,0 +1,311 @@
...
+
+Use
+---
+
+    This module can be executed as such:
+
+    ::
+
+        from jwql.jwql_monitors import monitor_cron_jobs
+        monitor_cron_jobs.status()
```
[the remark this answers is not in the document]
bhilbert4: I've added this in, so that we now have the option of running this via cron job as well as calling it within `data_containers.py`.

jwql/jwql_monitors/monitor_cron_jobs.py
bhilbert4: Done.

jwql/website/apps/jwql/data_containers.py
bhilbert4: I talked with @laurenmarietta about this a bit. It was her suggestion to put the call to `monitor_cron_jobs` in `data_containers` after I mentioned that it would be nice to have the table generated each time the dashboard is loaded. I do like the idea of the table being always up to date, but you're right that that needs to be balanced against how long it takes to run. I don't have a good intuition for how much longer it will take to run as the number of log files goes up. 

Do you want to keep it this way for now and switch to calling via cron job if the dashboard is slow later? Or should we just switch over now? Also at the moment, calling here is defaulting to `production_mode = False` while calling via command line `production_mode = True`.  I'm assuming we want to default to False for the time being?

bourque: @bhilbert4 Yes, I agree that we should try having the call to this module in `data_containers` and we can remove it later if we find it to be burdensome.  

I think that `production_mode` should be `True` in both cases actually; I don't think seeing how often the `dev` logs are being created is that useful to know in this use case.

jwql/jwql_monitors/monitor_cron_jobs.py
bhilbert4: I just did some timing tests and obtaining and sorting the timestamps of all log files is comparable to converting the time strings from the filenames to datetime objects and sorting those.

jwql/website/apps/jwql/data_containers.py
bhilbert4: Makes sense. I switched the default to True.

jwql/jwql_monitors/monitor_cron_jobs.py
bhilbert4: Some better timing tests are showing the use of timestamps is actually just slightly faster than going by the time information in the filenames.

bhilbert4:
Ok @bourque, I think I've addressed all the issues that you brought up.

jwql/jwql_monitors/monitor_cron_jobs.py
```diff
@@ -219,27 +223,27 @@ def missing_file_check(avg_time_between, uncertainty, latest_file):
...
+        Nested dictionary containing the status for all monitors. Top
+        level keys include all monitors. Within a given monitor, the
+        value is a dictionary containing 'missing_file' and 'status'
+        keys. 'missing_file' is a boolean describing whether or not
+        there is a suspected missing log file based on the timestamps
+        of the existing files. 'status' is a string that is either
+        'success' or 'failure'.
     """
     # Begin logging
-    configure_logging('monitor_cron_jobs', production_mode=production_mode)
```
bhilbert4: If you take this line out, then how is logging configured in the case where `monitor_cron_jobs` is not called from the command line?

bourque: @bhilbert4 In that case it wouldn't be configured.  But the problem is, if the script is called from the command line (as it will be in the `cron` job), then `configure_logging` would be called twice, causing two log files to be created for the same run.

I'll have to think about this.  I'm not sure that we would want a new log file created everytime the dashboard webpage is spawned.  That currently goes against our nominal 1-to-1 correspondence between `cron` runs and log files.

bourque: Alternatively, we could leave `configure_logging()` where you had it, it would still be picked up by the `cron` run.  But having it there goes against our `monitor_template` 🤷‍♂️

bhilbert4: Well, this is a monitor of monitors, so it's special, right? :) But yeah, I think the bigger issue is that we probably don't want a new log file each time the dashboard page is spawned.

bourque:
@bhilbert4 Thanks for the changes, the code looks good to go.  I just made some minor edits, namely docstring stuff, fixing a few lingering bugs, and resolving merge conflicts (see 27c072b, 982dd18, and 798ad2e).

I've tested this in production and `monitor_cron_jobs.py` runs successfully and creates the necessary outputs and log file.  Upon testing the web app locally to see if things show up on the dashboard, I can't get the table to show.  It might be related to the issues described above.  I'm going to try digging a bit more on that before I merge this.

bhilbert4:
Sounds good. Next time just tell me to shorten all my docstrings. No need for you to take your time to do it.

Do the other elements on the dashboard show up for you? In the most recent testing I did, I wasn't getting anything on the dashboard to show up. 

Maybe the Bokeh 1.0.1 to 1.0.2 update in the middle of all this is causing issues?

bourque:
BTW, that doesn't explain why you aren't seeing the other dashboard components. That i'm not sure about.  The others seem to work fine for me locally, and I'm using `bokeh` `1.0.2`

bhilbert4:
Ah I see. I interpreted `module docstring` as the singular docstring at the top. With an `s` on the end I would have thought to look elsewhere. : )

Hmm. But when I was testing locally, I owned the files and I was trying to change permissions, so I'm not sure why that didn't work. But if it works for you with the admin-owned files, then that's good enough for me. 

And I think the `try` `except` makes sense. Even if those files do always exist and are owned by `admin` in the future, under local testing another user won't be able to change permissions, right?

bourque:
I think it probably worked for you locally because at the time the component files were owned by you, but when I asked you to remove them so that the admin account could create and own them, it now breaks. 

I can go ahead and add the try/except. I think that's the best course of action. And yes, this should never cause any trouble when testing locally, since only the admin account would ever be able to change the permissions.

bourque:
Ok, after some intense digging, i've concluded that `bohek.embed.components` just doesn't work with `bokeh.models.widgets.DataTable`.   There was seemingly nothing I could do to get the components to render in the website when testing locally.

My work-around was to add a section to the `dashboard.html` template that handles special cases that use full HTML instead of components.  This required some modifications to `get_dashboard_components` function in `data_containers` as well.  See e12dcc8

@bhilbert4 Could you pull my latest changes and confirm that it works on your end as well?  And could you look over my latest commit and make sure it is coherent?

bhilbert4:
I'll have a look in detail tomorrow. One question though, based on the screen shot you showed me earlier: do we want `monitor_cron_job` to monitor itself? If it's going to be run each time the dashboard is spawned, it would seem like we probably don't want to monitor it, right?

bourque:
Good point, I agree.  Since it is being updated upon opening the dashboard and not by a `cron` job like the other monitors, there isn't really anything to monitor -- and if it breaks, we will know instantly.

This conversation is so
```
M E T A
E
T
A
```

bourque:
@bhilbert4 I can make the change to remove the cron status from the table, if you haven't already.

bourque:
@bhilbert4 nvm, I just made the change, it was really simple.

I think we are good to merge this if you agree with the changes i've made!

jwql/jwql_monitors/monitor_cron_jobs.py
```diff
@@ -0,0 +1,349 @@
...
+        TableColumn(field="result", title="Status", width=100, formatter=success_formatter),
+    ]
+    data_table = DataTable(source=source, columns=columns, width=800, height=280, index_position=None)
+
+    # Get output directory for saving the table files
+    output_dir = get_config()['outputs']
+    output_filename = 'cron_status_table'
+
+    # Save plot as components
+    script, div = components(data_table)
```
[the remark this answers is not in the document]
bourque: @bhilbert4 Right, good call. I will remove it.

jwql/website/apps/jwql/templates/base.html
```diff
@@ -19,8 +19,8 @@
 		<!-- IMPORTANT: jQuery first, then popper, then Bootstrap -->
 		<script src="{{ static('js/jquery-3.3.1.min.js') }}?{% now 'U' %}"></script>
 		<script src="{{ static('js/tinysort.min.js') }}?{% now 'U' %}"></script>
-	    <script type="text/javascript" src="http://cdn.pydata.org/bokeh/release/bokeh-0.12.5.min.js"></script>
-	    <script src="http://cdn.pydata.org/bokeh/release/bokeh-widgets-0.12.5.min.js"></script>
+	    <script type="text/javascript" src="http://cdn.pydata.org/bokeh/release/bokeh-1.0.1.min.js"></script>
+	    <script src="http://cdn.pydata.org/bokeh/release/bokeh-widgets-1.0.1.min.js"></script>
```
bhilbert4: These should be 1.0.2 now, right? Will we need to update these each time we update Bokeh?

jwql/jwql_monitors/monitor_cron_jobs.py
```diff
@@ -0,0 +1,327 @@
...
+    placed in the ``utils`` directory.
+"""
+
+from datetime import datetime
+import logging
+import numpy as np
+import os
+import time
+
+from bokeh.embed import components
```
bhilbert4: No need to import components now.

jwql/website/apps/jwql/templates/filesystem_monitor_full.html
```diff
@@ -5,9 +5,9 @@
         <meta charset="utf-8">
         <title>Bokeh Plot</title>
 
-<link rel="stylesheet" href="https://cdn.pydata.org/bokeh/release/bokeh-0.12.5.min.css" type="text/css" />
+<link rel="stylesheet" href="https://cdn.pydata.org/bokeh/release/bokeh-1.0.1.min.css" type="text/css" />
```
bhilbert4: These versions shouldn't matter (I think) because they'll be updated automatically the next time the filesystem monitor runs.

jwql/website/apps/jwql/templates/base.html
bourque: Actually I think we should keep them at `1.0.1` until we update to `1.0.2` in our `environment.yml` file.  Ideally, we will update these as we update `bokeh` across our entire project, but unless we implement something that is super cutting-edge, I don't think things will break if we leave these alone for some time.

bourque: I was using `1.0.2` in my attempts to debug the problems we were having, but i've switched it back to `1.0.1`

bourque:
@bhilbert4 Ok, I removed the unnecessary `components` import and now I believe we are using `bokeh` `1.0.1` across the board. Have I missed anything else?

Issue #220, opened 2018-12-14
Write tests for web app


laurenmarietta:
As mentioned in #208, it is increasingly important to develop tests for the increasingly complex web app!

Django actually has its [own tools for writing web app tests](https://docs.djangoproject.com/en/2.1/topics/testing/overview/) that seem to be very helpful for sending test requests and responses and such. 

And as a bonus, it looks like there is a [`pytest` plugin](https://pytest-django.readthedocs.io/en/latest/) that would let our web app tests be run with our usual `pytest` call rather than having to include a secondary Django test call. (Though hopefully a separate call wouldn't be too difficult to include into our Jenkins test case.)

bourque:
@laurenmarietta I am going to assign myself to this one as well -- I probably can muster up some time to help with this.

bourque:
Another thing to consider -- perhaps it is better to use [`pytest-django`](https://pytest-django.readthedocs.io/en/latest/index.html), which is a plugin for `pytest` for testing Django applications

bhilbert4:
Closing in favor of #969

PR #221, opened 2018-12-14
Tests for REST API

bourque:
This PR adds `test_api_views.py` and supporting documentation, which provide `pytest`s for the `api_views` module in the web app.

This perhaps belongs instead within the `tests.py` module in the `website` subpackage, but i'm not sure how that exactly works, perhaps @laurenmarietta has opinion on that.


bourque:
Note that this should not be merged until #202 is merged, so that we can isolate the `v0.16` release.  This PR is part of the `v0.17` release.

laurenmarietta:
I like these tests – very clever – and they all passed on my system! (Once I remembered to start up the server first...)

I'm not sure about where they actually belong, though. In the (hopefully near) future I intend to explore Django's built-in testing format (#220), which seems to do some magical stuff to generate the HTML requests/responses.

In the meantime, though, I'm comfortable with using these tests you've written here – at least until we establish what the web app testing should look like. But if you disagree, let me know.

Issue #186, opened 2018-11-15
Write test for basic astroquery.mast query


bourque:
Currently our Jenkins builds are failing (e.g. https://ssbjenkins.stsci.edu/blue/organizations/jenkins/STScI%2Fjwql/detail/PR-185/2/pipeline) for what seems like an issue with `monitor_mast.py` but the problem is actually that the query to the archive via `astroquery.mast` is returning no results.

Regardless of whatever the exact issue with that is, this gave me the idea that we should have a unit test that performs a simple query using `astroquery.mast` just to see if that service is returning results as expected.  This way, it will be more apparent that it is `astroquery.mast` that is the source of any issues.

hover2pi:
Addressed in #227

PR #227, opened 2018-12-17
astroquery test

hover2pi:
…h.io import in monitor_mast.py


hover2pi:
Fixes #186 and fixes `monitor_mast.py` so it imports properly from `bokeh` v1.0.

Issue #218, opened 2018-12-14
Develop software for Dark Current Monitor


bourque:
Design, develop, and integrate software for the dark current monitor.  For context not related to the software, see Jira issue: https://jira.stsci.edu/browse/JWQL-3

bhilbert4:
As I was thinking about high level design this afternoon, I was thinking about how this (and other) monitors would be run. 

Nominally the monitors will be run by cron job, right? But really what we want to do for a particular monitor is to run it each time a new, appropriate dataset shows up in the archive. So would it make more sense to run a cron job that queries the archive/filesystem, and if new files are present, the actual monitor is kicked off?

bhilbert4:
Nevermind. Assuming that the dark current monitor wraps around a file search function as described above, then it makes sense to run the dark_current_mointor by cron job on some regular schedule. That's what I get for thinking about this late on a Friday night.

bourque:
@bhilbert4 Yes, I imagine there will be a file search function of sorts in each monitor, that looks for new data.  However, I think you bring up a good point -- it would be nice to have the monitor update as soon as new data arrives.  

One thing that we could do is have a button on the web app -- "Update monitor" or something -- that spawns a new run of the given monitor.  The user could then be brought to a webpage that indicates that a new run has been executed and to check back in the future for an updated plot.  In this way, if a monitor is outdated because new data has arrived since the last cron run, at least a user can manually update things.

laurenmarietta:
I'm not sure if this is applicable here, but I know that the OTE commissioning procedures for how image analysis is done in the wavefront office includes something called a DAN (Data Availability Notification).  One machine receives a DAN when a user on another machine delivers files to a certain directory.

In an ideal world, the JWQL server would receive a DAN from MAST every time new data is available, and that could trigger monitor runs. I have no intuition for what it would take for this to be implemented, though.

JeffValenti:
JWQL should subscribe to the MAST [notification service](https://jwst-docs.stsci.edu/display/JDAT/Data+Discovery?q=notification%20service#DataDiscovery-Anticipateddata), if you want MAST to tell you when new data are available in the archive. Ask Caty Kaleida for more information. DANs are for communication between S&OC subsystems and are not available to "external" users.

laurenmarietta:
Should this issue be closed, since #229 was completed?

bourque:
Yep! I would say so!

PR #229, opened 2018-12-17
Dark Current Monitor

bhilbert4:
This PR (associated with #218) will include code to implement the dark current monitor. Currently only a skeleton exists.

This work is being done following the algorithm descriptions provided by the instrument teams on the [Dark Current Monitor](https://innerspace.stsci.edu/pages/viewpage.action?spaceKey=JWQLPROJ&title=Dark+Current+Monitor) page.


bhilbert4:
Note that #230 is addressed in the latest commit. The `filesystem_path` function has been added to `utils/utils.py`.

bhilbert4:
@bourque, I just pushed some updated dark monitor database definition text files. I propose getting rid of `nircam_dark_queries.txt` altogether and storing everything related to MAST queries in `nircam_dark_query_history.txt`

bourque:
@bhilbert4 Great, thanks!  Do you think this PR is ready for a review?

bhilbert4:
Not yet. There are a few details to add. The bigger question though is: Should I add code to interact with a database?

bourque:
@bhilbert4 Yeah, I think we should.  I can definitely help with that part if need be.  I think it would be a good starting point to just code up inserting data into the database directly from the dark monitor, then from there we can try to identify pieces that could be generalized.  I think this iterative approach would probably work out better than trying to build a monolithic generalized database interactivity module from the start.

bhilbert4:
Ok @bourque @laurenmarietta, I think this is ready for review. 

Outstanding issues:
`configure_logging` is failing for me for reasons I don't understand. I think this isn't a dark_monitor specific problem though.
I haven't done any development in terms of creating plots. I've been waiting for the Bokeh development work for that.
All of the tests are passing for me locally. Several of the tests require access to a handful of small fits files, which means these tests will fail on Travis. These tests have been flagged with `@pytest.mark.xfail` for now.
End to end tests of the `dark_monitor` are successful for NIRCam. Currently the other instruments do not have enough darks in the filesystem to trigger the monitor.

bourque:
Thanks @bhilbert4!  We will take a look soon, but it may take a while as this PR is pretty hefty!

jwql/database/database_interface.py
```diff
@@ -302,9 +306,20 @@ def get_monitor_columns(data_dict, table_name):
         column_name = column_definition[0]
         data_type = column_definition[1]
 
+        if 'array' in data_type:
+            dtype, _a, dimension = data_type.split('_')
```
bourque: @bhilbert4 This is a clever way to handle various `ARRAY` types, nice.

jwql/database/database_interface.py
```diff
@@ -365,10 +380,26 @@ class : obj
 
     return type(class_name, (base,), data_dict)
 
+
 # Create tables from ORM factory
-# NIRCamDarkQueries = monitor_orm_factory('nircam_dark_queries')
-# NIRCamDarkPixelStats = monitor_orm_factory('nircam_dark_pixel_stats')
-# NIRCamDarkDarkCurrent = monitor_orm_factory('nircam_dark_dark_current')
+definitions_dir = os.path.join(os.path.split(__file__)[0], 'monitor_table_definitions')
```
bourque: @bhilbert4 I don't see where/how this `definitions_dir` variable is used.  Could it be removed?

jwql/instrument_monitors/pipeline_tools.py
```diff
@@ -0,0 +1,249 @@
...
+    >>> from jwql.instrument_monitors import pipeline_tools
+    pipeline_steps = pipeline_tools.completed_pipeline_steps(filename)
+ """
+
+from collections import OrderedDict
+import copy
+import numpy as np
+
+from astropy.io import fits
+from jwst import datamodels
```
bourque: @bhilbert4 It looks like this `datamodels` import isn't actually used in this module.

jwql/instrument_monitors/pipeline_tools.py
```diff
@@ -0,0 +1,249 @@
...
+                         'rate': RampFitStep, 'refpix': RefPixStep, 'rscd': RSCD_Step,
+                         'saturation': SaturationStep, 'superbias': SuperBiasStep}
+
+# Readout patterns that have nframes != a power of 2. These readout patterns
+# require the group_scale pipeline step to be run.
+GROUPSCALE_READOUT_PATTERNS = ['NRSIRS2']
+
+
+def completed_pipeline_steps(filename):
+    """
```
bourque: @bhilbert4 I think `sphinx` might freak out if line 61 and 62 here are not combined, e.g.:

```python
"""Return a list of completed pipeline steps for a given file.
```

jwql/instrument_monitors/pipeline_tools.py
```diff
@@ -0,0 +1,249 @@
...
+    # Initialize using PIPE_KEYWORDS so that entries are guaranteed to
+    # be in the correct order
+    completed = OrderedDict({})
+    for key in PIPE_KEYWORDS.values():
+        completed[key] = False
+
+    header = fits.getheader(filename)
+    for key in PIPE_KEYWORDS.keys():
+        value = header.get(key)
```
bourque: @bhilbert4 Is there any reason to believe that some of these keywords may not exist in some headers?  I could see this line breaking in that case.  It might be save to wrap this around a `try`/`except` and raise an error if the value is not found.

jwql/instrument_monitors/pipeline_tools.py
```diff
@@ -0,0 +1,249 @@
...
+        if value == 'COMPLETE':
+            completed[PIPE_KEYWORDS[key]] = True
+    return completed
+
+
+def get_pipeline_steps(instrument):
+    """Get the names and order of the calwebb_detector1
```
bourque: @bhilbert4 Looks like the 72 character docstring limit got a bit messed up in this docstring.

jwql/instrument_monitors/pipeline_tools.py
```diff
@@ -0,0 +1,249 @@
...
+            # NIRCam, NISISS, FGS all do not need group scale as nframes is
+            # always a multiple of 2
+            steps.remove('group_scale')
+
+    # IPC correction currently not done for any instrument
+    steps.remove('ipc')
+
+    # Initialize using PIPE_KEYWORDS so the steps will be in the right
+    # order
+    req = OrderedDict({})
```
bourque: @bhilbert4 I think you mean "required" by `req` here.  It might be more readable to use `required_steps` or something like that as a variable name instead of `req`, otherwise it isn't entirely clear what `req` means.

jwql/instrument_monitors/pipeline_tools.py
```diff
@@ -0,0 +1,249 @@
...
+            if len(ndim_base) == 3:
+                cube = copy.deepcopy(image)
+            elif len(ndim_base) == 2:
+                cube = np.expand_dims(image, 0)
+        else:
+            ndim = image.shape
+            if ndim_base[-2:] == ndim[-2:]:
+                if len(ndim) == 2:
+                    image = np.expand_dims(image, 0)
+                elif len(ndim) > 3:
```
bourque: @bhilbert4 Thanks for putting in this check, great idea.

jwql/instrument_monitors/pipeline_tools.py
```diff
@@ -0,0 +1,249 @@
...
+    """Run the steps of calwebb_detector1 specified in the steps
+    dictionary on the input file
+
+    Parameters
+    ----------
+    input_file : str
+        File on which to run the pipeline steps
+
+    steps : collections.OrderedDict
+        Keys are the individual pipeline steps (as seen in the
```
bourque: @bhilbert4 Looks like you might be under-using the 72 character docstring limit here?

jwql/instrument_monitors/pipeline_tools.py
```diff
@@ -0,0 +1,249 @@
...
+        PIPE_KEYWORDS values above). Boolean values indicate
+        whether a step should be run or not. Steps are run in the
+        official calwebb_detector1 order.
+    """
+    first_step_to_be_run = True
+    for step_name in steps.keys():
+        if steps[step_name]:
```
bourque: @bhilbert4 Not a big deal here, but this could also be written as:

```python
for step_name in steps:
    if steps[step_name]:
```

It is slightly cleaner and doesn't require the `.keys()` method.

jwql/tests/test_data/dark_monitor/make_test_data.py
```diff
@@ -0,0 +1,25 @@
+#! /usr/bin/env python
+
+"""Create files for dark monitor testing"""
```
bourque: @bhilbert4 We should be consistent and add `Authors` and `Use` sections in the docstring here.

jwql/database/database_interface.py
bhilbert4: Ah yes. That's leftover from when I didn't realize that `get_monitor_columns` contains the assumed location of the text files.

jwql/instrument_monitors/pipeline_tools.py
bhilbert4: It was...until I decided that the extra speed associated with just using `astropy.fits` was worth it. I'll remove `datamodels`.

jwql/instrument_monitors/pipeline_tools.py
bhilbert4: `sphinx` is so picky....

jwql/instrument_monitors/pipeline_tools.py
bhilbert4: Good catch. I think a better solution is a `try/except` and if the keyword is not present, then just say that the associated pipeline step has not been completed. I don't think it's possible to have, for example, a file where dark subtraction has been done and the `S_DARK` keyword is not present.

jwql/instrument_monitors/pipeline_tools.py
bhilbert4: Fixed

jwql/instrument_monitors/pipeline_tools.py
bhilbert4: Updated to `required_steps`

jwql/instrument_monitors/pipeline_tools.py
bhilbert4: Lengthened line.

jwql/instrument_monitors/pipeline_tools.py
bhilbert4: Stripped off `keys()`.

jwql/tests/test_data/dark_monitor/make_test_data.py
bhilbert4: Oops. I didn't mean to actually include this in the repo. I used it to quickly generate some test data for a couple of the unit tests. Should I keep it in here, or remove it entirely?

bourque: @bhilbert4 If you think we won't likely use it again, probably best to remove it so there's less code to maintain.

bhilbert4: Seems unlikely. I'll remove it.

jwql/utils/instrument_properties.py
```diff
@@ -0,0 +1,228 @@
...
+        this is 10 microseconds = 1e-5 seconds
+
+    Returns:
+    --------
+    frametime : float
+        Readout time in seconds for the frame
+    """
+    instrument = instrument.lower()
+    if instrument == "nircam":
+        xs = xdim
```
bourque: @bhilbert4 Not entirely sure what `xs` and `ys` are supposed to represent here.  Perhaps these variable names could be a bit more descriptive?

jwql/utils/instrument_properties.py
```diff
@@ -0,0 +1,228 @@
+"""Collection of functions dealing with retrieving/calculating various
```
bourque: @bhilbert4 Really nice API and in-line documentation in this module!

jwql/utils/maths.py
```diff
@@ -0,0 +1,154 @@
+"""Various math-related functions used by the ``jwql`` instrument monitors.
```
bourque: @bhilbert4 I wonder if `calculations.py` would be a better name for this module.  What do you think? 

To me, something like 

```python
from jwql.utils.calculations import mean_image
```

reads a bit better than

```python
from jwql.utils.maths import mean_image
```

jwql/utils/maths.py
```diff
@@ -0,0 +1,154 @@
...
+-------
+
+    - Bryan Hilbert
+
+Use
+---
+
+    This module can be imported as such:
+
+    >>> import maths
```
bourque: @bhilbert4 I think this should be replaced with

```python
from jwql.utils import maths
```

jwql/utils/maths.py
```diff
@@ -0,0 +1,154 @@
...
+
+import numpy as np
+
+from astropy.modeling import fitting, models
+from astropy.stats import sigma_clip
+from scipy.optimize import curve_fit
+from scipy.stats import sigmaclip
+
+
+def double_gaussian(x, amp1, peak1, sigma1, amp2, peak2, sigma2):
```
bourque: @bhilbert4 Does this functionality not already exist in something like `scipy`?

jwql/utils/utils.py
```diff
@@ -33,13 +33,108 @@
...
+        else:
+            try:
+                shutil.copy2(input_file, out_dir)
+                success.append(input_new_path)
+            except:
+                failed.append(input_file)
+    return success, failed
+
+
+def download_mast_data_via_astroquery(query_results, output_dir):
```
bourque: @bhilbert4 Should we remove this until it is implemented? Otherwise, if we are to keep it, perhaps explain in the docstring what it is intended to eventually do.

jwql/utils/utils.py
```diff
@@ -33,13 +33,108 @@
...
+        input_new_path = os.path.join(out_dir, os.path.basename(input_file))
+        if os.path.isfile(input_new_path):
+            success.append(input_new_path)
+        else:
+            try:
+                shutil.copy2(input_file, out_dir)
+                success.append(input_new_path)
+            except:
+                failed.append(input_file)
+    return success, failed
```
bourque: @bhilbert4 It would be handy here to also add a call to `set_permissions` on newly-copied files.

jwql/utils/utils.py
```diff
@@ -33,13 +33,108 @@
...
 import re
+import shutil
 
 from jwql.utils import permissions
 from jwql.utils.constants import FILE_SUFFIX_TYPES, JWST_INSTRUMENT_NAMES_SHORTHAND
 
 __location__ = os.path.realpath(os.path.join(os.getcwd(), os.path.dirname(__file__)))
 
 
+def copy_files(files, out_dir):
```
bourque: @bhilbert4 It would be useful to have an accompanying `test_copy_files` for this in `test_utils.py`, but we could handle that in a separate PR if you'd rather pass on it.

jwql/utils/utils.py
```diff
@@ -33,13 +33,108 @@
...
+
+    output_dir : str
+        Directory into which the files will be downlaoded
+    """
+    # Set up the https connection
+    server = 'mast.stsci.edu'
+    conn = httplib.HTTPSConnection(server)
+
+    # Dowload the products
+    print('Number of query results: {}'.format(len(query_results)))
```
bourque: @bhilbert4 Since this function will usually be used for instrument monitors, it would be a good idea to change the `print` statements in this module to `logging.info()` statements.

jwql/utils/utils.py
```diff
@@ -33,13 +33,108 @@
...
+        # Download the data
+        uri = query_results[i]['dataURI']
+
+        print('uri is {}'.format(uri))
+
+        conn.request("GET", "/api/v0/download/file?uri="+uri)
+        resp = conn.getresponse()
+        file_content = resp.read()
+
+
```
bourque: @bhilbert4 Looks like some stray blank lines here.

jwql/utils/utils.py
```diff
@@ -275,3 +370,30 @@ def get_config():
...
+    Returns
+    -------
+    full_path : str
+        Full path to the given file, including filename
+    """
+    filesystem_base = get_config()["filesystem"]
+
+    # Subdirectory name is based on the proposal ID
+    subdir = filename[0:7]
```
bourque: @bhilbert4 I would suggest utilizing `filename_parser` for this, in case for some reason these indices change one day.

jwql/utils/utils.py
```diff
@@ -275,3 +370,30 @@ def get_config():
         settings = json.load(config_file)
 
     return settings
+
+
+def filesystem_path(filename):
```
bourque: Just making a note that this closes #230

jwql/utils/instrument_properties.py
```diff
@@ -0,0 +1,228 @@
...
+    ::
+
+        from jwql.utils import instrument_properties as inst
+        amps = inst.amplifier_info('my_files.fits')
+"""
+from copy import deepcopy
+
+from astropy.io import fits
+import numpy as np
+from jwst.datamodels import dqflags
```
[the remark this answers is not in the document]
bhilbert4: done

jwql/utils/instrument_properties.py
```diff
@@ -0,0 +1,228 @@
...
+
+from jwql.utils.constants import AMPLIFIER_BOUNDARIES, FOUR_AMP_SUBARRAYS, SUBARRAYS_ONE_OR_FOUR_AMPS
+
+
+def amplifier_info(filename, omit_reference_pixels=True):
+    """Calculate the number of amplifiers used to collect the data in a
+    given file using the array size and exposure time of a single frame
+    (This is needed because there is no header keyword specifying
```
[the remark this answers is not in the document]
bhilbert4: I'm not sure "LOL" was my reaction when I learned that... :)

jwql/utils/instrument_properties.py
```diff
@@ -0,0 +1,228 @@
...
+            # Right now this is used only for the NIRCam grism stripe
+            # subarrays, so we don't need this to be a general case that
+            # can handle any subarray orientation relative to any amp
+            # orientation
+            amp4_time = calc_frame_time(instrument, aperture, x_dim, y_dim,
+                                        4, sample_time=sample_time)
+            amp1_time = calc_frame_time(instrument, aperture, x_dim, y_dim,
+                                        1, sample_time=sample_time)
+
+            if np.isclose(amp4_time, frame_time, atol=0.001, rtol=0):
```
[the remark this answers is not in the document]
bhilbert4: It's very handy for testing...

jwql/utils/instrument_properties.py
bhilbert4: Pulled out `xs` and `ys` in favor or using `xdim` and `ydim`.

jwql/utils/maths.py
bhilbert4: Fixed

jwql/utils/maths.py
bhilbert4: Yeah, that sounds better. I'll switch it.

jwql/utils/maths.py
bhilbert4: Nothing built-in for fitting the sum of two Gaussians simultaneously that I found.

jwql/utils/utils.py
bhilbert4: done

jwql/instrument_monitors/common_monitors/dark_monitor.py
```diff
@@ -0,0 +1,913 @@
...
+    # nested list is subsequently ignored by MAST.)
+    # So query once for each dark template, and combine outputs into a
+    # single list.
+    query_results = []
+    for template_name in dark_template:
+        # Create dictionary of parameters to add
+        parameters = {"date_obs_mjd": {"min": start_date, "max": end_date},
+                      "apername": aperture, "exp_type": template_name}
+
+        query = monitor_mast.instrument_inventory(instrument, dataproduct=JWST_DATAPRODUCTS,
```
bourque: @bhilbert4 Nice use of our own software!

jwql/utils/utils.py
bhilbert4: Right now I have a call to this commented out in the dark monitor. When we talked about this in the past, we decided we wanted to rely just on the file system and not bother with getting things from MAST. Does that still make sense to you as a strategy?

bhilbert4: Actually, that commented out call is to `download_mast_data` rather than `download_mast_data_via_astroquery`. Same question applies though.

jwql/instrument_monitors/common_monitors/dark_monitor.py
```diff
@@ -0,0 +1,913 @@
...
+
+                # Get a list of all possible apertures from pysiaf
+                possible_apertures = list(Siaf(instrument).apernames)
+                possible_apertures = [ap for ap in possible_apertures if ap not in apertures_to_skip]
+
+                for aperture in possible_apertures:
+                    # Find the appropriate threshold for the number of
+                    # new files needed
+                    match = aperture == limits['Aperture']
+                    new_dark_threshold = limits['Threshold'][match]
```
bourque: @bhilbert4 This variable name confused me a bit, I was thinking it had to do with a threshold applied to pixel values.  Perhaps something like `num_files_threshold` would be better here.

jwql/utils/utils.py
bhilbert4: Done

jwql/instrument_monitors/common_monitors/dark_monitor.py
```diff
@@ -0,0 +1,913 @@
...
+                        ensure_dir_exists(os.path.join(self.output_dir, 'data'))
+                        self.data_dir = os.path.join(self.output_dir,
+                                                     'data/{}_{}'.format(self.instrument.lower(),
+                                                                         self.aperture.lower()))
+                        ensure_dir_exists(self.data_dir)
+
+                        # Copy files from filesystem
+                        dark_files, not_copied = copy_files(new_filenames, self.data_dir)
+
+                        # Short-term fix: if some of the files from the query are not
```
bourque: @bhilbert4 Yes, I think this code block should be removed.  If we encounter missing files in the filesystem, that is an issue beyond that of the dark monitor that we will have to resolve.

jwql/utils/utils.py
bhilbert4: Removed

jwql/utils/utils.py
bourque: @bhilbert4 Yes, I don't think we should be in the business of downloading and maintaining copies of data.  We should solely rely on the filesystem, and if something is missing in the filesystem, we should bring the issue up with the archive folks.

jwql/utils/utils.py
bhilbert4: Done

jwql/utils/utils.py
bhilbert4: Done. Test is passing.

jwql/instrument_monitors/common_monitors/dark_monitor.py
```diff
@@ -0,0 +1,913 @@
...
+
+    ::
+
+        python dark_monitor.py
+"""
+
+from copy import copy, deepcopy
+import datetime
+import logging
+import numpy as np
```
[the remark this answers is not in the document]
bhilbert4: Moved

jwql/instrument_monitors/common_monitors/dark_monitor.py
```diff
@@ -0,0 +1,913 @@
...
+        if not count:
+            filename = None
+        else:
+            filename = query.all()[0].baseline_file
+            print('Baseline filename: {}'.format(filename))
+        return filename
+
+    def identify_tables(self):
+        """
+        Determine which database tables to use for a run of the dark
```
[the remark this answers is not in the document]
bhilbert4: done

jwql/instrument_monitors/common_monitors/dark_monitor.py
```diff
@@ -0,0 +1,913 @@
...
+            where the dark monitor was run.
+        """
+        subq = session.query(self.query_table.aperture,
+                             func.max(self.query_table.end_time_mjd).label('maxdate')
+                             ).group_by(self.query_table.aperture).subquery('t2')
+
+        # Note that "self.query_table.run_monitor == True" below is
+        # intentional. Switching = to "is" results in an error in
+        # the query.
+        query = session.query(self.query_table).join(
```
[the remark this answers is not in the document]
bhilbert4: This was the only way I found that you could pull out a subset of rows from the table and return the most recent of those.

jwql/instrument_monitors/common_monitors/dark_monitor.py
```diff
@@ -0,0 +1,913 @@
...
+        the most recent query for the given ``aperture_name`` where the
+        dark monitor was executed.
+
+        Returns
+        -------
+        query_result : float
+            Date (in MJD) of the ending range of the previous MAST query
+            where the dark monitor was run.
+        """
+        subq = session.query(self.query_table.aperture,
```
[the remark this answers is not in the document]
bhilbert4: done

jwql/instrument_monitors/common_monitors/dark_monitor.py
bhilbert4: Recycling!

jwql/instrument_monitors/common_monitors/dark_monitor.py
bhilbert4: Changed to `file_count_threshold`

jwql/instrument_monitors/common_monitors/dark_monitor.py
bhilbert4: Removed. Is there a reason to keep the `download_mast_data` function? Seems like it could potentially be useful to someone...maybe.

jwql/instrument_monitors/common_monitors/dark_monitor.py
```diff
@@ -0,0 +1,913 @@
...
+                      "apername": aperture, "exp_type": template_name}
+
+        query = monitor_mast.instrument_inventory(instrument, dataproduct=JWST_DATAPRODUCTS,
+                                                  add_filters=parameters, return_data=True, caom=False)
+        if len(query['data']) > 0:
+            query_results.extend(query['data'])
+    return query_results
+
+
+class Dark():
```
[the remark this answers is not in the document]
bhilbert4: Added class docstring

jwql/instrument_monitors/common_monitors/dark_monitor.py
bourque: @bhilbert4 Yeah we can keep that around for now.  It has a nice place in the `utils` module.

jwql/instrument_monitors/common_monitors/dark_monitor.py
```diff
@@ -0,0 +1,955 @@
...
+            completed_steps = pipeline_tools.completed_pipeline_steps(filename)
+            steps_to_run = pipeline_tools.steps_to_run(required_steps, completed_steps)
+
+            #logging.info('Working on file: {}'.format(filename))
+            #logging.info('Required pipeline steps: {}'.format(required_steps))
+            #logging.info('Completed pipeline steps: {}'.format(completed_steps))
+            #logging.info('Pipeline steps that remain to be run: {}'.format(steps_to_run))
+
+            # Run any remaining required pipeline steps
+            if any(steps_to_run.values()) is False:
```
bourque: @bhilbert4 Nice use of `any()`, not something you see everyday ...

jwql/instrument_monitors/common_monitors/dark_monitor.py
```diff
@@ -0,0 +1,955 @@
...
+
+        Returns
+        -------
+        output_filename : str
+            Name of fits file to save mean and stdev images within
+        """
+        output_filename = '{}_{}_{}_to_{}_mean_slope_image.fits'.format(self.instrument.lower(),
+                                                                        self.aperture.lower(),
+                                                                        self.query_start, self.query_end)
+        mean_slope_dir = os.path.join(get_config()['outputs'], 'monitor_darks/mean_slope_images/')
```
bourque: @bhilbert4 It would be slightly safer/more consistent to use:

```python
mean_slope_dir = os.path.join(get_config()['ouputs'], 'monitor_darks', 'mean_slope_images') 
```

jwql/instrument_monitors/common_monitors/dark_monitor.py
```diff
@@ -0,0 +1,955 @@
...
+        # Calculate a mean slope image from the inputs
+        slope_image, stdev_image = calculations.mean_image(slope_image_stack, sigma_threshold=3)
+        mean_slope_file = self.save_mean_slope_image(slope_image, stdev_image, slope_files)
+        #logging.info('Sigma-clipped mean of the slope images saved to: {}'.format(mean_slope_file))
+
+        # Search for new hot/dead/noisy pixels----------------------------
+        # Read in baseline mean slope image and stdev image
+        # The baseline image is used to look for hot/dead/noisy pixels,
+        # but not for comparing mean dark rates. Therefore, updates to
+        # the baseline cane be minimal.
```
bourque: @bhilbert4 Small typo here

jwql/instrument_monitors/common_monitors/dark_monitor.py
```diff
@@ -0,0 +1,955 @@
...
+                self.pixel_table.entry_date == subq.c.maxdate
+            )
+        )
+
+        count = query.count()
+        if not count:
+            filename = None
+        else:
+            filename = query.all()[0].baseline_file
+            print('Baseline filename: {}'.format(filename))
```
bourque: @bhilbert4 I'd suggest switching this to a `logging` statement and moving it back to the `run` method.

jwql/instrument_monitors/common_monitors/dark_monitor.py
```diff
@@ -0,0 +1,955 @@
...
+            2D stdev image
+        """
+        try:
+            with fits.open(filename) as hdu:
+                mean_image = hdu['MEAN'].data
+                stdev_image = hdu['STDEV'].data
+            return mean_image, stdev_image
+        except (FileNotFoundError, KeyError) as e:
+            #logging.warning('Trying to read {}: {}'.format(filename, e))
+            print(e)
```
bourque: @bhilbert4 This line can be removed

jwql/instrument_monitors/common_monitors/dark_monitor.py
```diff
@@ -0,0 +1,955 @@
...
+
+        db_entries = session.query(self.pixel_table) \
+            .filter(self.pixel_table.type == pixel_type) \
+            .filter(self.pixel_table.detector == self.detector) \
+            .all()
+
+        already_found = []
+        if len(db_entries) != 0:
+            for _row in db_entries:
+                print('row:', _row)
```
bourque: @bhilbert4 Since this will nominally be run in a `cron` job, we could probably remove these `print` statements

jwql/tests/test_utils.py
```diff
@@ -248,17 +248,19 @@
 def test_copy_files():
     """Test that files are copied successfully
     """
-    # Get an example file to be copied
-    data_dir = os.path.join(os.path.dirname(__file__), 'test_data/dark_monitor/')
-    file_to_copy = os.path.join(data_dir, 'test_image_ff.fits')
+    # Create an example file to be copied
```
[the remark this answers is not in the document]
bhilbert4: There is a newer version of this that doesn't rely on a fits file.

jwql/instrument_monitors/common_monitors/dark_monitor.py
```diff
@@ -0,0 +1,955 @@
...
+            self.detector = header['DETECTOR']
+            self.x0 = header['SUBSTRT1']
+            self.y0 = header['SUBSTRT2']
+            self.xsize = header['SUBSIZE1']
+            self.ysize = header['SUBSIZE2']
+            self.sample_time = header['TSAMPLE']
+            self.frame_time = header['TFRAME']
+            self.read_pattern = header['READPATT']
+        except KeyError as e:
+            print(e)
```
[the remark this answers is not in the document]
bhilbert4: Done

jwql/instrument_monitors/common_monitors/dark_monitor.py
bhilbert4: Updated

jwql/instrument_monitors/common_monitors/dark_monitor.py
bhilbert4: Fixed

jwql/instrument_monitors/common_monitors/dark_monitor.py
bhilbert4: Done

jwql/instrument_monitors/common_monitors/dark_monitor.py
bhilbert4: Removed

jwql/instrument_monitors/common_monitors/dark_monitor.py
bhilbert4: Done.

bhilbert4:
I'l also uncomment all the `logging` lines now, in preparation for testing.

jwql/instrument_monitors/common_monitors/dark_monitor.py
```diff
@@ -0,0 +1,988 @@
...
+        try:
+            with fits.open(filename) as hdu:
+                mean_image = hdu['MEAN'].data
+                stdev_image = hdu['STDEV'].data
+            return mean_image, stdev_image
+        except (FileNotFoundError, KeyError) as e:
+            logging.warning('Trying to read {}: {}'.format(filename, e))
+
+    @log_fail
+    @log_info
```
laurenmarietta: It might be worthwhile to add these decorators to the `__init__()` function, too, since then any log output there would also be send to the terminal. (Or at least that's how I think these decorators work...)

jwql/instrument_monitors/common_monitors/dark_monitor.py
```diff
@@ -0,0 +1,988 @@
...
+A mean slope image as well as a standard deviation slope image is
+created by sigma-clipping on a pixel by pixel basis. The mean and
+standard deviation images are saved to a fits file, the name of which
+is entered into the ``<Instrument>DarkCurrent`` database table.
+
+The mean slope image is then normalized by an existing baseline slope
+image. New hot pixels are identified as those with normalized signal
+rates above a ``hot_threshold`` value. Similarly, pixels with
+normalized signal rates below a ``dead_threshold`` are flagged as new
+dead pixels.
```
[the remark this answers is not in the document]
bhilbert4: I would say yes, unless the instrument teams want to do some more complex analysis when looking for bad pixels (e.g. line/function fitting to the ramp files). But in this case we have to make the slope images for the dark monitor anyway, so rather than make them again in a bad pixel monitor, why not work on them now.

jwql/utils/calculations.py
```diff
@@ -0,0 +1,166 @@
+"""Various math-related functions used by the ``jwql`` instrument
+monitors.
```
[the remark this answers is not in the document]
bhilbert4: I originally called it `maths` (so as not to be confused with the built-in `math`). @bourque suggested `calculations`. When writing it, I envisioned it as a module that can contain general mathematical functions that can be used by a variety of monitors. I figure lots of monitors will need to calculate means/stdevs across apertures/detectors, and fit various functions to data.

bhilbert4:
No query history is ok if you're running on new database tables, or checking for instrument/aperture combinations that haven't been called before. I destroyed and remade the tables a bunch of times while testing. I don't remember what state they were in when I finished.

jwql/instrument_monitors/common_monitors/dark_monitor.py
```diff
@@ -0,0 +1,988 @@
...
+from pysiaf import Siaf
+from sqlalchemy import func, Table
+from sqlalchemy.sql.expression import and_
+
+from jwql.database.database_interface import base, session
+from jwql.database.database_interface import NIRCamDarkQueries, NIRCamDarkPixelStats, NIRCamDarkDarkCurrent
+from jwql.database.database_interface import NIRISSDarkQueries, NIRISSDarkPixelStats, NIRISSDarkDarkCurrent
+from jwql.database.database_interface import MIRIDarkQueries, MIRIDarkPixelStats, MIRIDarkDarkCurrent
+from jwql.database.database_interface import NIRSpecDarkQueries, NIRSpecDarkPixelStats, NIRSpecDarkDarkCurrent
+from jwql.database.database_interface import FGSDarkQueries, FGSDarkPixelStats, FGSDarkDarkCurrent
```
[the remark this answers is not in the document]
bhilbert4: I think that's ok.  See the `identify_tables` function to see where the tables are used. PyCharm probably isn't picking up on them being used since they are wrapped in the `literal_eval` function. It could be switched to call `monitor_orm_factory` on the three text files for the given instrument. I guess the downside of that is that you would then be recreating the ORM many times, since the dark monitor loops over instrument/aperture combinations.

jwql/instrument_monitors/common_monitors/dark_monitor.py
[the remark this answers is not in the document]
laurenmarietta: That's great!

jwql/instrument_monitors/common_monitors/dark_monitor.py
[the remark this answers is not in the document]
laurenmarietta: Ah, I hadn't thought about the fact that the ORM would be re-made every time. That's a good enough reason for me to keep it the way it is now.

(I haven't seen that stuff with `eval` before, very neat!)

jwql/utils/calculations.py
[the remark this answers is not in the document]
laurenmarietta: Gotcha, if it will be shared by many monitors in the future then a general name is definitely better.

bourque:
FYI, I am in the process of testing this on the test server, and plan to push a few more changes.  I'm busy with meetings so may not get to it for the next day or two.

jwql/utils/instrument_properties.py
```diff
@@ -0,0 +1,229 @@
...
+        colpad = 6
+        if 'acq1' in aperture.lower():
+            colpad = 12
+        rowpad = 1
+        if amps == 4:
+            fullpad = 1
+        else:
+            fullpad = 0
+
+    return ((1.0 * xdim / amps + colpad) * (ydim + rowpad) + fullpad) * sample_time
```
laurenmarietta: From what I can tell, this function only gets run if the aperture is in the short `SUBARRAYS_ONE_OR_FOUR_AMPS` list, but I did notice that if it were ever run for MIRI or NIRSpec it would throw an error. Might be worth it to add a case that throws a `NotImplementedError` or something here for those instruments.

jwql/tests/test_pipeline_tools.py
```diff
@@ -0,0 +1,181 @@
...
+    miri_req_steps = pipeline_tools.get_pipeline_steps('miri')
+    miri_steps = ['dq_init', 'saturation', 'firstframe', 'lastframe',
+                  'linearity', 'rscd', 'dark_current', 'refpix', 'jump', 'rate']
+    not_required = ['group_scale', 'ipc', 'superbias', 'persistence']
+    miri_dict = OrderedDict({})
+    for step in miri_steps:
+        miri_dict[step] = True
+    for step in not_required:
+        miri_dict[step] = False
+    assert miri_req_steps == miri_dict
```
laurenmarietta: I see that FGS and NIRISS aren't included here - is the `get_pipeline_steps` function accurate for them?

jwql/utils/constants.py
```diff
@@ -92,3 +92,11 @@
                               'NRCB1', 'NRCB2', 'NRCB3', 'NRCB4']
 
 NIRCAM_LONGWAVE_DETECTORS = ['NRCA5', 'NRCB5']
+
+AMPLIFIER_BOUNDARIES = {'nircam': {'1': [(0, 0), (512, 2048)], '2': [(512, 0), (1024, 2048)],
+                                   '3': [(1024, 0), (1536, 2048)], '4': [(1536, 0), (2048, 2048)]}
+                        }
```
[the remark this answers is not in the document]
bhilbert4: Oops. Yeah, I forgot to do that. I was a little nervous because NIRISS' detector is rotated 90 degrees compared to NIRCam, and I wasn't sure how the code would react to 4 horizontal blocks on the detector rather than 4 vertical blocks.

jwql/utils/constants.py
```diff
@@ -92,3 +92,11 @@
...
+
+FOUR_AMP_SUBARRAYS = ['WFSS128R', 'WFSS64R']
+
+SUBARRAYS_ONE_OR_FOUR_AMPS = ['SUBGRISMSTRIPE64', 'SUBGRISMSTRIPE128', 'SUBGRISMSTRIPE256']
```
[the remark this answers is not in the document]
bhilbert4: These should be for any instrument's apertures (although I think NIRCam may be the only instrument that has apertures which can be read out with either 1 or 4 amps). NIRISS may have subarrays that are read out with 4 amps, but I'm not sure. I can check with Kevin Volk.

jwql/tests/test_pipeline_tools.py
bhilbert4: We should add tests for them as well. The results for both should be identical to NIRCam.

bourque:
One thing that is not ideal with this is the logging that happens when the JWST pipeline is run.  The `jwst` package also sets a root logger, and so the logging for the dark monitor inherently also receives any logging that happens through `jwst`.  This is a known issue, and the `jwst` developers plan to refactor this behavior eventually (https://github.com/spacetelescope/jwst/issues/2390).

I attempted a few methods to supress the logging around the calls to `jwst` (namely `logging.propagate = False` and `logging.disabled = True`, but nothing seemed to work.  

For now, I guess we will just have to live with our logs for the dark monitor having various pipeline information in it.

jwql/utils/constants.py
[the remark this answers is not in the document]
bhilbert4: I've asked Kevin Volk to confirm the details of NIRISS' amplifier readouts.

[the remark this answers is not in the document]
bhilbert4: I've also asked the same of the MIRI/NIRSpec teams

jwql/utils/instrument_properties.py
bhilbert4: I've asked the MIRI and NIRSpec teams for functions to calculate frame times.

jwql/tests/test_pipeline_tools.py
bourque: I can take care of this

bourque:
I have tested this on the test server, and it has completed successfully.  Spot checks of the database contents look reasonable.  It took about 8 hours to complete (is this expected @bhilbert4?).

One more change I would like to make:  make sure that the paths that are being stored in the database tables only contain basenames, and not full directory names.

bourque:
@bhilbert4 @laurenmarietta Here is a diff showing my changes from the last few days, could one or both of you take a quick look (:eyes:) and make sure i'm not overlooking anything? 

https://github.com/spacetelescope/jwql/pull/229/files/8e889d4e998f066a4bc528ab0e4bd89dd509737a..39486a3c7092a2243ba2d46ee8849a7474e38605

bhilbert4:
Looks good to me. The only part I'm curious about is the change from looking for the 'PIXELDQ' extension to the 'DQ' extension in `amplifier _info`. For ramp files, there is usually a `PIXELDQ` extension anf a `GROUPDQ` extension. But maybe once the ramp-fitting step is done those are merged into a single `DQ` extension?

Also, I have a frame time calculation function for MIRI. I need a little time to understand it since it is very different than the NIRCam/NIRISS/FGS functions. Do you want to hold off on adding other instrument-specific details until a future PR, or do you want them in this PR? Also, from what Misty has told me about MIRI amplifiers, there are some edits that should happen in some of the amplifier-related code to support MIRI.

bourque:
Thanks @bhilbert4, yes lets hold off on non-NIRCam edits/improvements for a future PR.

RE: `PIXELDQ`:  Hmm ok, I made that switch because the code was crashing, saying that there was no such thing as a `PIXELDQ` extension.  I assumed it was some sort of typo and switched it to `DQ`, which the code seems to be running fine on.  

Is this a line of code that we need to make more robust or add conditions to perhaps?

bhilbert4:
Running through some of the files in the filesystem, it does look like the extension name changes to `DQ` after ramp-fitting is done. Makes sense since there are no longer groups at that point. The dark monitor is designed to run on slope images, meaning ramp-fitting should always have been run on the inputs. So maybe we should just add a check that raises an error if `DQ` extension is not present?

bourque:
@bhilbert4 Sounds like a good compromise.  I'll add that in.

bourque:
I think this is all set to go, assuming my final test run (currently executing) completes successfully.

bourque:
Testing completed successfully! I noticed a possible few lingering issues with database entries, but considering this is in working order, I will merge this and describe the possible issues in #374

bhilbert4:
Huzzah!

bourque:
Awesome work @bhilbert4! :cake:

PR #242, opened 2019-01-02
[WIP] Use ajax for loading large data

### LINKED ISSUE #151: Add loading widget
It would be reassuring for users of the web application so see some sort of loading screen/widget while things are loading, instead of wondering if the application has frozen.

bourque:
This PR adds and changes several components of the web app to use asynchronous JavaScript and XML (ajax) to dynamically load data into the webpages.  It also adds a loading animation.  Currently a work in progress; more details to follow.

Closes #151

### DIFF
--- a/jwql/website/apps/jwql/static/css/loader-magnify.css
+++ b/jwql/website/apps/jwql/static/css/loader-magnify.css
@@ -1,0 +1,74 @@
+@keyframes lds-magnify {
+  0% {
+    -webkit-transform: translate(2px, 2px);
+    transform: translate(2px, 2px);
+  }
+  33.33% {
+    -webkit-transform: translate(102px, 2px);
+    transform: translate(102px, 2px);
+  }
+  66.66% {
+    -webkit-transform: translate(42px, 102px);
+    transform: translate(42px, 102px);
+  }
+  100% {
+    -webkit-transform: translate(2px, 2px);
+    transform: translate(2px, 2px);
+  }
+}
+@-webkit-keyframes lds-magnify {
+  0% {
+    -webkit-transform: translate(2px, 2px);
+    transform: translate(2px, 2px);
+  }
+  33.33% {
+    -webkit-transform: translate(102px, 2px);
+    transform: translate(102px, 2px);
+  }
+  66.66% {
+    -webkit-transform: translate(42px, 102px);
+    transform: translate(42px, 102px);
+  }
+  100% {
+    -webkit-transform: translate(2px, 2px);
+    transform: translate(2px, 2px);
+  }
+}
+.lds-magnify {
+  position: relative;
+}
+.lds-magnify > div {
+  -webkit-transform: scale(0.4);
+  transform: scale(0.4);
+  -webkit-transform-origin: 100px 100px;
+  transform-origin: 100px 100px;
+}
+.lds-magnify > div > div {
+  -webkit-animation: lds-magnify 1.5s linear infinite;
+  animation: lds-magnify 1.5s linear infinite;
+  position: absolute;
+}
+.lds-magnify > div > div div:nth-child(1) {
+  width: 96px;
+  height: 96px;
+  border-radius: 50%;
+  border: 12px solid #000000;
+  background: #ffd04f;
+}
+.lds-magnify > div > div div:nth-child(2) {
+  width: 17px;
+  height: 51px;
+  -webkit-transform: rotate(-45deg);
+  transform: rotate(-45deg);
+  background: #000000;
+  border-radius: 0 0 8px 8px;
+  position: absolute;
+  top: 68px;
+  left: 85px;
+}
+.lds-magnify {
+  width: 200px !important;
+  height: 200px !important;
+  -webkit-transform: translate(-100px, -100px) scale(1) translate(100px, 100px);
+  transform: translate(-100px, -100px) scale(1) translate(100px, 100px);
+}
--- a/jwql/website/apps/jwql/templates/archive.html
+++ b/jwql/website/apps/jwql/templates/archive.html
@@ -6,11 +6,52 @@
 
 	<!-- Custom styles and scripts for this template -->
 	<link rel="stylesheet" href="https://use.fontawesome.com/releases/v5.0.13/css/all.css" integrity="sha384-DNOHZ68U8hZfKXOrtjWvjxusGo9WQnrNx2sqG0tfsghAvtVlRW3tvkXWZh58N9jp" crossorigin="anonymous">
-	<!-- <link href="{{staticURL}}css/loader.css?{% now 'U' %}" rel="stylesheet"> -->
+
+	<script type="text/javascript">
+	function make_proposal_array() {
+		$.ajax({
+			url: 'http://127.0.0.1:8000/ajax/FGS/archive/',
+			success: function(data){
+
+        		// Add content to the proposal array div
+        		for (var i = 0; i < data.thumbnails.proposals.length; i++) {
+
+        			// Parse out useful variables
+        			inst = data.inst;
+        			prop = data.thumbnails.proposals[i];
+        			thumb = data.thumbnails.thumbnail_paths[i];
+        			n = data.thumbnails.num_files[i];
+
+        			// Build div content
+        			content = '<div class="proposal text-center">';
+        			content += '<a href="/' + inst + '/archive/' + prop + '/" id="proposal' + i + '" proposal="' + prop + '"';
+        			content += '<span class="helper"></span><img src="{{ static("thumbnails/") }}' + thumb + '" alt="" width=100%>';
+        			content += '<div class="proposal-color-fill" ></div>';
+        			content += '<div class="proposal-info">';
+        			content += '<h3>' + prop + '</h3>';
+        			content += '<h6>' + n + ' Files</h6>';
+        			content += '</div></a></div>';
+
+        			// Add the content to the div
+        			$("#proposal-array")[0].innerHTML += content;
+
+				// Replace loading screen with the proposal array div
+            	document.getElementById("loading").style.display = "none";
+            	document.getElementById("proposal-array").style.display = "block";
+        		};
+		}});
+	};
+	</script>
+
 
 {% endblock %}
 
 {% block content %}
+
+	<style type="text/css">
+		div#proposal-array { display: none; }
+		div#loading { text-align:center; z-index: 1000; }
+	</style>
 
     <main role="main" class="container">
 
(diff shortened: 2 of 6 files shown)


bourque:
Thanks @pep8speaks, glad to see you decided to start working again in 2019!

bourque:
@pep8speaks

bourque:
@pep8speaks suggest a diff

bourque:
I believe this PR is ready to be reviewed.  Changes include:

- Added new `get_base_url()` function to `utils.py` that returns a `base_url` based on the user running the code (i.e. if `admin_account`, return `dljwql.stsci.edu`, if not, return `localhost`
- Modified `test_api_views` to use new `get_base_url` function
- Rearranged functions in `utils.py` to be in alphabetical order
- Various minor edits based on `flake8` and @pep8speaks suggestions.
- Added `json_dumps_params` parameter to `JsonResponse()` calls in `api_views.py`, which makes the rendered JSON object be indented and nicely formatted
- Created a `thumbnails_ajax()` and `archive_thumbnails_ajax()` functions in `data_containers` and `views.py`, respectively to return data needed for the `ajax()` method in the `archive.html` and `thumbnails.html` templates
- Refactored `thumbnails()` function in `data_containers` to only return data that is otherwise not returned by `thumbnails_ajax()`
- Added a loading animation to the `archive.html` and `thumbnails.html` templates, along with supporting `css`.
- Added various javascript functions to `jwql.js`, `archive.html`, and `thumbnails.html` to support the loading of data dynamically using `ajax` method.
- Fixed indentation level in `show_only()` function in `jwql.js`
- Disabled the button to Unlooked images since that feature no longer works with these updates.

More on the last bullet: This PR currently breaks the Unlooked page, since the `ajax` method requires knowledge of which `proposal` is being used, and with unlooked images there is no way of knowing that.  I could spend some time and effort to build some workarounds, but I wonder if it is time to re-factor how we support the concept of "unlooked images".  We should probably better define what we mean by "unlooked", and how we want to keep track of which images have been looked at and which have not.  I'm inclined to tackle this issue in a future, separate PR.

I am assigning @laurenmarietta as reviewer since this is mostly web app related code.  @laurenmarietta I am aware that you will be away for some time, so no rush on this!  Apologies that this PR kind of snowballed into a big one, but most of these changes are related in some way!

jwql/website/apps/jwql/static/css/loader-magnify.css
```diff
@@ -0,0 +1,74 @@
...
+  position: absolute;
+  top: 68px;
+  left: 85px;
+}
+.lds-magnify {
+  width: 200px !important;
+  height: 200px !important;
+  -webkit-transform: translate(-100px, -100px) scale(1) translate(100px, 100px);
+  transform: translate(-100px, -100px) scale(1) translate(100px, 100px);
+}
```
laurenmarietta: Someday, I would love to hear how you did this magic!!

jwql/website/apps/jwql/templates/about.html
```diff
@@ -57,6 +57,9 @@ <h4>Acknowledgements</h4>
         {% endfor %}
         </ul>
 
+        <h4>Other attributions:</h4>
+        The loading animation 'magnify' is provided by <a href="https://loading.io">loading.io</a>
+
```
laurenmarietta: Unrelated thought: should we include an attribution to the Astropy project?

jwql/website/apps/jwql/templates/thumbnails.html
```diff
@@ -2,6 +2,48 @@
...
+    <style type="text/css">
+        div#thumbnail-array { display: none; }
+    </style>
+
+    <script type="text/javascript">
+	/**
+	 * Updates the thumbnail-array div with interactive images of thumbnails
+	 * @param {Object} data - The data returned by the update_thumbnails_page AJAX method
+	 */
+    function update_thumbnail_array(data) {
```
laurenmarietta: Same question here.

jwql/website/apps/jwql/static/js/jwql.js
```diff
@@ -150,4 +157,93 @@ function sort_by_thumbnails(sort_type) {
...
+ * Updates the img_show_count component
+ * @param {Object} data - The data returned by the update_thumbnails_page AJAX method
+ */
+function update_show_count(data) {
+
+    num_rootnames = Object.keys(data.file_data).length;
+    content = 'Showing ' + num_rootnames + '/' + num_rootnames + ' activities';
+    content += '<a href="https://jwst-docs.stsci.edu/display/JDAT/File+Naming+Conventions+and+Data+Products" target="_blank" style="color: black">';
+    content += '<span class="help-tip mx-2">i</span></a>';
+    $("#img_show_count")[0].innerHTML = content;
```
laurenmarietta: I think I'm having a little trouble working out why, for the `thumbnails.html` page, divs for the `img_show_count` and sort options are written here in `jwql.js`, but similar/identical divs are still written in the HTML for `archive.html`. I would expect they would be handled in the same place? Sorry if I'm missing something simple!

jwql/website/apps/jwql/templates/archive.html
```diff
@@ -2,73 +2,123 @@
...
-				<!-- <a -->
-			</div>
-		{% endfor %}
-		</div>
-		<a id="no_proposals_msg" style='display: none'>No proposals match the search criteria.</a>
-
-	</main>
+        <h3 id='title'>Archived {{ inst }} Images</h3>
+
+        <p id='img_show_count'></p>
```
laurenmarietta: Small thing: originally, the link and span below for the JDox tip were included in the `<p>` tag for the `img_show_count`, which kept the little (?) in line with the text about how many proposals were shown. I think the `</p>` here needs to be deleted for that to still work. Same with line 60 in `thumbnails.html`.

jwql/website/apps/jwql/static/js/jwql.js
```diff
@@ -75,53 +81,53 @@ function search(n_proposals) {
```
laurenmarietta: In bad news, I am seeing that the search function isn't working correctly now. Typing in a particular number isn't accurately filtering out proposals that don't start with that number. Do you see this behavior, too? 

I'm not sure at the moment what might be causing it - maybe a misinterpretation of what's being typed into the search box, or a misassignment of the `proposal` attribute to each thumbnail div. If you don't have an immediate idea of why this might be happening, I'm happy to do a deeper dive.

laurenmarietta: Though the sort functions are still working, interestingly...

jwql/website/apps/jwql/static/css/loader-magnify.css
bourque: @laurenmarietta This is pretty much copy pasta from the website that I got this from, so I would also like to know how this magic works! 👀

jwql/website/apps/jwql/templates/about.html
bourque: @laurenmarietta Good idea!  I was thinking that we could also add an attribution for Ann Fields who helped us envision our logo improvements.  We should open a separate 'update-attributions' PR for these.

jwql/website/apps/jwql/templates/archive.html
```diff
@@ -2,73 +2,123 @@
...
+        div#proposal-array { display: none; }
+    </style>
+
+    <script type="text/javascript">
+    /**
+     * Updates various compnents on the archive page
+     * @param {String} inst - The instrument of interest (e.g. "FGS")
+     * @param {String} base_url - The base URL for gathering data from the AJAX view.
+     */
+    function update_archive_page(inst, base_url) {
```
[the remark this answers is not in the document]
bourque: @laurenmarietta Ah yes, sorry I should have mentioned this before --

I am finding that any javascript function that references `src="{{static()}}"` in some way (e.g. `static("thumbnails/")` does not work when they live in `jwql.js`.  This seems to indicate that `static()` points to a path that is relative to the `website/apps/jwql/static/` directory instead of `website/apps/jwql/static/js/` directory.  Perhaps this has something to do with how we define the static paths in our `settings.py` file.  For now, I just kept things within the HTML template so that it works, but I do think we should figure out a better way in a separate PR, if you agree.

jwql/website/apps/jwql/static/js/jwql.js
bourque: @laurenmarietta Good point.  Probably this javascript function could be generalized and used in both templates.  I'll try to do that.

jwql/website/apps/jwql/static/js/jwql.js
bourque: @laurenmarietta Hmmm, good catch, I'll look into this.

bourque:
@laurenmarietta My latest commit (82681a9) should address the issue with the JDox page help tip button and the suggestion to use a generalized `update_show_count` function in both `archive.html` and `thumbnails.html`.

The only lingering issue is that of the search bar not working as expected, though I don't seem to see that behavior.

jwql/website/apps/jwql/templates/archive.html
```diff
@@ -2,73 +2,118 @@
...
+                for (var i = 0; i < data.thumbnails.proposals.length; i++) {
+
+                    // Parse out useful variables
+                    prop = data.thumbnails.proposals[i];
+                    thumb = data.thumbnails.thumbnail_paths[i];
+                    n = data.thumbnails.num_files[i];
+
+                    // Build div content
+                    content = '<div class="proposal text-center">';
+                    content += '<a href="/' + inst + '/archive/' + prop + '/" id="proposal' + i + '" proposal="' + prop + '"';
```
laurenmarietta: Okay, I figured out the search issue. Weird that still don't see it...

The javascript function that decides which thumbnails to hide and show when a value is entered into the search bar looks at the different id values for the thumbnails. These are `id=proposal#` where the `#` was originally taken from the loop iterator in Django, and started at 1. The new way (here) is a javascript for loop that starts at 0! 

When I tested it, changing the `i` in this line to `(i + 1)` fixed the problem.

bourque:
@laurenmarietta Ahhh, I understand now.  Sorry, before I was thinking you were referring to the search bar on the homepage.  I do see the issue on my end, and your suggested fix does resolve it.  I've fixed it in the latest commit (8ccbf2c).

bourque:
@laurenmarietta With your final approval, I think this is ready to merge.

jwql/website/apps/jwql/templates/archive.html
```diff
@@ -2,73 +2,118 @@
...
+        <h3 id='title'>Archived {{ inst }} Images</h3>
+
+        <p id='img_show_count'></p>
+        <hr>
+
+        <!-- Filter and sort options -->
+        <div class="d-flex flex-row">
+            <!-- Search Bar -->
+            <div class="input-group mb-3">
+              <input id='search_box' type="text" class="form-control" placeholder="Search" aria-label="Search by proposal" oninput="search('{{ num_proposals }}');">
```
laurenmarietta: So close!! Unfortunately, now I am seeing 
> Showing 17/{{ num_proposals }} proposals

when I type something into that search bar in `archive.html`. I think that it's because of how this `num_proposals` variable is getting passed into the javascript `search()` funtion.

Problem is, this looks completely correct to me. Do you have any ideas why that might be happening?

bourque: @laurenmarietta Ahh good catch again! :scream: It is indeed because of the way `num_proposals` is being passed.  The problem is that the way I restructured things, the variable is no longer available in the view.  I'll have to come up with a workaround.

bourque:
@laurenmarietta Ok, turns out the fix was rather simple (4ec1016).

PR #245, opened 2019-01-10
Integrate auth.mast

### LINKED ISSUE #201: Integrate auth.mast with the web app
[auth.mast](https://auth.mastdev.stsci.edu) is a service provided by ASB that will allow users of a web application to authenticate through MAST.  This service can help us build an authentication layer on our web app and restrict content to users who have a MAST account.

bourque:
This PR includes changes to the web app to utilize the [`auth.mast`](https://auth.mastdev.stsci.edu/info) service for authenticating users.

Closes #201


bourque:
Much of these changes will disappear when this branch can be rebased after #242 is merged.

bourque:
I think this is mostly ready to jetty, just need to fix some lingering issues with this working properly from the dev server.

bourque:
This is now ready for review!  @laurenmarietta Could you take a look whenever you have a chance?  If you would prefer to wait until we do a code review on this, that would be fine. You may also want to wait until #242 is merged so that most of these perceived changes disappear.

laurenmarietta:
@bourque would you mind re-merging `master` into this branch, so the changes from #242 don't appear? I'm planning to review this today.

bourque:
@laurenmarietta I've resolved some merge conflicts and the branch should be up to date now!

requirements.txt
```diff
@@ -1,6 +1,7 @@
 asdf>=2.3.0
 astropy>=3.0
 astroquery==0.3.9
+authlib==0.10
```
laurenmarietta: Does this new dependency also need to be added to `setup.py`?

jwql/website/apps/jwql/oauth.py
```diff
@@ -0,0 +1,177 @@
...
+        access_token_params=None,
+        refresh_token_url=None,
+        authorize_url='https://{}/oauth/authorize'.format(auth_mast),
+        api_base_url='https://{}/1.1/'.format(auth_mast),
+        client_kwargs=client_kwargs)
+
+    return oauth
+
+
+JWQL_OAUTH = register_oauth()
```
laurenmarietta: Is there a particular reason this global variable is at the bottom, rather than the top?

jwql/website/apps/jwql/views.py
```diff
@@ -47,8 +48,7 @@
 from .data_containers import thumbnails
 from .data_containers import thumbnails_ajax
 from .forms import FileSearchForm
-from jwql.utils.utils import get_base_url, get_config, JWST_INSTRUMENTS, MONITORS, INSTRUMENTS_CAPITALIZED
```
laurenmarietta: Looks like the line of `jwql.utils.utils` imports accidentally got deleted here.

jwql/website/apps/jwql/views.py
```diff
@@ -194,6 +194,43 @@ def archive_thumbnails_ajax(request, inst, proposal):
     data = thumbnails_ajax(inst, proposal)
 
     return JsonResponse(data, json_dumps_params={'indent': 2})
+ 
+
+def authorize(request):
```
laurenmarietta: I might suggest putting this new function, along with `login` and `logout` into a separate module, rather than within `views.py`. I think that might make more organizational sense.

jwql/website/apps/jwql/oauth.py
```diff
@@ -0,0 +1,177 @@
...
+            redirect_uri = os.path.join(get_base_url(), 'authorize')
+            return JWQL_OAUTH.mast_auth.authorize_redirect(request, redirect_uri)
+
+        return fn(request, user)
+
+    return check_auth
+
+
+def auth_info(fn):
+    """A decorator function that will return user credientials along
```
laurenmarietta: Small typo here - "credientials"

jwql/website/apps/jwql/templates/base.html
```diff
@@ -75,7 +78,26 @@
 	        	</ul>
 	        </div>
 
-	    	<a class="navbar-right" href="https://github.com/spacetelescope/jwql" target="_blank"><img src={{ static('img/githublogo.svg') }} height=25 width=25></a>
+	        <!-- Oauth login and GitHub -->
+	        <div class="navbar-right">
+	            <ul class="navbar-nav mr-auto">
+	            	{% if user.ezid %}
+	            		<li>
+	            			<a class="nav-link" href="https://auth.mastdev.stsci.edu/tokens" id="oauth_user" style="transform:translateY(20%);">{{ user.ezid }}</a>
```
laurenmarietta: So, I am still getting a `UndefinedError: 'user' is undefined` for what seems to be every page except the homepage. I think this is because the `user` variable, which is used here in `base.html` and thus is loaded on every webpage, isn't always getting passed into the `context` dictionary (like it is on line ~298 of `views.py`). 

I imagine adding `user` as an argument to all the views and as an entry in each view's context dict might fix this?

laurenmarietta: I tried to do this for the `instruments()` view and it is unfortunately not quite as simple as I would have hoped. It seems like Django passing specific variables into the view (e.g. `inst`) messes up the `auth_info` decorator. Hopefully you might have better luck with it? Or an idea for another kind of fix?

requirements.txt
bourque: @laurenmarietta It probably does, but there are probably a bunch of dependencies that we are missing in `setup.py` since we have been neglecting it, so perhaps updating `setup.py` deserves its own issue/PR (along the lines of the discussion in #261)

jwql/website/apps/jwql/oauth.py
bourque: @laurenmarietta Yes, since it is calling the `register_oauth()` function, the `JWQL_OAUTH` variable must be defined after that function is defined.  Putting it at the top would cause a `NameError` stating that `register_oauth` is not defined.

One might argue this is bad practice.  I guess its a matter of convenience.  Doing it this way allows a user to do:

```python
from jwql.website.apps.jwql.oauth import JWQL_OAUTH
```

instead of having to do:

```python
from jwql.website.apps.jwql.oauth import register_oauth
jwql_oauth = register_oauth()
```

I'm not exactly sure the best way to handle this 🤷‍♂️

jwql/website/apps/jwql/views.py
bourque: @laurenmarietta Good idea, that makes sense.  I probably can just stick them in `oauth.py` so all of the authentication stuff is contained in one place.

jwql/website/apps/jwql/templates/base.html
bourque: @laurenmarietta Good catch, I was so fixated on getting this working on the homepage that I neglected to notice that it broke all of the other pages 😱 .  Yes, I think the problem is just as you describe, I will try to find a solution.

bourque: @laurenmarietta I think I have come up with a solution for this (f73df96) -- it was a matter of passing `user` to each view, and doing some tricky stuff with `**kwargs` in the `auth_info` decorator.

Though this appears to be working, i'm not 100% convinced that this is the best way to do it.  Perhaps an issue to tackle in a future issue/PR.  I think there may exist a better way in separating `context` needed for `base` from `context` needed for the other templates.  At the very least, perhaps we could define a convenient `get_base_context()` function that can be called in each view.

jwql/website/apps/jwql/oauth.py
laurenmarietta: Ahh, I see now. I've gotten so used to functions (where the order doesn't matter in the same way) that I completely forgot that it would matter in this case!

This seems fine to me then, even if it's a little funny.

jwql/website/apps/jwql/templates/base.html
laurenmarietta: A lovely fix - all the pages are working now. Thank you!

I definitely don't have any intuition as to whether it is normal to pass the user as an argument into all the views... As per usual, I'd say the Django docs are a good resource for how to do things sleek-ly. I have just barely skimmed it, but it seems like Django might have a way to let us integrate our customized MAST authentication into their usual authentication system? For a future PR: https://docs.djangoproject.com/en/2.1/topics/auth/

bourque: @laurenmarietta Thanks for looking into that.  I will open a separate "make passing base context better" issue for this.

jwql/website/apps/jwql/oauth.py
```diff
@@ -0,0 +1,262 @@
...
+    base_url = get_base_url()
+    if '127' in base_url:
+        domain = '127.0.0.1'
+    else:
+        domain = base_url.split('//')[-1]
+
+    response = redirect("/")
+    cookie_args = {}
+    # cookie_args['domain'] = domain
+    # cookie_args['secure'] = True
```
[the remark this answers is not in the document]
bourque: @laurenmarietta Ah yes, I forgot to mention these.  This is something to also figure out in a future PR -- I got this idea from Christian Mesh who indicated that we should set certain cookie properties to adhere to best security practices.  However, when I tried setting these, I was finding strange behavior when trying to log out of a user account, so I just gave up for now.  I can open a separate issue for figuring this out.

PR #248, opened 2019-01-15
Expand filename_parser to handle time series and stage 3 filenames

### LINKED ISSUE #237: Update filename_parser to handle time series and stage 3 filenames
There have been recent updates to JWST filename conventions, specifically for time series observations and stage 3 data products, as described in [this](https://jwst-docs.stsci.edu/display/JDAT/File+Naming+Conventions+and+Data+Products) and [this](https://jwst-pipeline.readthedocs.io/en/stable/jwst/introduction.html#output-file-and-associations).  We should update `utils.filename_parser()` to handle this.

bourque:
This PR modifies `utils.filename_parser()` to be able to handle filenames for time series observations (e.g. `jw00733003001_02101_00002-seg001_nrs1_rate.fits`) and stage 3 products (e.g. `jw80600-o009_t001-epoch1_miri_f1130w_i2d.fits`). 

Corresponding tests have also been added to `test_utils.py`, and `test_utils.py` was modified to utilize `pytest.mark.parametrize`.

Closes #237


bourque:
I don't think there is anything in @pep8speaks to really worry about.  A lot of these are because of code alignment, which makes the code easier to read.

bourque:
@laurenmarietta 

> looking at the JDox page, it says that the source ID in association files can be up to 5 digits. When that is being parsed in lines 118 and 128, the ID is restricted to 3 digits (which is the limit for the target ID). I worry that could end in an error?

Good catch!  We can probably can get around this by adding another `filename_type` wherein a `source_id` is used instead of a `target_id`.  I'll see if I can implement that. 

> Which makes me think: the best way to test this ultimately would be to run the filename_parser on all the filenames we currently have in the filesystem (which we determined were all valid names, right?) and make sure it can handle them all. Probably not a good idea as the filesystem scales up, but might be good to do earlier on so we're sure we've covered all the bases. I'd be happy to try and do that if you like.

This is a good idea!  Feel free to open a separate issue/PR for this.  We could write a unit test to test if **all** filenames pass, instead of having a unit test that tests a given file individually -- using the latter could inflate our test suite by a lot, if that makes sense.

laurenmarietta:
Got it - I'm on it!

bourque:
@laurenmarietta My latest commit (https://github.com/spacetelescope/jwql/pull/248/commits/a35108c472e32cd0fca78eec808a922be83339d1) should address the concerns about `source_id`.  I also moved some code around to avoid some redundant code, and added declarations for raw strings to avoid some `DeprecationWarnings` (see [relevant StackOverflow](https://github.com/spacetelescope/jwql/pull/248/commits/a35108c472e32cd0fca78eec808a922be83339d1))

laurenmarietta:
I'm writing the test for the whole filesystem (only 333 out of ~10,000 are failing!), and there are some specific filetypes that are failing: 
- jw90001001002_02101_00001_nis_**o001**_crf.fits
- jw00307001001_02101_00001_nis_**a3001**_**crfints**.fits
- jw00322002007_02101_00001_nrcalong_**c1001**_crf.fits

So it looks like 
1) `crfints` should be added to the `FILE_SUFFIX_TYPES` list, and 
2) there is another common filetype where the AC_ID is after the _instrument_ rather than after the _proposal number_. 

Thought I'd share so you could include those changes in this PR?

jwql/utils/utils.py
```diff
@@ -109,31 +94,108 @@ def filename_parser(filename):
...
+        r"(?P<parallel_seq_id>\d{1})"\
+        r"(?P<activity>\w{2})"\
+        r"_(?P<exposure_id>\d+)"\
+        r"_(?P<detector>((?!_)[\w])+)"
+
+    # Stage 3 filenames with target ID, e.g. "jw80600-o009_t001_miri_f1130w_i2d.fits"
+    stage_3_target_id = \
+        r"jw" \
+        r"(?P<program_id>\d{5})"\
+        r"-(?P<ac_id>(o|c|a|r)\d{3})"\
```
laurenmarietta: This also just made me realize that AC_IDs with `c`, `a`, or `r` can have 4 digits where the regex string is currently looking for 3. Also on lines 126, 135, and 145.

bourque: @laurenmarietta 😱 Nice catch.  I'll try to get around this one too.  Perhaps there is a regex option where i can say "3 or 4".

bourque:
@laurenmarietta 

> there is another common filetype where the AC_ID is after the instrument rather than after the proposal number.

I suggest that we hold off on coding a workaround for this one.  I would like to ask the powers that be over in the archive if this is expected, since it goes directly against what they have documented.  Or perhaps the filesystem that we are using to test needs to be updated.

Johannes-Sahlmann:
Examples of FGS GS acquisition and fine-guide file names in MAST:
jw00788003001_gs-acq1_2019160140055_uncal.fits
jw00788003001_gs-fg_2019160140451_uncal.fits
Note that these are of PRODUCT CATEGORY=AUXILIARY (as opposed to SCIENCE)

bourque:
Thanks @Johannes-Sahlmann 

Some additional documentation I've been made aware of thanks to Kim Duprie:

https://innerspace.stsci.edu/pages/viewpage.action?pageId=94092600#SummaryofAssociationFileNamingConvention

https://innerspace.stsci.edu/display/SCSB/JWST+Science+Data+Products

bourque:
I'm going to move this back to a `WIP` until I can make sense of all of these different filename types.

bourque:
On second thought, I think we should go ahead and merge these changes and worry about future changes in a separate PR, especially considering this PR is holding up the progress of #255 

Looking at all of the possibilities described in [this document](https://innerspace.stsci.edu/pages/viewpage.action?pageId=94092600), I think we are going to have to make some rather significant changes, and we can worry about those whenever we update our filesystem to a newer copy (still waiting to hear back from Mary Romelfanger about this).  That way, we should have some of these filetypes available to us for testing.

@laurenmarietta Were there any other changes necessary for this PR?

PR #253, opened 2019-01-21
style guide: naming conventions

Johannes-Sahlmann:
Add standards for instrument names and program IDs.

### DIFF
--- a/style_guide/style_guide.md
+++ b/style_guide/style_guide.md
@@ -67,6 +67,16 @@ Additionally, developers of this project should be mindful of application securi
 - Each function/method should have at minimum a description, `Parameters` (if necessary), and `Returns` (if necessary) sections
 
 
+`jwql`-Specific Variable Value/Type Standards
+---------------------------------------------
+
+To the extent possible, `jwql` shall define frequently-used variable types/values consistently. A list of adopted standards is provided below:
+
+- **JWST instrument names**: In all internal references and structures (e.g. dictionaries) instrument names shall be all lower-case strings, i.e. one of `fgs`, `miri`, `niriss`, `nircam`, `nirspec`. When variations are required for interfaces, e.g. `Nircam` for MAST, `NIRCam` or `NIRCAM` for SIAF, etc. these should be defined as dictionaries in [`jwql/utils/utils.py`](https://github.com/spacetelescope/jwql/blob/master/jwql/utils/utils.py) and imported from there.
+
+- **Program/proposal identifiers**: JWST program IDs shall be stored and referred to internally as integers and parsed to strings only when needed. For example, the inputs `"001144"` and `"1144"` shall both be converted to an integer variable with value `1144`.  
+
+
 Tools and Library Recommendations
 ---------------------------------
 
--- a/style_guide/style_guide.md
+++ b/style_guide/style_guide.md
@@ -72,7 +72,7 @@ Additionally, developers of this project should be mindful of application securi
 
 To the extent possible, `jwql` shall define frequently-used variable types/values consistently. A list of adopted standards is provided below:
 
-- **JWST instrument names**: In all internal references and structures (e.g. dictionaries) instrument names shall be all lower-case strings, i.e. one of `fgs`, `miri`, `niriss`, `nircam`, `nirspec`. When variations are required for interfaces, e.g. `Nircam` for MAST, `NIRCam` or `NIRCAM` for SIAF, etc. these should be defined as dictionaries in [`jwql/utils/utils.py`](https://github.com/spacetelescope/jwql/blob/master/jwql/utils/utils.py) and imported from there.
+- **JWST instrument names**: In all internal references and structures (e.g. dictionaries) instrument names shall be all lower-case strings, i.e. one of `fgs`, `miri`, `niriss`, `nircam`, `nirspec`. When variations are required for interfaces, e.g. `Nircam` for MAST, `NIRCam` or `NIRCAM` for SIAF, etc. these should be defined as dictionaries in [`jwql/utils/constants.py`](https://github.com/spacetelescope/jwql/blob/master/jwql/utils/constants.py) and imported from there.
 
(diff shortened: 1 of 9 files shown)


Johannes-Sahlmann:
@bourque : I made some related changes to https://github.com/spacetelescope/jwql/blob/master/jwql/utils/utils.py but have not made the corresponding changes in the code that uses the modified global variables. I want to wait for your initial review to see what you think. 
The instrument name variables are of course used almost everywhere and we have to balance effort against benefit when trying to align with a common naming convention and using the centrally defined global variables.

laurenmarietta:
A quick thought: the `JWST_INSTRUMENTS` list is also used to dynamically generate text in the web app, which is why they used the mixed-case names. It might be a good idea to add a `JWST_INSTRUMENTS_MIXED_CASE` list if the original one is made lowercase.

Also, I wonder whether it would make sense to use this PR to move such lists and dictionaries into a `constants.py` file, as you made for MIRaGe, rather than the `utils.py` one?

Johannes-Sahlmann:
@laurenmarietta The mixed case instrument names can be retrieved from JWST_INSTRUMENT_NAMES_MIXEDCASE

I would support the creation of a constants.py file. @bourque what do you think about that?

bourque:
@Johannes-Sahlmann @laurenmarietta Sounds good, I am in favor of tackling the `constants.py` issue in this PR as well.  Get two birds stoned at once. 

Just to re-iterate: With @Johannes-Sahlmann changes, we have:

- `JWST_INSTRUMENT_NAMES` : list of instruments, everything lowercase (e.g. `nircam`)
- `JWST_INSTRUMENT_NAMES_SHORTHAND`: maps shorthand notation to full notation (e.g. `nrc:nircam`)
- `JWST_INSTRUMENT_NAMES_MIXEDCASE`: maps lowercase to mixed case (e.g. `nircam:NIRCam`)
- `JWST_INSTRUMENT_NAMES_UPPERCASE`: maps lowercase to upper case (e.g. `nircam:NIRCAM`)
- `JWST_INSTRUMENT_NAMES_MAST`: maps lowercase to special MAST case (e.g. `nircam:Nircam`) (aside: why would MAST do this??) 

I think this is all good, and gives us full coverage of all of the possibilities, or at least the ones that we care about for now.  @Johannes-Sahlmann, I agree with the description that you added in the Style Guide. In code, let's default to all lowercase.  Looks good.

style_guide/style_guide.md
```diff
@@ -67,6 +67,16 @@ Additionally, developers of this project should be mindful of application securi
...
 
 
+`jwql`-Specific Variable Value/Type Standards
+---------------------------------------------
+
+To the extent possible, `jwql` shall define frequently-used variable types/values consistently. A list of adopted standards is provided below:
+
+- **JWST instrument names**: In all internal references and structures (e.g. dictionaries) instrument names shall be all lower-case strings, i.e. one of `fgs`, `miri`, `niriss`, `nircam`, `nirspec`. When variations are required for interfaces, e.g. `Nircam` for MAST, `NIRCam` or `NIRCAM` for SIAF, etc. these should be defined as dictionaries in [`jwql/utils/utils.py`](https://github.com/spacetelescope/jwql/blob/master/jwql/utils/utils.py) and imported from there.
+
+- **Program/proposal identifiers**: JWST program IDs shall be stored and referred to internally as integers and parsed to strings only when needed. For example, the inputs `"001144"` and `"1144"` shall both be converted to an integer variable with value `1144`.
```
bourque: @Johannes-Sahlmann I wonder if there is any value in defaulting to storing program identifiers as strings, instead of integers.  Though program IDs are comprised of integers, they are not meant to be used as numbers to be added, subtracted, etc.  I'm struggling to think of scenarios in which it would be useful to have one of these identifiers in code as a `int` class.

Another aspect that I think may be important for consistency sake is ensuring the program identifiers use all 6 digits with zero-padding when necessary (i.e. `001144` instead of `1144` or `01144`), which would be easier to enforce with strings.

Do you have any thoughts about this?

bourque: @Johannes-Sahlmann Those are good points.  In that case, I think what you have here in the `style_guide` looks good.  In the meantime, let's try to strive for allowing functionality for both zero-padded and non-zero-padded program IDs when possible; I assume this is an easy fix for our web app search bar.

Johannes-Sahlmann:
How is it possible that all jenkins checks have passed in this PR?
`test_monitor_mast.py` includes `from jwql.utils.utils import JWST_INSTRUMENTS` which is a variable that does not exist in the current commit. Locally, the test fails for that reason.

The only explanation I have is that that test is not executed.

laurenmarietta:
Interestingly, it looks like Jenkins did try to run that test and it did fail... But then for some reason it didn't consider that as a failed build? I don't know why that would be.
screen shot 2019-01-25 at 11 57 01 am

Johannes-Sahlmann:
I am not sure my local tests and the jenkins tests cover all the side-effects of these changes, in particular in the web-app. 
How should we go about making a systematic inspection?

bourque:
@Johannes-Sahlmann Yeah, it doesn't hurt to run the website locally and click through all of the buttons to make sure everything works as expected.  Perhaps this will be alleviated one day with #220 but we are not there yet.

Johannes-Sahlmann:
Thanks @bourque . That helped to catch and fix some more dependencies.

bourque:
@laurenmarietta @Johannes-Sahlmann  In regards to your earlier comments, I've opened up an issue for this (#262)

bourque:
@Johannes-Sahlmann Is this ready for review?

Johannes-Sahlmann:
Ready for review!

bourque:
Though the checks are not really passing (#262) all of the tests pass locally when I try them, and the web app appears to be working as expected, so I am going to go ahead and merge this.  If there are indeed issues with the Jenkins tests, we will pick them up when addressing #262

PR #258, opened 2019-01-24
Logging guidelines

catherine-martlin:
Putting together logging guidelines.


bourque:
@catherine-martlin FYI -- I added a bunch of text here to beef this up.  But we will still probably want to update this/put the finishing touches once we merge in #327.

bourque:
@catherine-martlin This looks good to go for me, but since ive added a fair bit to this PR, could you take a look through and let me know if we are missing anything, or if anything is wrong?

catherine-martlin:
@bourque Just read through the pages/updates and it looks great! Thanks for flushing this out/wrapping it up while I've been MIA on other project deadlines!

PR #265, opened 2019-01-29
Test filename_parser with whole filesystem

### LINKED ISSUE #255: Add unit test to parse all filesystem names
Add a test to `test_utils.py` to run filename_parser on each file in the filesystem, as mentioned in #248:

> ...the best way to test this ultimately would be to run the `filename_parser` on all the filenames we currently have in the filesystem (which we determined were all valid names, right?) and make sure it can handle them all...
> 
> ...We could write a unit test to test if all filenames pass, instead of having a unit test that tests a given file individually -- using the latter could inflate our test suite by a lot, if that makes sense.

laurenmarietta:
This PR includes a new test in `test_utils.py` that runs `utils.filename_parser` on every file in our filesystem. I've marked this test as `xfail`, as there are currently 331 files in the filesystem that cannot be parsed, and I don't think that should lead to a build failing.

It might be worthwhile to still pass the information about which files are failing to the user in some way, though, since `xfail` doesn't show the failure message. Thoughts?

This PR also populates the `reason` argument for both `xfail` tests in `test_utils`.

Tagging @cracraft as a reviewer because of the work you did (and are doing!) with the filesystem monitor.

Closes #255.


jwql/tests/test_utils.py
```diff
@@ -175,6 +177,35 @@ def test_filename_parser(filename, solution):
...
+@pytest.mark.xfail(reason='Known non-compliant files in filesystem')
+def test_filename_parser_whole_filesystem(capsys):
+    """Test the filename_parser on all files currently in the filesystem.
+    """
+    # Get all files
+    filesystem_dir = get_config()['filesystem']
+    all_files = []
+    for dirName, _, fileList in os.walk(filesystem_dir):
+        for file in fileList:
+            if file.endswith('.fits'):
```
bourque: @laurenmarietta Good idea to check for `.fits` files.  There will probably come a time when rando MAST files are in the filesystem.

jwql/tests/test_utils.py
```diff
@@ -175,6 +177,35 @@ def test_filename_parser(filename, solution):
...
+                all_files.append(os.path.join(dirName, file))
+
+    # Run the filename_parser on all files
```
bourque: @laurenmarietta I think this is a good way to do this for now.  I do see this becoming potentially a problem when your filesystem gets to be huge -- this test might take forever to run.  But perhaps when it comes to that, we can implement `multiprocessing` or something to speed it up.

jwql/tests/test_utils.py
```diff
@@ -175,6 +177,35 @@ def test_filename_parser(filename, solution):
     assert filename_parser(filename) == solution
 
 
+@pytest.mark.xfail(reason='Known non-compliant files in filesystem')
+def test_filename_parser_whole_filesystem(capsys):
```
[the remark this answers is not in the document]
laurenmarietta: As far as I can tell, it gives you access to the captured system output (`sysout`?) that would otherwise be handled by `pytest`. It is only here because I was using it earlier to print out the names of the files that were non-compliant, even with the `xfail` being present. So I will remove it now!

Here's where I learned about it: https://stackoverflow.com/a/41533292/9910708

jwql/tests/test_utils.py
```diff
@@ -175,6 +177,35 @@ def test_filename_parser(filename, solution):
...
+    """Test the filename_parser on all files currently in the filesystem.
+    """
+    # Get all files
+    filesystem_dir = get_config()['filesystem']
+    all_files = []
+    for dirName, _, fileList in os.walk(filesystem_dir):
```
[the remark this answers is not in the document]
laurenmarietta: Whoops, that's what I get for lazy copy-and-pasting. Thanks!

jwql/tests/test_utils.py
laurenmarietta: Yes, I agree completely. This test already takes many times longer to run than the other tests in `test_utils`, I think just because of walking through the whole directory. So we should definitely keep an eye on it as things scale up!

PR #270, opened 2019-01-31
Fix unit tests in Jenkins builds

### LINKED ISSUE #262: Jenkins not actually running tests due to references to config.json
Despite Github indicating that checks are passing, it seems the Jenkins is not actually running tests.  [Here](https://ssbjenkins.stsci.edu/blue/organizations/jenkins/STScI%2Fjwql/detail/PR-158/7/pipeline/18#step-35-log-1) is an example of where it is choking -- a few tests that require access to `config.json`.

I think the temporary solution is to mark these tests as `xfail` until we can get a proper private jenkins server that has access to our `config.json` file (#144)

bourque:
This PR attempts to get our tests actually running in our Jenkins builds.

Closes #262


bourque:
@SaOgaz @Johannes-Sahlmann @laurenmarietta Could one of you all review this?

PR #280, opened 2019-02-05
Update install_requires in setup.py to adhere more closely to python best practices

### LINKED ISSUE #268: Update install_requires in setup.py
After some research on StackOverflow and discussion with @SaOgaz and @laurenmarietta on the subject of `setup.py` dependencies and `requirements.txt`/`environment.yml` dependencies (#261, #94), we should probably update the list of dependencies in `install_requires` in `setup.py` to better reflect what is actually needed to install and run our application.

In short: `setup.py` should _list the range of dependencies that allow a user to run any code in our repository_, while `environment.yml` should _list a set of specific dependencies that allow the application to operate on a specific architecture_

bourque:
This PR updates the dependencies in `setup.py` `install_requires` so that it is being used more appropriately, as discussed in #268.

Closes #268


laurenmarietta:
Just to test this super thoroughly, I created a new conda environment with only Python (and numpy) and tried to run the test suite after just running `setup.py` instead of the environment file. I don't necessarily expect users to do this, but it brought some things up:

- Firstly, I didn't get too far, because I'm getting an import error for `jwst`: "The JWST Calibration pipeline package has not been publicly released yet." I imagine that `jwql` depends on the development version of the pipeline, but I poked through our docs and didn't see any notes about that. I have `jwst=0.7.8rc9.dev257` in my original `jwql` environment, while this new minimal one has `jwst=0.0.0`. Long story short: do we need to be more specific about the version of the pipeline that's required, either in `setup.py`/`environment.yml` or in our docs?

- Secondly, I'm just curious what your rationale is for leaving out some of the dependencies in the other files from `setup.py`? This includes `asdf`, `ipython`, `jinja2`, `postgresql`, `pyscopg2`, `python-dateutil`, and `sphinx-automodapi`. Some (like `asdf` and `jinja2`) seem pretty significant.

bhilbert4:
I'm not sure what it means when you specify version 0.0.0 of something, but I will mention that so far, all versions of the JWST pipeline are development versions (i.e. there has been no formal release to the public yet). The current version of the pipeline is 0.13.0b or thereabouts. (0.7.8rc9 is very old and should be updated before you try running the pipeline on anything.)

I think that by having just `jwst`  with no version information in the environment file, the latest version is what it grabbed? Anyway, maybe none of that really matters for the purposes of the requirements file. I would say that the version number in there needs to move to something more recent though, assuming that 0.0.0 is grabbing....nothing? Or something old?

bourque:
@laurenmarietta @bhilbert4 Yes, my motivation for not including a version for `jwst` here is that the latest version would automatically be grabbed, but my guess is that the latest (and only) version on `pip` (where `install_requires` is gathering things from) is `0.0.0`.  My guess is that if we try supplying a more recent version number here, `setup.py` will choke on it.

On the other hand, not supplying the version number in `environment.yml` should retrieve the latest version on the `astroconda-dev` `conda` channel, which is probably more like a version that @bhilbert4 is describing.

bourque:
@laurenmarietta 

>* Secondly, I'm just curious what your rationale is for leaving out some of the dependencies in the other files from `setup.py`? This includes `asdf`, `ipython`, `jinja2`, `postgresql`, `pyscopg2`, `python-dateutil`, and `sphinx-automodapi`. Some (like `asdf` and `jinja2`) seem pretty significant.

My motivation here is that those dependencies are never actually directly imported anywhere within our package, and thus do not form a minimal set of requirements for the package to run.  I *think* that those dependencies would be picked up by proxy of other packages that we *do* depend on (e.g. when installing `jwst`, `asdf` should come with it because `asdf` is a dependency of `jwst`).  But on second thought, some of these are indeed needed in order for the code to run (e.g. `pyscopg2` would be needed to actually perform any sort of PostgreSQL transaction).

So I think I will modify the `install_requires` to include a few more of these.  I think the underlying problem here is that I still dont 100% understand what exactly should go in `install_requires` :eyes:

bourque:
@laurenmarietta Further to that point -- something like `ipython` is clearly not needed to actually operate code within `jwql`, but just something extra that we include for developer's sake.  I think things like that don't need to be included in `install_requires`

laurenmarietta:
Thanks for all the helpful elaboration and explanation!

Now I see that the `jwst` problem is totally a pip v. conda problem. So, if we wanted to include a functional version of `jwst` in `setup.py` alone, we'd have to figure out how to include a conda install from there, which sounds at best difficult and at worst impossible. Since it certainly works from `environment.yml` I will just drop it!

And your reasoning for leaving some of those packages out of `install_requires` makes sense - I didn't realize how many of those are installed as dependencies of the included packages. Though adding back dependencies that are needed for the package (not like `ipython`) but not installed as dependency-dependencies sounds like a good plan.

bourque:
@laurenmarietta Ok, I think this is good to go now.  I added dependencies for `jinja2` and `psycopg2`.  I tried adding `postgresql` but apparently it doesn't exist in `pip`?  Either way, I was able to set up a test environment like you did earlier and I was able to perform some database operations, so I guess it is not needed afterall (or it comes with the install of `psycopg2`).

Issue #210, opened 2018-12-04
Display which version of jwql is being used in the web app


bourque:
A cool idea inspired by the `exoctk` project -- We could put a small piece of text in the corner of the web app somewhere that indicates which version of `jwql` is being served.

frboffi:
yes! this is important to have!

bhilbert4:
What's the easiest way to find the version being run? Read it from `setup.py`? It looks like we don't currently have `__version__` set to anything.

bourque:
@bhilbert4 Yeah, I would say normally the best way is to do 

```python
jwql.__version__
```

but we don't have it set up.  I've never actually looked into doing that before.  Maybe you could include that as part of this PR?  I'm not sure what one has to do to provide this, but I imagine someone on StackOverflow has done this before.

laurenmarietta:
We could also add a version badge to the README:
https://img.shields.io/github/release/spacetelescope/jwql.svg

bhilbert4:
Lots of options: https://packaging.python.org/guides/single-sourcing-package-version/
Seems like reading from `setup.py` might be easiest. Most of the other options involve defining the version number somewhere other than in `setup.py`. Do you see any options in there that are more appealing?

For a moment I was excited about option 5 (using `pkg_resources.get_distribution`), but since that only knows about the installation metadata, it returned a very old version number for me, as I installed jwql a long time ago and I don't re-run `setup.py` for each new version.

bhilbert4:
But perhaps one of these options can make it such that the version number in `jwql` is stored in only a single place, rather than in `setup.py` and `conf.py`?

bourque:
@bhilbert4 Would it be possible to read in the `VERSION` in `setup.py` from the `jwql.__init__.py` file?  Something like:

```python
with open('../setup.py') as f:
    data = f.readlines()
__version__ == #parse out VERSION
```

Even if we had a separate file that stored the version, I think we would still need a similar code in `jwql.__init__.py` to define the `__version__` variable, unless I am misunderstanding.

bhilbert4:
Yes, that sounds right. It seems like common practice is to read in the version from somewhere (setup.py, __init__.py, etc). I'm wondering if we read in the version number from setup.py, that we could then also add the code to conf.py to read the version number in from setup.py.

bourque:
[...]

PR #296, opened 2019-02-13
Add version number

bhilbert4:
This PR adds code that reports the version of JWQL being run by the web app. The version number is printed in the bottom right corner of the page.

This also adds a `__version__` attribute to `jwql`.

```python
import jwql
print(jwql.__version__)
```

And finally, it adds a badge to the README listing the current release version.


bhilbert4:
The way the code is currently written, it works if a string is supplied to the javascript `version_url` function. But for some reason it fails when the `{{version}}` is supplied, even though it is available and a string.

bhilbert4:
Closes #210

bhilbert4:
Ok @bourque , @laurenmarietta , I think this is ready to review. The final change I made was to swap out the hard-coded jwql version numbers in `conf.py` with `jwql.__version__`. Assuming that this change is ok, the JWQL version number would then only have to be manually updated in `setup.py` in the future.

jwql/tests/test_setup_info.py
```diff
@@ -0,0 +1,28 @@
+#! /usr/bin/env python
```
bourque: @bhilbert4 Thanks for including a test for this!

jwql/tests/test_setup_info.py
```diff
@@ -0,0 +1,28 @@
...
+"""Tests for the ``setup.py`` module.
```
bourque: @bhilbert4 To be consistent with the names of the other tests, I would suggest renaming this module to `test_setup.py`

jwql/tests/test_setup_info.py
```diff
@@ -0,0 +1,28 @@
...
+---
+
+    These tests can be run via the command line (omit the ``-s`` to
+    suppress verbose output to ``stdout``):
+
+    ::
+
+        pytest -s test_setup_info.py
+"""
+import pytest
```
bourque: @bhilbert4 I don't think this import is actually needed afterall

jwql/website/apps/jwql/static/js/jwql.js
```diff
@@ -248,4 +248,14 @@ function update_thumbnails_page(inst, proposal, base_url) {
             document.getElementById("loading").style.display = "none";
             document.getElementById("thumbnail-array").style.display = "block";
         }});
-};
\ No newline at end of file
+};
+
+/* JS function to construct the URL corresponding to a specific GitHub release
```
bourque: @bhilbert4 Nice javascript docs!

jwql/website/apps/jwql/templates/base.html
```diff
@@ -115,6 +115,11 @@
 	        		<img src={{ static('img/STScIlogo_dark.png') }}   height=36>
 	        		<img src={{ static('img/JWSTlogo.png') }}   height=36>
 	        	</a>
+     	      	<div style="float:right; width:180px; text-align: center; color:white; font-size:12px">
```
bourque: @bhilbert4 I think its good habit to use a `css` `class` or `id` here instead of defining a `style` in the `div` itself.  You could do something like:

```html
<div id='version-div'>
```

and then in `jwql.css`:

```css
#version-div {
    float: right;
    width: 180px;
    text-align: center;
    color: white;
    font-size: 12px;
}
```

This way, styles are more consolidated in `jwql.css` instead of scattered around the templates themselves.

jwql/tests/test_setup_info.py
bhilbert4: Renamed.

jwql/tests/test_setup_info.py
bhilbert4: Removed. I reran the test and it still passes.

jwql/website/apps/jwql/static/js/jwql.js
bhilbert4: I copy from the best.

jwql/website/apps/jwql/templates/base.html
bhilbert4: Thanks. I was intending to get to that after figuring out my problems with the version text, and then it slipped my mind. Fixed now.

bhilbert4:
Comments all addressed.

PR #305, opened 2019-02-22
auth.mast example notebook

bourque:
This PR adds a `jupyter` notebook with examples showing how to integrate the `auth.mast` service into a `django` web application.  Hopefully it can be useful for others at ST wishing to integrate `auth.mast` like we have done!

### DIFF
--- a/notebooks/integrating_auth_mast_service.ipynb
+++ b/notebooks/integrating_auth_mast_service.ipynb
@@ -6,7 +6,7 @@
    "source": [
     "**This notebook serves to demonstrate how to integrate the `auth.mast` service into a Django web application.  The code was adopted from the solution derived for the `jwql` web application.  Much of the code was written by Christian Mesh of the Archive Science Branch.  Thank you Christian!**\n",
     "\n",
-    "**Please note that this notebook does not work right out of the box.  There are various tokens and metadata that must be provided, and the code within is designed for a specific project.  It may be beneficial to structure the code in a different way for different applications.  This notebook simply serves as a guide.**\n",
+    "**Please note that this notebook does not work right out of the box, and some of the cells will not run without error.  There are various tokens and metadata that must be provided, and the code within is designed for a specific project.  It may be beneficial to structure the code in a different way for different applications.  This notebook simply serves as a guide.**\n",
     "\n",
     "**Resources:**\n",
     "- The `auth.mast` service can be found here: https://auth.mast.stsci.edu/\n",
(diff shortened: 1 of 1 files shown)


bourque:
@laurenmarietta  Could you review this? This hopefully will become more clear when I can finally give my code review on this stuff.  Feel free to wait until that happens, if you prefer.

notebooks/integrating_auth_mast_service.ipynb
```diff
@@ -0,0 +1,455 @@
+{
+ "cells": [
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "**This notebook serves to demonstrate how to integrate the `auth.mast` service into a Django web application.  The code was adopted from the solution derived for the `jwql` web application.  Much of the code was written by Christian Mesh of the Archive Science Branch.  Thank you Christian!**\n",
+    "\n",
+    "**Please note that this notebook does not work right out of the box, and some of the cells will not run without error.  There are various tokens and metadata that must be provided, and the code within is designed for a specific project.  It may be beneficial to structure the code in a different way for different applications.  This notebook simply serves as a guide.**\n",
```
laurenmarietta: Are you actually able to run these functions in the notebook? It seems more to me like a collection of functions and things to integrate into a web app, but that can't be run here in this notebook format. (I got an error from the 4th cell that seems to be related to the function not existing inside the scope of a web app, for instance.) If I haven't misinterpreted the notebook, I think it might be helpful to be a bit clearer about that in the intro section up top: that the notebook isn't meant to be run really at all, but rather to serve as a reference for how to set up authentication within a web app. Does that make sense/sound reasonable?

notebooks/integrating_auth_mast_service.ipynb
```diff
@@ -0,0 +1,455 @@
...
+  },
+  {
+   "cell_type": "code",
+   "execution_count": null,
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "# Set some necessary global parameters\n",
+    "client_id = ''  # Provided by ASB\n",
+    "client_secret = ''  # Provided by ASB\n",
```
laurenmarietta: What does someone have to do to get the `client_id` and `client_secret` values? Are those just copied directly from the `config.json` on the VM, or are they individual-specific? (I can't remember how I got mine, unfortunately...) It might be a good idea to put a sentence or two about where to get those, either from JWQL members in the config file or by emailing ASB, as I anticipate that would be a question someone might ask.

notebooks/integrating_auth_mast_service.ipynb
bourque: @laurenmarietta Yes, that makes sense, and I can add some more text here to make it more clear -- this notebook is merely a reference, not an operational piece.

notebooks/integrating_auth_mast_service.ipynb
bourque: In the particular case of JWQL, they are indeed grabbed from the `config.json` file.  They are specific to the application, not an individual user.  But yes, I can be more clear here about this here.

FYI, I am hopefully going to be working with @cmesh-stsci soon to get a dummy project with its own `client_id` and `client_secret` that can be shared and used as an example.  If/when that happens, we could just put those tokens here.

bourque:
@laurenmarietta I think i've addressed your concerns, let me know if there is anything else!

PR #313, opened 2019-02-28
Update API docs

bourque:
This PR includes changes to get our API docs up to date with the current state of the repository.  It also includes a few changes that were necessary in order to get the ReadTheDocs builds to complete, namely, adding a workaround in the `jwql/__init__.py` file to avoid needing the `setup.py` file to define the `__version__`, since ReadTheDocs doesn't seem to have access to `setup.py` in this way.


bourque:
@laurenmarietta Could you review?

Unfortunately, ReadTheDocs is still not able to render docs for modules that require the `config.json` file, since importing those modules will result in an error since the ReadTheDocs environment doesn't contain a `config.json` file.  I'm not sure how to get around this problem at the moment :(

laurenmarietta:
I am happy to review this, buuut I have just realized I don't know how to see the changes you've made on readthedocs... is there a way to see the readthedocs for a specific branch, or generate the HTML files locally? Or shall I just review the changes of the raw `rst` files?

bourque:
@laurenmarietta Ah yes, good question.  I realize that I should add you as a maintainer to the project on ReadTheDocs, and in that way you should be able to see the build logs and the output product.  Are you able to see https://readthedocs.org/dashboard/jwql/?  If not, I think you just need to sign up/sign in to a ReadTheDocs account and I can add you as a maintainer.

If you would like to build and see the docs locally, I think you just have to do:

```
cd jwql/docs/
make html
open build/html/index.html
```

I've been keeping a separate branch called `update-docs` on `upstream` for the purposes of testing changes to API docs on ReadTheDocs.  On ReadTheDocs, there is a setting where you can point to a specific branch, and while testing I point to `update-docs`.  When the changes get merged, I switch it back to point to `master`.

docs/source/edb.rst
```diff
@@ -0,0 +1,15 @@
+*****
+tests
```
laurenmarietta: I think this should say `edb` instead of `tests`?

laurenmarietta:
I just had to link to my GitHub account and I can see the page you linked to, so that's good news! It looks like the "latest" version of the docs points to that `update-docs` branch, is that right?

I also added your instructions for how to build the docs locally to the wiki: https://github.com/spacetelescope/jwql/wiki/Adding-Sphinx-API-documentation

docs/source/edb.rst
bourque: Whoops!  Good catch!

bourque:
@laurenmarietta Sweet!  Yes, `latest` points to the `update-docs` branch right now so that I could test things out, but after we merge this, I will change this back to point to `master` ("Admin" > "Advanced Settings" > "Default Branch").

Thanks for adding that to the wiki page! Looks great!

Issue #294, opened 2019-02-13
Build production web server


bourque:
Build a web server, similar to our development web server, for production purposes.

bourque:
@laurenmarietta and I chatted with @mfox22 today about the infrastructure that exists in ASB.  We should have a look, and possibly use some of it for our web application: https://grit.stsci.edu/mast/asb-core/blob/master/doc/README.md

bourque:
I reached out to Christian Mesh about this.  We are going to meet sometime and try to hack together a working example.

mfox22:
I'd like to sync up with you all too. Let's discuss strategies for:
- Sharing code (automation, test frameworks, Ansible plays)
- Determining if there are opportunities to share infrastructure

bourque:
Some notes from meeting with Christian:

Jenkins Build:
push to pypi
generate custom conda file from template
artifact generate conda file
automatically run delpoy

Jenkins Deploy:
param env (dev/test/ops)
conda file that links to artifcat 

Ansible playbook (or just use ssh for now):
On each machine:
- stop service
- delete conda environment *
- create conda env
- start service

(Could put in HOME directory, but better to use "Host setup" w/ 50 GB storage inside of /opt/jwql/ for apps & logs)
cron @reboot command to start service

Also need to svc acct setup ssh into host as svc account from Jenkins

Look at common-mast for a simpler example

bourque:
After a meeting with the JWST mission office, we determined that we should have a conversation again with someone in ASB to explore an ASB-powered solution for this.

bourque:
I have sent an email to ITSD to ask for a production web server!

bourque:
The production web server has been built!  However, there are a few more things to take care of:

- [ ] Implement SSO layer / AD group authentication
- [ ] Change domain name to jwql.stsci.edu

bourque:
This is now complete!

PR #317, opened 2019-03-01
Pass base context through context processor

### LINKED ISSUE #273: Find better way to pass context to base HTML template
Currently, our `base.html` template requires some data to be passed to it, such as:

```python
context = {'inst': '',
           'inst_list': JWST_INSTRUMENT_NAMES,
           'tools': MONITORS,
            'user': user}
```

This means the each view defined in `views.py` requires this context to be gathered and passed.  This is a bit annoying and not very DRY.  Perhaps there is a more elegant solution.

bourque:
This PR adds `context_processor.py` and contains corresponding changes in `views.py` and `settings.py` to allow for the passing of `context` needed by the `base.html` template to occur in one function, instead of defining the same `context` in each view.

Closes #273


bourque:
@laurenmarietta Sorry to keep pouring on the review requests 😅

laurenmarietta:
(I see that #322 will restrict access further, I just thought that there was already a requirement to log into any page with the `auth_info` decorator...)

bourque:
@laurenmarietta This behavior is expected; we had set up the log in button and various authentication-related views to authenticate a user, but we hadn't restricted any part of the web app based off of that yet, so basically it has been a meaningless button.  But yes, indeed #322 will introduce some restricted areas of the web app, in anticipation of #294.

bourque:
@laurenmarietta You may be thinking of the period of time in which some of the views were broken because we were not passing the `user` information to them?  I think that was the case at some point in time, though my memory is also hazy.

Issue #318, opened 2019-03-04
Build function or module that resets the database


bourque:
Now that we have a dedicated sandbox database for development purposes, it would be convenient to have a means to reset the database (i.e. delete and remake all tables, leaving a database with empty tables).

PR #321, opened 2019-03-04
Reset database

bourque:
This PR adds a `reset_database.py` module, which will reset the database given in the `connection_string` key in the `config.json` config file.  This can be used to easily reset the development database.

### DIFF
--- a/jwql/database/database_interface.py
+++ b/jwql/database/database_interface.py
@@ -208,7 +208,7 @@ class Monitor(base):
     monitor_name = Column(String(), nullable=False)
     start_time = Column(DateTime, nullable=False)
     end_time = Column(DateTime, nullable=True)
-    status = Column(Enum('SUCESS', 'FAILURE'), nullable=True)
+    status = Column(Enum('SUCESS', 'FAILURE', name='monitor_status'), nullable=True)
     affected_tables = Column(postgresql.ARRAY(String, dimensions=1), nullable=True)
     log_file = Column(String(), nullable=False)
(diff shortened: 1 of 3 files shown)


bourque:
@hover2pi You've once thought about a database before, mind reviewing this one?

jwql/database/database_interface.py
```diff
@@ -208,7 +208,7 @@ class Monitor(base):
     monitor_name = Column(String(), nullable=False)
     start_time = Column(DateTime, nullable=False)
     end_time = Column(DateTime, nullable=True)
-    status = Column(Enum('SUCESS', 'FAILURE'), nullable=True)
+    status = Column(Enum('SUCESS', 'FAILURE', name='monitor_status'), nullable=True)
```
bourque: Note that this is a small bug fix that doesn't have anything to do with `reset_database.py`

laurenmarietta:
(Closes #318.)

bourque:
@laurenmarietta I envision that we could use this to completely wipe out and start from scratch with our development database, particularly in the case where we need to create new tables, or restructure existing tables, or have made mistakes in the kind of data we are storing.

This made me realize though that someone could use this to reset the production database, which would not be good.  I should add something in the code here that doesn't allow one to do that.

laurenmarietta:
Yes, I think that is a great idea! I was afraid to test it myself mainly because I was worried I would accidentally delete something I shouldn't... Including some checks would be helpful.

PR #322, opened 2019-03-04
Hide certain content if user is not authenticated

### LINKED ISSUE #312: Build function to hide exclusive access content in web app
Before we build an external facing production web server (#294), we should build in a convenient and secure way to hide content within the web app if the user is not authenticated.

bourque:
This PR contains modifications necessary to hide specific areas of the web app (namely the viewing of preview images and thumbnails) if the user is not authenticated.

Closes #312 
Pertains to #294


bourque:
@laurenmarietta I think this is ready for review, could you take a look?  

I went back and forth on allowing the API views to be rendered if the user is not authenticated.  I decided to keep those views open since I don't believe they contain any sensitive data (aside from filesystem paths, which we will have to solve in #336 before we go public (#294)).

bourque:
Note that this should not be merged until #332 is!

jwql/website/apps/jwql/templates/not_authenticated.html
```diff
@@ -0,0 +1,15 @@
+{% extends "base.html" %}
+
+{% block preamble %}
+
+    <title>About - JWQL</title>
```
laurenmarietta: Tiny change - this page should probably have an accurate name.

```suggestion
    <title>Not Authenticated - JWQL</title>
```

jwql/website/apps/jwql/oauth.py
```diff
@@ -177,46 +185,46 @@ def auth_info(fn):
...
+            return fn(request, user, **kwargs)
 
-        # If user is not authenticated, return no credentials
         else:
-            response = {'ezid' : None, "anon": True}
+            template = 'not_authenticated.html'
+            context = {'inst': '',
+                       'tools': MONITORS,
+                       'user': user,
+                       'version': jwql.__version__}
```
laurenmarietta: This isn't really necessary, but thanks to the neato context processors you added, this could just be:
```python
context = {'inst': ''}
```

jwql/website/apps/jwql/templates/not_authenticated.html
bourque: @laurenmarietta Good catch!

PR #323, opened 2019-03-05
Always output instrument from filename_parser

### LINKED ISSUE #308: JWQL web app home page example query error
When submitting the example program number `86600` in the search field of the web app home page, an error is raised: 
screen shot 2019-02-27 at 15 48 59

laurenmarietta:
Closes #308.

Turns out, the way that I was originally parsing the instrument out of a filename in `forms.py` only worked for certain kinds of filenames. After all of our updates to `filename_parser`, the instrument might appear in a filename in a variety of formats.

This PR adds another key to the output of `filename_parser` so that an `instrument` key is always returned.


laurenmarietta:
I'm going to merge this now so that bug isn't present for our DSMO demo today.

PR #338, opened 2019-03-14
Remove fail threshold in Jenkinsfile

bourque:
This PR removes that failure threshold in the `Jenkinsfile` in hopes to actually catch failed tests


bourque:
The fact that the build for be7d161 failed, wherein one test failed, indicates that this fix works I believe.  I pulled in @laurenmarietta's recent branch that fixes this failed test (it got lost in the switch to using a `develop` branch -- but that is a one-off issue)

PR #340, opened 2019-03-20
NIRSpec Data-Trending

DanielKuebi:
Basically the same as for MIRI Data-Trending, but more monitors and more trended mnemonics. 
Further information will follow soon.


bourque:
Unfortunately this PR includes changes that removes vital files (e.g. `WEBPAGE_TEMPLATE.md`, MIRI data trending materials, etc.) -- I'm not sure why.

We won't be able to merge this branch as-is.   One solution to fix this is to create a whole new branch and corresponding PR that just includes the NIRSpec-specific changes.  However, this would cause the additions to be made by me, and thus @DanielKuebi would not get the GitHub credit.

Another solution is to re-add the lost materials in a commit to this branch, which I think is the way to go.

PR #347, opened 2019-03-29
Don't return full paths in API views

bourque:
This PR modifies `data_containers.py` to only return basenames to thumbnails and preview images instead of full paths, as to no expose internal directory structure in API views.  It will be up to the user to reconstruct the full paths via the `filesystem` key in `config.json`.


bourque:
Thank you @codecov bot! 🤖

laurenmarietta:
Scratch that - the `test` directory is, but the files inside it aren't...

bourque:
@laurenmarietta Whoops!  I ran some `chmod` commands and this directory and its contents should be all set now.

Issue #254, opened 2019-01-22
Troubleshoot web app loading times


laurenmarietta:
The fact that pages on the web app can load really slowly has been a longstanding issue that we should address sooner rather than later.

While on the VPN from my house, I have been waiting 10 (!) minutes for one MIRI proposal worth of thumbnails to load. These times are prohibitive for any future users, and I'm sure they're because I did something wrong when I first set the web app up.

Maybe the Django debug toolbar will be able to help?

bourque:
@laurenmarietta I definitely agree that the django debug toolbar could be useful for this.  FYI, I attempted to get it installed, but can't seem to get it working (#249).  You might have better luck.

My hunch is that the hangup has something to do with iterating through the filesystem to grab the data to display.  If that is indeed the case, perhaps we can strive to get appropriate data from an `astroquery.mast` query instead.  Generally speaking, database queries are faster than file searches, especially over a network.

laurenmarietta:
Closed via #348.

PR #348, opened 2019-03-29
Fix slow loading times for some webpages

bourque:
This PR includes changes to speed up the loading times of some webpages, namely the `archived_proposals` and `thumbnails` views.

Changes include:
- Added test module that tests the loading times for various webpages and ensures that they load in under 30 seconds
- Added corresponding API docs for said test module
- Removed support for unlooked images altogether, as splitting the files into groups via their `EXPSTART` is what was causing much of the delay.  We should re-think how we want to define/implement the concept of "unlooked" images in a future issue/PR (likely we will want to use `astroquery` to perform a database query of new datasets, instead of doing FITS I/O operations).


bourque:
Closes #254

PR #354, opened 2019-04-02
Web app tests

### LINKED ISSUE #220: Write tests for web app

bourque:
This PR adds various tests for the web application.

Edit:  This PR partially addresses #220, but does not completely solve it.  We should still look into using available `django` functionality to write a more robust test suite for the web app.


bourque:
It was relatively straightforward to write tests for something like `data_containers`, since that module simply contains callable functions with limited passing of parameters.  It is not clear to me, however, how one should write tests for functions that depend on `request` objects (i.e. `view.py`, `oauth.py`, etc).  It seems to me that one would have to programmatically have to construct a meaningful `GET` requests, perhaps with the `requests` library or some sort of built in Django module.  

I dabbled in this a bit, trying something like:

```python
from django.test.client import Client

from jwql.website.apps.jwql import context_processors
from jwql.website.apps.jwql import oauth

client = Client()
request = client.get('{}/about/'.format(get_base_url()))
request = oauth.authorize(request)
context = context_processors.base_context(request)
```

however, I run into errors like:

```
django.core.exceptions.ImproperlyConfigured: Requested setting AUTHLIB_OAUTH_CLIENTS, but settings are not configured. You must either define the environment variable DJANGO_SETTINGS_MODULE or call settings.configure() before accessing settings.
```

or

```
AttributeError: module 'django.conf.global_settings' has no attribute 'ROOT_URLCONF'
```

or

```
django.core.exceptions.AppRegistryNotReady: Apps aren't loaded yet.
```

bourque:
Update:  I was able to get a test to run! (see 910aa97).  Unfortunately, I couldn't find a way to get this working within the `jwql.tests` subpackage with the rest of our tests.  Thus, it seems we may have to keep our package tests separate from our web app tests.

I will explore if it is possible to run the web app tests as part of our nominal Jenkins build.

bourque:
It may be possible to run the tests with Jenkins, but it will have to be the private Jenkins server, as the website needs `config.json` (https://ssbjenkins.stsci.edu/blue/organizations/jenkins/STScI%2Fjwql/detail/PR-354/6/pipeline/21#step-40-log-1)

jwql/tests/test_data_containers.py
```diff
@@ -0,0 +1,206 @@
...
+    assert len(proposals) > 0
+
+
+@pytest.mark.xfail
+def test_get_dashboard_components():
+    """Tests the ``get_dashboard_components`` function."""
+
+    dashboard_components, dashboard_html = data_containers.get_dashboard_components()
+    assert isinstance(dashboard_components, dict)
+    assert isinstance(dashboard_html, dict)
```
catherine-martlin: Do we care about testing if the dictionary has stuff in it or not or is testing if it exists enough?

catherine-martlin: There's a couple other times in here where we test for a dict but not if it's empty or not - if we decide to change that here we'd want to do that in the other tests too.

bourque: @catherine-martlin Good idea! I can add those checks, they would be nice to have.

bourque:
@catherine-martlin Removing `thumbnails()` should be fine here since it is not actually used anywhere in the web app anymore.  Much or all of that functionality has been ported over to `thumbnails_ajax()`, which provides the same data as `thumbnails()` did, only in an asynchronous fashion to allow for the loading screen to work.

Issue #144, opened 2018-09-11
Create private Jenkins server that has access to central storage


bourque:
We have several tests (and presumably many more to come) that require access to the `config.json` file.  However, since we cannot commit that file to the repository, we need a setup wherein Jenkins has access to the file via a central storage location.

bourque:
@SaOgaz is already working on this, but I just wanted to formally create a GitHub issue for it.

bourque:
This issue is not so much related to our code base, so I am going to move it to Jira.  See https://jira.stsci.edu/browse/JWQL-20

bourque:
Testing to see if I can successfully link this to the existing Jira issue.

https://jira.stsci.edu/browse/JWQL-20

SaOgaz:
Current status: @rendinam and @jhunkeler were looking to a way to get access to the config file on central store with a private jenkins run. Although the private run can access central store, we want a way to tell jenkins where that file is, without storing the directory location in our repository. This is the current blocker.

PR #356, opened 2019-04-03
Configure private jenkins tests

### LINKED ISSUE #344: Determine way to mark specific tests to only run on private Jenkins server
Now that we have a private Jenkins server that can access our `config.json` file on central storage (#203), we can now run the handful of tests that require that file.  However, we need to find a way to mark those test to only be triggered if the server is the private Jenkins server, and not the public one.

@SaOgaz suggests there is a way to do this in `pytest`, or perhaps we could build our own decorator.

bourque:
This PR includes several modifications to the test suite in attempt to only run certain tests on the private jenkins server (#144).  It also contains some minor aesthetic changes. 

Closes #344


bourque:
@bhilbert4 Would you mind reviewing this one? 

The important changes here is adding `@pytest.mark.skipif` decorator for tests that require the `config.json` file (and thus access to central storage) -- adding this will cause our public Jenkins builds to skip the test since it would inevitably fail.

There are some other changes too, mostly aesthetics, and removing `print` statements since they are not really needed within the Jenkins report.

jwql/tests/test_preview_image.py
```diff
@@ -51,26 +49,27 @@ def test_directory(test_dir=TEST_DIRECTORY):
...
     os.mkdir(test_dir)  # creates directory
     yield test_dir
-    print("teardown test directory")
     if os.path.isdir(test_dir):
         os.rmdir(test_dir)
 
 
+@pytest.mark.skipif(os.path.expanduser('~') == '/home/jenkins',
+                    reason='Requires access to central storage.')
 def test_make_image(test_directory):
```
bhilbert4: Do you need to add an `xfail` decorator here for Travis tests?

jwql/tests/test_utils.py
```diff
@@ -244,9 +244,8 @@
 
 ]
 
-
-@pytest.mark.xfail(raises=FileNotFoundError,
-                   reason='User must manually supply config file.')
+@pytest.mark.skipif(os.path.expanduser('~') == '/home/jenkins',
+                    reason='Requires access to central storage.')
```
bhilbert4: Should you keep the `xfail` decorator here for Travis tests?

jwql/tests/test_preview_image.py
bourque: @bhilbert4 I don't believe that `xfail` is needed here, as we would now like these tests to run and pass on the private jenkins server.  Using `skipif` here will skip the test entirely on the public jenkins server, which wouldn't be able to pass the test (as it requires access to central storage).

I'm not sure why are you mentioning Travis here(?) as we don't use Travis.

bourque:
@bhilbert4 Thanks for the review, I don't think there is anything else needed to be done on your part!

jwql/tests/test_preview_image.py
bhilbert4: Sorry. Too many repos with too many different CI testing setups to keep track of...

bourque: 😆 no worries

PR #357, opened 2019-04-08
Support stage 3 preview images and thumbnails

### LINKED ISSUE #309: Update generate_preview_images to handle different filename types
While testing `generate_preview_images.py` on the updated filesystem (#292), I notice that it chokes on certain filename types, e.g. `jw00307-o003_t001_niriss_f277w-clearp_i2d.fits` results in

```
02/27/2019 13:25:58 PM CRITICAL: Traceback (most recent call last):
  File "/home/jwqladm/repositories/jwql/jwql/utils/logging_functions.py", line 238, in wrapped
    func(*a, **kw)
  File "/home/jwqladm/repositories/jwql/jwql/utils/logging_functions.py", line 203, in wrapped
    func(*a, **kw)
  File "generate_preview_images.py", line 512, in generate_preview_images
    grouped_filenames = group_filenames(filenames)
  File "generate_preview_images.py", line 626, in group_filenames
    observation = filename_parts['observation']
KeyError: 'observation'
```

This is because this particular filename type does not have an `observation` component to it.

bourque:
This PR modifies `preview_image.py` and `generate_preview_images.py` to support stage 3 data products.

Closes #309


bourque:
@bhilbert4 Would you be able to review this before our sprint deadline?

jwql/jwql_monitors/generate_preview_images.py
```diff
@@ -26,7 +26,7 @@
         python generate_preview_images.py
 """
 
-from glob import glob
+import glob
```
bourque: @bhilbert4 I changed this to be a bit more consistent with the use of `glob` in our other modules

jwql/utils/preview_image.py
```diff
@@ -212,13 +212,6 @@ def get_data(self, filename, ext):
                     yd, xd = data.shape[-2:]
                     dq = np.ones((yd, xd), dtype="bool")
 
-                # Collect information on aperture location within the
```
bourque: @bhilbert4 This was crashing on stage 3 data products, as some of them don't have `SUBSTRT*` or `SUBSIZE*` in their headers.  But upon further investigation, these attributes didn't seem to be used anywhere?  So I just removed them altogether.  Do you know if these are needed?

bourque: @bhilbert4 Oh I see now, this actually is used for the creation of mosaics.  I should be able to get around this with some sort of parameter passing and/or `try/except`

bhilbert4:
Sure, I'll give this a look in the next couple days.

jwql/jwql_monitors/generate_preview_images.py
```diff
@@ -496,35 +497,151 @@ def find_data_channel(detectors):
...
+    See module docstring for further details."""
 
     # Begin logging
     logging.info("Beginning the script run")
 
-    filesystem = get_config()['filesystem']
-    preview_image_filesystem = get_config()['preview_image_filesystem']
-    thumbnail_filesystem = get_config()['thumbnail_filesystem']
+    # Process programs in parallel
+    program_list = [os.path.basename(item) for item in glob.glob(os.path.join(get_config()['filesystem'], '*'))]
```
bhilbert4: I like the idea of breaking it up by program and creating a separate thread for each one.

jwql/utils/preview_image.py
```diff
@@ -212,13 +212,18 @@ def get_data(self, filename, ext):
...
 
+
                 # Collect information on aperture location within the
                 # full detector. This is needed for mosaicking NIRCam
                 # detectors later.
-                self.xstart = hdulist[0].header['SUBSTRT1']
-                self.ystart = hdulist[0].header['SUBSTRT2']
-                self.xlen = hdulist[0].header['SUBSIZE1']
-                self.ylen = hdulist[0].header['SUBSIZE2']
+                try:
```
bhilbert4: So what happens in the case where these keywords are not present? (Is that for level 3 products?) The `xstart` and `ystart` attributes are used in `generate_preview_images`, line 305.

bourque: @bhilbert4 From what I can gather, these keywords are only missing in level 3 products.  And `xstart` and `ystart` only seem to be used when creating mosaics, so the question arises -- is there a need to create mosaics with level 3 products?

bhilbert4: I don't think so. At that point mosaics will have already been created for the appropriate data types. So I guess as long as `grouped_files` contains only a single file when it searches for a level 3 file basename, then `create_mosaic` won't be called. I suspect this will be true, but I'm not 100% certain.

bourque: @bhilbert4 Ok, I think this is the case.  We will go with this for now and deal with any possible bugs later.

PR #365, opened 2019-04-15
Improve test_preview_image

laurenmarietta:
I have been struggling with tests failing locally, and I traced it back to not enough things being deleted in the teardown script for `test_preview_image`. (It wasn't the permissions thing after all, @bourque!)

This PR:
- Adds an `if `statement to avoid errors for making the local test directory if it already exists
- Removes all files in the local test directory during teardown
- Removes all thumbnails/jpgs in the test directory on central storage during teardown (Though now I am less sure as to why having files there was causing problems... hopefully this doesn't break anything unexpected?)
- Parameterizes `test_make_image()`

### DIFF
--- a/jwql/tests/test_preview_image.py
+++ b/jwql/tests/test_preview_image.py
@@ -76,10 +76,15 @@ def get_test_fits_files():
     filenames : list
         List of filepaths to FITS files
     """
+    # Get the files from central store
     if not ON_JENKINS:
         filenames = glob.glob(os.path.join(get_config()['test_dir'], '*.fits'))
         assert len(filenames) > 0
         return filenames
+
+    # Or return an empty list
+    else:
+        return []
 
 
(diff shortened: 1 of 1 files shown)


jwql/tests/test_preview_image.py
```diff
@@ -47,15 +52,44 @@ def test_directory(test_dir=TEST_DIRECTORY):
         Path to directory used for testing
 
     """
-    os.mkdir(test_dir)  # creates directory
+    # Set up local test directory
+    if not os.path.isdir(test_dir):
```
bhilbert4: There is an `ensure_dir_exists` function in utils.py that does this. It also sets the permissions correctly on the directory if the directory needs to be created.

laurenmarietta: Ah, I completely forgot about that util. Thank you!

PR #371, opened 2019-04-19
Release 0.19.0

bourque:
This PR updates jwql `master` branch to version 0.19.0.

PR #370, opened 2019-04-19
Release 0.19.0

bourque:
This PR updates jwql `develop` branch to version 0.19.0.

PR #372, opened 2019-04-19
Redirect to page user was on upon login/logout

### LINKED ISSUE #334: Login/logout should not redirect back to homepage
Currently, the web app redirects to the homepage when clicking the login/logout button, regardless of which page the user was currently on.  We should switch this so that the page redirects to the page the user was on.

bourque:
This PR changes `oauth.py` so that upon login and logout, the user is redirected back to the page that they were on, instead of redirected back to the homepage.

Closes #334

### DIFF
--- a/jwql/website/apps/jwql/oauth.py
+++ b/jwql/website/apps/jwql/oauth.py
@@ -48,6 +48,8 @@ import jwql
 from jwql.utils.constants import MONITORS
 from jwql.utils.utils import get_base_url, get_config
 
+PREV_PAGE = ''
+
 
 def register_oauth():
     """Register the ``jwql`` application with the ``auth.mast``
@@ -118,7 +120,7 @@ def authorize(request):
     cookie_args['httponly'] = True
 
     # Set the cookie
-    response = redirect("/")
+    response = redirect(PREV_PAGE)
     response.set_cookie("ASB-AUTH", token["access_token"], **cookie_args)
 
     return response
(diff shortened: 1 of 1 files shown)


bourque:
Note that this should not be merged until after release 0.19 (#370, #371)

bourque:
@laurenmarietta Would you be able to review this?  Now that #371 is done this can be reviewed/merged now.

jwql/website/apps/jwql/oauth.py
```diff
@@ -266,7 +271,7 @@ def logout(request):
         Outgoing response sent to the webpage
     """
 
-    response = redirect("/")
+    response = redirect(PREV_PAGE)
```
laurenmarietta: I found that this change didn't let me log out if I hadn't logged in on the same web session. When I clicked the "log out" button, I would just get an error.

But I fixed it by just adding the same lines from `login()` to the `logout()` function, e.g.:
```python
global PREV_PAGE
PREV_PAGE = request.META.get('HTTP_REFERER')
response = redirect(PREV_PAGE)
```
Is this an okay change to make? I don't really understand global variables still.

bourque: @laurenmarietta Nice catch.  Yeah, this should be fine to do, and I think it makes sense why it broke for you in that situation; in that case the `PREV_PAGE` variable wouldn't have been set in the global scope.

PR #373, opened 2019-04-19
Fix passing Jenkins builds despite failed unit tests

### LINKED ISSUE #364: Jenkins builds still passing despite numerous failed unit tests
It seems that Jenkins builds are passing even though there may be several failed unit tests, for example:

https://ssbjenkins.stsci.edu/blue/organizations/jenkins/STScI%2Fjwql/detail/PR-322/11/pipeline/21/#step-39-log-21

#338 attempted to fix this, but it doesn't seem to be having an effect afterall.

bourque:
This PR attempts to configure the `Jenkinsfile` such that Jenkins builds will fail upon failed unit tests.

Closes #364


bourque:
@bhilbert4 Would you be able to review this?

bhilbert4:
Sure, I'll give it a look in the next day or two.

jwql/tests/test_loading_times.py
```diff
@@ -48,6 +49,8 @@
     urls.append('{}/{}/'.format(instrument, rootname))
 
 
+@pytest.mark.skipif(os.path.expanduser('~') == '/home/jenkins',
+                    reason='Requires access to central storage.')
```
laurenmarietta: Just looking at what part of this test requires access to the `config.json`, it's just getting the admin account username. Is that too much of a security risk that we couldn't just hard-code it into `get_base_url()`?

bourque: @laurenmarietta I'm not 100% sure if that would or wouldn't break any sort of security policy, so I'm inclined to err on the side of caution 🤷‍♂️

laurenmarietta: Okay, makes sense!

jwql/tests/test_dark_monitor.py
```diff
@@ -70,6 +72,8 @@ def test_get_metadata():
     assert monitor.frame_time == 10.5
 
 
+@pytest.mark.skipif(os.path.expanduser('~') == '/home/jenkins',
```
bhilbert4: Can the Jenkins server not query MAST for some reason? I don't understand why this needs to be skipped on Jenkins. Information from the config.json file isn't used, and it doesn't rely on access to central store.

jwql/tests/test_dark_monitor.py
```diff
@@ -122,6 +128,8 @@ def test_noise_check():
     assert np.all(noisy[1] == np.array([3, 9]))
 
 
+@pytest.mark.skipif(os.path.expanduser('~') == '/home/jenkins',
```
bhilbert4: Similar for this test. It seems like it doesn't require any information that isn't available when running on Jenkins.

jwql/tests/test_dark_monitor.py
bourque: @bhilbert4 See https://ssbjenkins.stsci.edu/blue/organizations/jenkins/STScI%2Fjwql/detail/PR-373/12/pipeline/21#step-39-log-19

I think the `dark_monitor` tests are failing because the tests need to instantiate the `Dark()` class (i.e. `dark_monitor.Dark(testing=True)`), and in doing so, needs to configure logging (since `Dark()` is decorated with `log_info` and `log_fail`), and since `logging_functions` need access to the config file, it ultimately fails.

I will look into trying to find a way around this.

bourque:
@bhilbert4 I think i've found a good workaround for the `dark_monitor` tests, which involved refactoring the `__init__` method so that the `Dark` class could be instantiated without having to deal with logging.  I think this is the way we should probably move forward with future monitors any way -- it is kind of bad practice to have the `__init__` method doing a lot of the processing.  

Could you take a look at my changes again?  Here is the diff between now and you last review: https://github.com/spacetelescope/jwql/pull/373/files/3467de90b23ad7ff07448062067b658a815513ab..5f55c29e8eb73a25845e688dd1ad511ed82b0504

PR #379, opened 2019-04-22
Fix dashboard

### LINKED ISSUE #375: Dashboard error
As @cracraft reminded me, the JWQL dashboard page is throwing an error again.

It looks like there is a new directory in the `outputs` directory that it doesn't know how to handle: `miri_total_file_count_by_type`.

We should probably make the `get_dashboard_components()` function more robust to these kinds of problems.

laurenmarietta:
Closes #375.

This PR fixes the bug where the dashboard page does not load, throwing an error.

I got around it by adding checks in `data_containers/get_dashboard_components()` to only attempt to load HTML/JavaScript files from known monitors. I also added a feature to programmatically generate formatted plot names from the component files.

This means that a developer adding a new monitor to the dashboard page will need to explicitly add the monitor to `name_dict` in `data_containers/get_dashboard_components()` for it to be loaded. However, once a monitor is added to that dictionary, every plot in the output directory for that monitor will be shown.

### DIFF
--- a/jwql/website/apps/jwql/data_containers.py
+++ b/jwql/website/apps/jwql/data_containers.py
@@ -145,44 +145,60 @@ def get_dashboard_components():
     """
 
     output_dir = get_config()['outputs']
-    name_dict = {'': '',
-                 'monitor_mast': 'Database Monitor',
-                 'database_monitor_jwst': 'JWST',
-                 'database_monitor_caom': 'JWST (CAOM)',
-                 'monitor_filesystem': 'Filesystem Monitor',
-                 'filecount_type': 'Total File Counts by Type',
-                 'size_type': 'Total File Sizes by Type',
-                 'filecount': 'Total File Counts',
-                 'system_stats': 'System Statistics'}
-
-    # Exclude monitors that can't be saved as components
-    exclude_list = ['monitor_cron_jobs', 'miri_data_trending',
-                    'trainings_data_15min', 'trainings_data_day']
+    name_dict = {
+        '': {
+            'name': '',
+            'plots': {}
+        },
+        'monitor_mast': {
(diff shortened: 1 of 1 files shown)


cracraft:
Is there a way to view the finished results, or do I just have to read code?

bourque:
@cracraft I think you should be able to see it in action by running the web app locally:

```
cd jwql/jwql/website/
python manage.py runserver
```
then visit `http://127.0.0.1:8000/` in a browser.

laurenmarietta:
You would have to switch into this branch first, though. This should do it:
```
git checkout -b laurenmarietta-fix-dashboard develop
git pull https://github.com/laurenmarietta/jwql.git fix-dashboard
```

cracraft:
Thanks for the instructions. The plots are displaying now, but we've got some odd behavior, and I don't know if that goes in this ticket or not. Some plots show a long history of queries (many dates in the plots) and others show only single points. Would that have happened as part of this update?

bourque:
@cracraft It could be that some of the plots only have a single point because the filesystem monitor hasn't been running every day; I can look into that.  The fact that plots are rendering should indicate that this PR is correctly fixing the issue at hand though.

cracraft:
While the plots haven't been running every day, they have been tested before, so I assumed the data would still be there. (unless part of the database got reset?)

bourque:
@cracraft Yeah, since the database schema was overhauled in #328, the database had to be reset

cracraft:
Ok, then I'd say that since I can now see the plots in this branch, it fixes the problem.

bourque:
Thanks @laurenmarietta, this looks good and works well on my end.  And it looks like the filesystem monitor plots are now updating with new data points now that #328 is in `master`!

Issue #283, opened 2019-02-07
Build custom 404 page


bourque:
Whenever we encounter an error in the web app, we are presented with a Django page with a lot of information that perhaps is informative for developers but not really necessary for users, and as @laurenmarietta pointed out, perhaps displaying some of this information is insecure.  We could improve this experience by having a custom 404 page.

bourque:
@spacetelescope/jwql We could get fun and creative here.  Anyone is welcome to contribute a 404 page, and if we come up with multiple 404 pages, we could randomly pick a template to render upon request.

bourque:
Example from `acsql`: https://acsql.stsci.edu/foo

PR #382, opened 2019-04-22
Custom 404 Page - Lost in Space!

laurenmarietta:
I heeded the call in #283 and designed a custom 404 page for the JWQL project!

This PR includes that page, as well as the infrastructure to randomly select an HTML template and render that in response to a HTTP 400, 403, 404, or 500 error.

Note that **the page will only appear locally if the `DEBUG` variable within `settings.py` is set to `False`, and then the server is run with the command `manage.py runserver --insecure`**. Thanks, [StackOverflow](https://stackoverflow.com/questions/5836674/why-does-debug-false-setting-make-my-django-static-files-access-fail)!


jwql/website/apps/jwql/static/css/jwql.css
```diff
@@ -386,6 +386,27 @@ display : inline;
   margin-bottom: 0;
 }
 
+/*Video for space 404 page*/
```
bourque: :dizzy: magic 💫

jwql/website/apps/jwql/templates/404_space.html
```diff
@@ -0,0 +1,34 @@
...
+    <title>404 Not Found - JWQL</title>
+
+{% endblock %}
+
+{% block content %}
+
+    <main role="main" class="container">
+
+        <!--Add JavaScript to make text fade in after video loads-->
+        <script>
```
bourque: :dizzy: moar magic :dizzy:

jwql/website/jwql_proj/urls.py
```diff
@@ -49,6 +49,13 @@
 from django.contrib import admin
 from django.urls import include, path
 
+from ..apps.jwql import views
+
+# Define custom error page views
```
bourque: Neat!

PR #387, opened 2019-04-25
Updated jenkins file with new build matrix

### LINKED ISSUE #271: Consider adding new OSes to Jenkins
Currently, our test suite is just being run for Linux. We should consider adding more builds that test different operating systems - and maybe even versions of Python!

obviousrebel:
Added build matrix with python version 3.5 and 3.6. This involved adding a separate environment file for python 3.5 and adopting a naming convention to differentiate between py35 and py36 environment files.

Only one OS is in the matrix (linux-stable) as SSB currently doesn't have any OSx nodes. I am working to see if it is possible to have one added and then this can be updated to include OSx in the build matrix.

Closes #271


laurenmarietta:
Thanks for taking this on, @obviousrebel! This PR will close #271.

Any idea why the Jenkins build is continuing on 30 minutes now, though?

obviousrebel:
@laurenmarietta I'm looking at that now, the install command has the -q flag, might have to unset that to see whats going on but I imagine its downgrading all the packages to versions compatible with py3.5

obviousrebel:
It seems that the pinned versions of some packages in the environment.yml file are causing issues when switching python versions as they may be unavailable for that version. I suggest we create an additional environment.yml file for the python 3.5 packages. Would this be acceptable?

laurenmarietta:
Sure, if the `conda install` command changes you're working on don't pan out, I think it would be fine to make a separate environment file. 
We can just update the setup docs so people know they have a choice between 3.5 and 3.6 when they're creating their conda environment.

obviousrebel:
Okay, everything looks good now. May want to decide on better names for the environment files though.

laurenmarietta:
@obviousrebel just in case this comes up in the future: I ran into a similar problem when trying to add more builds to the [JWST MAGIC repo](https://github.com/spacetelescope/jwst_magic). We don't have pinned versions, but it did take an egregious amount of time to update the environment to Python 3.5.

I ended up getting around it by _first_ creating an environment with the specified Python version, _then_ updating it given the contents of the `environment.yml` file. That brought the total time down from 40 minutes to 7! 🎉 

Here's the pipeline, for reference:
https://ssbjenkins.stsci.edu/blue/organizations/jenkins/STScI%2Fjwst_magic/detail/JWSTFGS-156-python-builds/7/pipeline/23

PR #377, opened 2019-04-22
dark_monitor for all instruments

bourque:
This PR makes appropriate changes to the `dark_monitor` to get it working for all JWST instruments. 

Closes #374

Issue #292, opened 2019-02-13
Get access to updated filesystem


bourque:
Since the start of this project, we have been using a copy of the b-string data that DMS used for their development and testing.  However, DMS has moved on to the d-string data.  It would be helpful if we could get access to this; perhaps there are updated data products that more closely reflect what we will see during commissioning/operations. 

I have reached out to Katie Kaleida and Mary Romelfanger on the matter.

bourque:
Katie has pointed me to a new filesystem.  I will make a copy of it in our central store area and try running some of our software over it to see if anything breaks.

bourque:
Things that are broken include #308 , #309

bourque:
I will close this ticket since we now have a copy of an updated filesystem.  Any further issues pertaining to its contents can go through separate issues/PRs.

PR #397, opened 2019-05-03
Skip tests

### LINKED ISSUE #363: Make test xfails into skips
We have a lot of `pytest.mark.xfail` tests in our suite that I think should be converted into skips -- like @bourque added in #356: https://github.com/spacetelescope/jwql/blob/f3d4a3303fa2c786df21937c8cf7831c3e63b20c/jwql/tests/test_preview_image.py#L56

In my opinion, it is more accurate to skip tests on the Jenkins server if they can't be run there, rather than marking them as `xfail`. I think some of those tests really should be passing on our local machines, but marking them `xfail` makes it hard to tell if something has actually broken.

laurenmarietta:
Closes #363.

This PR:
- Rewrites tests in our test suite to use `skip` instead of `xfail`, so that the test suite is more accurate when run both on Jenkins and locally
- Updates `test_api_views.py` so that the API tests are run on the dev server during Jenkins builds
- Renames one of the file roots in `test_api_views` that was removed from the filesystem

Note that there is a temporary `if` statement in `test_api_views` that skips any test that throws the "502 Bad Gateway" error (#407) . This should be removed in the future.


laurenmarietta:
Two issues I've found:

1) The Python 3.5 Jenkins build is throwing an assertion error - see #398.

2) `test_api_views()` is failing for seemingly random views. I also see this when visiting the pages on the dev server, e.g. https://dljwql.stsci.edu/api/nircam/thumbnails/ returns:
Screen Shot 2019-05-03 at 4 29 54 PM

I restarted the dev server but that didn't help. 🤷‍♀

bourque:
@laurenmarietta In regards to your concerns, I think we've got a handle on (2) with #407, and as for (3), I think you're right --the tests in `test_api_views` were written based on the contents of the filesystem before the recent update (#292) and so some of the hard coded program numbers and rootnames may be outdated.  I can update these in this PR.  

As to why these files were in the previous filesystem but not the new one ... 🤷‍♂ 

I am planning to reach out to the archive folks soon with some questions related to the filesystem contents (see comments in #377), I can bring this concern up then.

jwql/tests/test_api_views.py
```diff
@@ -17,14 +17,25 @@
...
-import urllib.request
+from urllib import request, error
 
 from jwql.utils.utils import get_base_url
 from jwql.utils.constants import JWST_INSTRUMENT_NAMES
 
+# Determine if tests are being run on jenkins
+ON_JENKINS = os.path.expanduser('~') == '/home/jenkins'
+
+# Determine if the local server is running
```
bourque: @laurenmarietta Clever!

jwql/tests/test_loading_times.py
```diff
@@ -49,11 +52,10 @@
     urls.append('{}/{}/'.format(instrument, rootname))
 
 
-@pytest.mark.skipif(os.path.expanduser('~') == '/home/jenkins',
-                    reason='Requires access to central storage.')
+@pytest.mark.skipif(ON_JENKINS, reason='Requires access to central storage.')
 @pytest.mark.parametrize('url', urls)
 def test_loading_times(url):
-    """Test to see if the given ``url`` returns a webpage sucessfully
+    """Test to see if the given ``url`` returns a webpage successfully
```
bourque: @laurenmarietta Thanks for catching all of my spelling mistakes 😆

PR #401, opened 2019-05-07
central store update to filesystem monitor 

cracraft:
So this code is only functional up to a point. Right now, it's crashing on the add to database section for the filesystem monitor. The counting goes smoothly, just takes a long time to run. I had attempted to introduce logic that would record which directories were already counted, then sum only those that had not been counted and add that for the 'all' category, but that failed, so I stripped it out. I will update if I get those working, but for now, there is duplicate file 'walking' for the area all. I have not been able to test the added plotting, since the database update portion is failing. If anyone has a solution for that part, I'd love to hear it. The log monitor_filesystem_2019-05-07-09-17.log contains the crash information.


bourque:
@bourque You should look into this RE: database issues

jwql/jwql_monitors/monitor_filesystem.py
```diff
@@ -133,15 +137,82 @@ def get_global_filesystem_stats(general_results_dict):
...
+        # to put in TB, have to multiply values by 1024 to get in bytes, then
+        # divide by 1024 ^ 4 to put in TB
+        total = int(stats[0]) / (1024 ** 3)
+        free = int(stats[2]) / (1024 ** 3)
+        central_storage_dict[area]['size'] = total
+        central_storage_dict[area]['available'] = free
+
+        # do an os.walk on each directory to count up used space
+
+        for dirpath, _, files in os.walk(fullpath):
```
bourque: @cracraft As you elude to in your comments, this is definitely the part that is taking the most time.  I wonder if it would be faster to do a `du -s` command here and use `subprocess.check_output` to get the results (akin to what you do for `df`), instead of measuring each file individually.  A crude test on my end shows that running `du` on the `preview_images` directory takes ~28 minutes to complete.  Considering `thumbnails` would be about the same size of directory, and the other directories are much smaller, I would reckon that the whole process would take roughly an hour.

Does that seem faster than what is currently implemented?

cracraft: I did a test of the code as it stands, and the file walking method took about half an hour to get through logs, outputs, test, preview_images, and thumbnails. It then took another hour to calculate for jwql, the upper level directory, that was also double-counting the earlier mentioned directories. I'd like to figure out how to avoid the double-counting that it's doing, and I think it would speed up quite a bit.

jwql/jwql_monitors/monitor_filesystem.py
```diff
@@ -165,7 +238,10 @@ def initialize_results_dicts():
     instrument_results_dict = {}
     instrument_results_dict['date'] = now
 
-    return general_results_dict, instrument_results_dict
+    central_storage_dict = {}
+    central_storage_dict['date'] = now
```
bourque: @cracraft I think this is the reason that the code is crashing when trying to update the database -- because the database table for the `central_storage` table employs a `unique=True` condition on the `date` column, we won't be able to add multiple records when the `date` is the same for each record, as it is initialized as such here.

I think you could solve this by simply removing the `unique` constraint on the `date` column.  I think this will be fine since we will be running this script via a cron job that executes once a day.

bourque: FYI, here is the error being thrown:

```
06/03/2019 15:10:23 PM CRITICAL: Traceback (most recent call last):
  File "/Users/bourque/anaconda2/envs/jwql/lib/python3.6/site-packages/sqlalchemy/engine/base.py", line 1244, in _execute_context
    cursor, statement, parameters, context
  File "/Users/bourque/anaconda2/envs/jwql/lib/python3.6/site-packages/sqlalchemy/engine/default.py", line 552, in do_execute
    cursor.execute(statement, parameters)
psycopg2.IntegrityError: duplicate key value violates unique constraint "central_storage_date_key"
DETAIL:  Key (date)=(2019-06-03 11:32:43.48001) already exists.
```

cracraft: I have removed the unique constraint and tested it again and still get the same error. I'm not sure what's going on with this.

bourque: @cracraft I can try testing this out and digging into it.  I'll report back.

bourque: @cracraft It took 15 hours to run, but it appears my latest changes (https://github.com/spacetelescope/jwql/pull/401/commits/e7355147d5eff29f7f35cdd3207fd8564b7d543e) results in a successful completion of the filesystem monitor.  Could you try pulling these changes and testing on your end?  If it works on your end, this PR should be ready to go.

cracraft: I didn't pull your edits over before getting back to my testing, but pretty much have the same edits in place and have the code running. I've added more logging so we can see how long it's taking to search each directory and see where the time sink is in this. I've also started adding some code to reduce the duplicate counting, so I'm not ready to commit it just yet.

cracraft:
I had a successful test run of this version. The code runs in 2.5 hours on my computer and successfully logs all the paths so we can see where the longest time is spent on counting as well as creating all requested plots. I'm ready to merge this version once it has been reviewed. I still have concerns about how the counting is happening for the filesystem, as the total space shown on the plot is smaller than both available and used. That's due to the total size being the actual files in the system while used and free refer to the full 8 TB file system in central store. We may want to re-discuss what we want to show and how best to represent that, as the current plot will confuse people. But this is good for now, so far as I can tell.

bourque:
@bourque review this

bourque:
@cracraft I tested this out in the dev environment and everything appears to work well!  So I think this PR can now be merged.

It does seem like the 'free' and 'used' sizes recorded for the central storage area to appear as 0TB sometimes -- we can address that bug in a future PR.  I would like to merge what we have here so we can get it into our next release.

PR #402, opened 2019-05-08
Remove EDB interface code

### LINKED ISSUE #368: remove EDB interface code from jwql package
The code ported from jwql to https://github.com/spacetelescope/jwst-dms-edb should be removed and replaced by imports of `jwst-dms-edb`

Johannes-Sahlmann:
Add dependency on jwedb package  (https://github.com/spacetelescope/jwst-dms-edb)

Closes #368


jwql/tests/test_edb.py
```diff
@@ -15,16 +15,12 @@
 
     ::
 
-        pytest -s test_edb_interface.py
+        pytest -s test_edb.py
```
laurenmarietta: I think this filename change should also be updated in `docs/source/test.rst`?

Johannes-Sahlmann: yes, I missed that. thank you!

Issue #403, opened 2019-05-08
Jenkins build failing 


Johannes-Sahlmann:
The build of #402 is failing which seems to be unrelated to the changes it introduces.
`CondaEnvException: Conda Env Exception: Unable to determine environment`

https://ssbjenkins.stsci.edu/blue/organizations/jenkins/STScI%2Fjwql/detail/PR-402/3/pipeline

obviousrebel:
I believe this is related to the latest commit to the jenkins shared utils library https://github.com/spacetelescope/jenkins_shared_ci_utils/commit/61fdd601b48119d4e18329dd3d2c290f69ef2850
CC @rendinam

rendinam:
Build 402 has been restarted after rolling back the utils library to the previous version.

Johannes-Sahlmann:
Thanks. the build has passed.

PR #404, opened 2019-05-10
Explicitly request use of conda in Jenkinsfile

rendinam:
This change causes the creation of a new conda environment used to host the build as per the `conda_packages` docs here: https://github.com/spacetelescope/jenkins_shared_ci_utils/blob/master/README.md and removes the `with_env` commands which are then no longer necessary.

The error noted in issue #403 was caused by the Jenkinsfile not requesting the use of conda (by populating the `conda_packages` list for a given build configuration). A leftover pre-installed conda in the job's container was used by default, which allowed the `conda` commands specified in the build configuration to execute, though on a conda base environment, rather than a new target environment, which is typically not desired.

As an additional safeguard, future build containers on the Jenkins host will do away with the built-in conda installation to make the behavior more consistent.


laurenmarietta:
Thanks so much for submitting this. Good to know how not setting `conda_packages` could lead to unexpected behavior!

I'll wait until @bourque is back to merge this, so he can see your note.

PR #410, opened 2019-05-14
Edb access via web app login

### LINKED ISSUE #329: Make EDB interface work with either token authentication or website login
I tried to perform an EDB mnemonic query from the development server. I first logged in using 2fA and then attempted the query. It failed (see attached) because it is expecting a valid MAST token to be present in the json.config. 

The web app should be able to allow both ways of authenticating for MAST in a transparent way.

screen shot 2019-03-06 at 15 32 50

Johannes-Sahlmann:
This PR introduces changes to the web app code that allow MAST authentication for EDB queries (and any other service provided by astroquery.mast) when the user has logged into the web app.

closes #329

### DIFF
--- a/jwql/website/apps/jwql/data_containers.py
+++ b/jwql/website/apps/jwql/data_containers.py
@@ -29,8 +29,17 @@ import tempfile
 
 from astropy.io import fits
 from astropy.time import Time
-from astroquery.mast import Mast
+
 import numpy as np
+
+# astroquery.mast import that depends on value of auth_mast
+# this import has to be made before any other import of astroquery.mast
+from jwql.utils.utils import get_config, filename_parser
+mast_flavour = '.'.join(get_config()['auth_mast'].split('.')[1:])
+from astropy import config
+conf = config.get_config('astroquery')
+conf['mast'] = {'server': 'https://{}'.format(mast_flavour)}
+from astroquery.mast import Mast
 
 from jwedb.edb_interface import mnemonic_inventory
 from jwql.edb.engineering_database import get_mnemonic, get_mnemonic_info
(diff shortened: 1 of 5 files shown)


Johannes-Sahlmann:
@laurenmarietta @bourque to make this work with the mast dev server I had to make the imports in `data_containers.py` in a specific, non-compliant order. Maybe there is a more elegant way to do this that I do not know about.

jwql/website/apps/jwql/data_containers.py
```diff
@@ -29,19 +29,28 @@
 
 from astropy.io import fits
 from astropy.time import Time
-from astroquery.mast import Mast
+
 import numpy as np
 
+# astroquery.mast import that depends on value of auth_mast
```
bourque: @Johannes-Sahlmann I think this is fine.  It's not a problem to stray from conventions as long as there is a good reason and it is well documented, as is the case here.

jwql/website/apps/jwql/views.py
```diff
@@ -287,8 +287,8 @@ def dashboard(request):
 
     return render(request, template, context)
 
-
-def engineering_database(request):
+@auth_info
```
bourque: @Johannes-Sahlmann Nice use of `@auth_info`!

laurenmarietta:
I am revisiting this in the context of #422, and realized I am quite confused: is the token that we are now getting upon log-in (`access_token`) the same as the MAST token that we are supposed to manually put in our `config.json` and/or set as an environment variable? Or a different token?

I thought it had to be the same token, since it enables you to perform the EDB query just with log-in, but the two tokens don't match for me.

Johannes-Sahlmann:
The two tokens are different (by design I believe).

Issue #398, opened 2019-05-03
test_get_pipeline_steps failing on 3.5 Jenkins


laurenmarietta:
Seems as though the new Python 3.5 Jenkins build added in #387 doesn't quite succeed.

The assertion in `test_get_pipeline_steps()` in `test_pipeline_tools.py` is failing:
https://ssbjenkins.stsci.edu/blue/organizations/jenkins/STScI%2Fjwql/detail/develop/17/pipeline/22

I assume (hope!) this is a simple version issue.

(Also, why isn't this failing the Jenkins build??? The question that never goes away...)

obviousrebel:
The failing test I believe is not failing the build because Jenkins is not ingesting the results.xml file generated by pytest as it states it is invalid junit format.

In order for pytest to create junit valid xml reports, a pytest.ini file must be present to specify the junit family (which is present in this repo). Unfortunately, pytest added this feature in 4.2 and the latest available version of pytest for python 3.5 is 3.8.1.

Jenkins is complaining that
```
Attribute 'line' is not allowed to appear in element 'testcase'.
```

Which is what makes the result.xml file invalid (to jenkins). One 'hacky' way around this would be a script or sed command that runs on only the python 3.5 build to regex match and remove that attribute (line=*) from all test cases in the results.xml file to appease Jenkins and allow processing of the test results.

bourque:
#412 will hopefully solve the false Jenkins reporting, but we will still need to figure out why `test_get_pipeline_steps` is failing (e.g. https://ssbjenkins.stsci.edu/blue/organizations/jenkins/STScI%2Fjwql/detail/PR-410/3/pipeline/22#step-147-log-34)

bourque:
I have been able to reporoduce the failing test locally, and I believe the issue is that for some reason python 3.5 stores the `req_steps` in a different order than the `steps_dict`, and since the `assert` statement is comparing two `OrderedDict`s, the `assert` fails.  For example:

```
        print(instrument)
        print(req_steps)
        print(steps_dict)
```

returns:

```
OrderedDict([('dq_init', True), ('saturation', True), ('superbias', True), ('refpix', True), ('linearity', True), ('persistence', True), ('dark_current', True), ('jump', True), ('rate', True), ('firstframe', False), ('ipc', False), ('lastframe', False), ('group_scale', False), ('rscd', False)])
[...]

PR #412, opened 2019-05-14
Add sed command to remove invalid attrs from junit xml

obviousrebel:
Addresses an issue found in #398 (this does not address the specific failing test, only the issue where the failing test is not failing the build)

Pytest failures are not failing the build for python 3.5. This was due to the latest version of pytest for python 3.5 is pytest 3.8.1. Junit family config option was added in pytest 4.2 which is needed to make the pytest junit xml ingestible by jenkins.

Added sed command to (hopefully) manually strip unknown attributes from the xml to please jenkins.

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -23,7 +23,7 @@ withCredentials([string(
           "python setup.py install"]
       bc.test_cmds = [
           "pytest -s --junitxml=results.xml --cov=./jwql/ --cov-report=xml:coverage.xml",
-          "sed -i 's/file=\".*\"//g;s/line=\".*\"//g' results.xml",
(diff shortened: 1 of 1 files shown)


obviousrebel:
Okay, I believe python 3.5 failed tests are now correctly making the build unstable. get_pipeline_steps test is failing and now the build is correctly showing unstable.
The failed jenkins build is in this case a good thing @bourque

laurenmarietta:
!!! 😲So awesome that you got this to work! Thank you!!

bourque:
Thanks @obviousrebel! Looks great! (or as great as regex could possibly look 😅)

PR #415, opened 2019-05-14
Add download buttons to EDB page 

### LINKED ISSUE #302: Add download button(s) to EDB web page
Users may want to download the results of their manual exploration of the EDB mnemonics. Appropriate functionality for download should be added.

Johannes-Sahlmann:
This PR introduces functionality in the web app code that allows a user to download the results of EDB mnemonic queries.

Closes #302


jwql/website/apps/jwql/data_containers.py
```diff
@@ -266,18 +268,28 @@ def get_edb_components(request):
...
                     display_table.write(tmp, format='jsviewer')
-                mnemonic_exploration_result.html_file = path
-                mnemonic_exploration_result.html_file_content = open(path, 'r').read()
+                mnemonic_exploration_result.html_file_content = open(path_for_html, 'r').read()
+
                 # pass on meta data to have access to total number of mnemonics
                 mnemonic_exploration_result.meta = meta
 
+                # save file locally to be available for download
+                STATIC_DIR = os.path.join(settings.BASE_DIR, 'static')
```
bourque: @Johannes-Sahlmann I don't think this is actually used as a global variable?  And so I think it could be named with lowercase instead of uppercase per PEP8.

jwql/website/static/mnemonic_exploration_result_table.csv
```diff
@@ -0,0 +1,14 @@
+description,sqlDataType,subsystem,tlmIdentifier,tlmMnemonic,unit
```
bourque: @Johannes-Sahlmann Is there a reason to commit this data file to the repository?

bourque:
@Johannes-Sahlmann Sorry I may have jumped the gun on this, I just realized that this PR is marked as still in draft stage (I still haven't been able to figure out how to do this myself 😮)

jwql/website/apps/jwql/data_containers.py
Johannes-Sahlmann: thanks. I changed it to lower case.

jwql/website/static/mnemonic_exploration_result_table.csv
Johannes-Sahlmann: this was an accidental commit. I removed it.

Johannes-Sahlmann:
I spent quite some time and several unsuccessful attempts to find an elegant solution to serve the files for download (using temporary files and MEDIA_ROOT, MEDIA_URL combinations, and other possible solutions I found online). 
The only way I got this to work was to use the existing static data directory. The EDB web page is writing two files there (when download is request) that always have the same names and will therefore be overwritten.
I would be happy to learn about other solutions, but for now this is functional code that satisfies the basic requirements.

laurenmarietta:
Hmm. That question of how to handle the file is tricky... The solution you present here is definitely a good one - the only foreseeable concern I see is what happens if two people are trying to download different data at almost the same time. Since both users would be pointed to the same file, one of them might end up downloading data from a query someone else performed slightly more recently. (Clearly this is an edge case and very well might never happen, but still.)

The only alternative option I can think of at the moment might be to save the files with unique names based on the query time on central store, e.g. somewhere like `jwql/temp_edb_files/`. Then there could be a cron job that cleans out that directory every day, hour, five minutes, etc. That is certainly more complicated than the current solution, though, so I'm not sure if it would really be worth the trouble... Thoughts?

jwql/website/apps/jwql/data_containers.py
```diff
@@ -281,18 +308,30 @@ def get_edb_components(request):
...
-                mnemonic_exploration_result.html_file_content = open(path, 'r').read()
-                # pass meta data to have access to number of mnemonics
+                mnemonic_exploration_result.html_file_content = open(path_for_html, 'r').read()
+
+                # pass on meta data to have access to total number of mnemonics
                 mnemonic_exploration_result.meta = meta
 
+                # save file locally to be available for download
+                static_dir = os.path.join(settings.BASE_DIR, 'static')
+                if not os.path.isdir(static_dir):
```
laurenmarietta: Both here and at line 268, this `if` statement can be replace with `utils.ensure_dir_exists()`.

jwql/website/apps/jwql/views.py
```diff
@@ -309,7 +310,7 @@ def engineering_database(request, user):
         Outgoing response sent to the webpage
 
     """
-    edb_components = get_edb_components(request)
+    edb_components = get_edb_components(request, settings)
```
laurenmarietta: Just curious: is there a reason why the settings need to be imported in `views.py` and passed into this function? Rather than just imported in `data_containers.py`?

jwql/website/apps/jwql/data_containers.py
Johannes-Sahlmann: done, thanks.

jwql/website/apps/jwql/views.py
Johannes-Sahlmann: Probably a leftover from developing the code. I moved this as suggested.

Johannes-Sahlmann:
@laurenmarietta The cron job solution would be fine with me, but I will need help to set it up and test it. Should we defer this to a future issue/PR ?

laurenmarietta:
Leaving that for a future issue sounds good to me, yup!

Issue #406, opened 2019-05-10
Unable to login when running web app locally


Johannes-Sahlmann:
I want to log into the web app when running it locally to develop code related to MAST authentication. On the current `develop` branch when I click the `Login` button on the top right of the page, I get a 'Server not found' (We can’t connect to the server at oauth.) error.

On the current `master` branch, I get a different error:
```
OperationalError at /login/
no such table: django_session
```

When I do the same on our development server, I get to the page where I enter my AD credentials. After doing that I get a different error:
```
NoReverseMatch at /authorize/
Reverse for '' not found. '' is not a valid view function or pattern name.
```

Do other users have the same problem?

Maybe something is missing in my config file? If so, the application should raise an informative error.

laurenmarietta:
Very strange... I have run into that second error before, but only ever on a feature branch. I cannot reproduce what you are seeing on `develop` or `master`.

I would say the "auth_mast", "client_id", "client_secret", and "mast_token"  keys in your config file are the important ones for this? I assume yours are correct, though, so I doubt those are causing the problem.

I am very curious if anyone else has run into this, and in what contexts.

Johannes-Sahlmann:
I think it must be related to my config file. I had `auth_mast` pointing to an empty string. Once I put in the correct address I am getting one step further. 
now I need to check for `client_id` and `client_secret`

It would be great to catch such problems and return an informative error. I am not sure where these checks should be done, though.

Johannes-Sahlmann:
After setting `client_id` and `client_secret` I can log in when running locally.

Johannes-Sahlmann:
I am however still unable to log in when using the development server.

laurenmarietta:
Hmm... Do you also have your MAST token set as an environment variable (`$MAST_API_TOKEN`)? As described here: https://auth.mast.stsci.edu/info I'm not _sure_ if the dev server web app uses that, but I imagine it might.

I 100% agree about having a more informative error message. The errors that you were experiencing seemed pretty specific; maybe we can add a try/except clause and raise a more informative error if, say, that "no such table: django_session" message comes up. I'll make a new issue for that.

Johannes-Sahlmann:
thinking about it a bit more: Does it make sense that a user is required to have a mast token stored in a local environment variable in order to log into the jwql web app running on the server?
It seems to me that this defeats the purpose of using auth.mast

bourque:
I think the error:

```
NoReverseMatch at /authorize/
[...]

PR #417, opened 2019-05-14
Fix web app redirect issues

bourque:
This PR fixes a bug wherein the web app crashes on the dev server when trying to redirect to the previous page after login; if the user has not yet visited another page within the web app, there exists no view for the previous page.

Addresses concerns in #406


bourque:
@laurenmarietta @Johannes-Sahlmann A super simple edit that seems to fix the error:

```
NoReverseMatch at /authorize/
Reverse for '' not found. '' is not a valid view function or pattern name.
```

I tested on the dev server and it seems to be working fine.

PR #419, opened 2019-05-15
Upload JWQL to PyPI

### LINKED ISSUE #418: Make JWQL pip installable
Currently our `jwql` package is only installable by cloning the repository and running `setup.py`.  It would be easier for users (and perhaps easier for us when distributing our code (#294)) if it were also uploaded to PyPI and `pip` installable.

bourque:
This PR makes some minor changes to the `setup.py` and environment files to get the `jwql` in proper order for uploading to PyPI.

The project has been uploaded and the homepage is here: https://pypi.org/project/jwql/

Closes #418


bourque:
@Johannes-Sahlmann Do you have a working solution/know of a way to automatically upload new package versions to PyPI?  I thought you may have recently figured this out through your work on `jwedb`.

Johannes-Sahlmann:
Yes, I have a working solution for deployment to PyPI (https://github.com/spacetelescope/pysiaf), however that is using TRAVIS and not JENKINS. 
This is from https://docs.travis-ci.com/user/deployment/pypi/ and the principle is to add the code below to the .travis.yaml file:
```
deploy:
  provider: pypi
  skip_existing: true
  user: johannes-sahlmann
  password:
    secure: 
VERYLONGPASSCODESTRING
  on:
    branch: master
    tags: true
```
A quick search did not point me to an obvious equivalent for JENKINS, but I may just have missed it.

bourque:
> Yes, I have a working solution for deployment to PyPI (https://github.com/spacetelescope/pysiaf), however that is using TRAVIS and not JENKINS.
> This is from https://docs.travis-ci.com/user/deployment/pypi/ and the principle is to add the code below to the .travis.yaml file:
> 
> ```
> deploy:
>   provider: pypi
>   skip_existing: true
>   user: johannes-sahlmann
>   password:
>     secure: 
> VERYLONGPASSCODESTRING
>   on:
>     branch: master
>     tags: true
> ```
> 
> A quick search did not point me to an obvious equivalent for JENKINS, but I may just have missed it.

Thanks @Johannes-Sahlmann, I will dig for some Jenkins documentation about this.

bourque:
@obviousrebel @rendinam Do you have any working examples of how to deploy a package to PyPI via the Jenkins build?  I'm thinking it may be as easy as adding appropriate `python setup.py sdist bdist_wheel` and `twine upload` commands to the Jenkins build commands, but we would need to supply the PyPI account password somehow.

obviousrebel:
@bourque I haven't done it but I believe it can be done by adding the credentials to jenkins credentials config and wrapping the commands to upload to PyPI with a `withCredentials(){}` wrapper might be able to do it.

bourque:
@obviousrebel Thanks for the note.  In my latest commit (24bc470) I added some new credential tokens for the `pypi` username and password, as well as some new build and test commands to create a wheel and upload it to `pypi` via `twine`.

I think the new tokens will have to be stored somewhere server side(?).  If so, let me know when is a good time and I can Slack you the username and password.

obviousrebel:
@bourque Apologies I should be ready today to add the tokens to jenkins.

bourque:
I've got this working and the Jenkins build is uploading the package to `pypi`, or skipping that step if the distribution already exists.  

For some reason, one of the `test_api_views` tests are failing (`api/miri/thumbnails/`):

```
self = <http.client.HTTPResponse object at 0x7f41e4e6d198>, amt = 57110
    def _safe_read(self, amt):
        """Read the number of bytes requested, compensating for partial reads.
        Normally, we have a blocking socket, but a read() can be interrupted
        by a signal (resulting in a partial read).
        Note that we cannot distinguish between EOF and an interrupt when zero
        bytes have been read. IncompleteRead() will be raised in this
        situation.

        This function should be used when <amt> bytes "should" be present for
        reading. If the bytes are truly not available (due to EOF), then the
        IncompleteRead exception can be used to detect the problem.
        """
        s = []
        while amt > 0:
            chunk = self.fp.read(min(amt, MAXAMOUNT))
            if not chunk:
>               raise IncompleteRead(b''.join(s), amt)
E               http.client.IncompleteRead: IncompleteRead(81738 bytes read, 57110 more expected)
```

This doesn't appear to have anything to do with the changes in the PR, and is hopefully random?  
I think I will go ahead and merge this and hopes the issue fixes itself.  If it doesn't, we can try solving it in a separate PR.

PR #420, opened 2019-05-17
Leading zeros not required for proposal number

### LINKED ISSUE #260: Smart search for program numbers (without zeros)
As @Johannes-Sahlmann mentioned in #253, it would be preferable if the JWQL search feature didn't enforce trailing zeros. That is, to search for data from program 778, you shouldn't have to type 00778 into the search.

Should be an easy fix, but will require parsing what are currently strings into ints.

laurenmarietta:
Closes #260.

This PR: 
- Updates the homepage proposal search form to search by integer value rather than string value (i.e. accepts "327" not only "00327")
- Updates all URLs that include a proposal number to accept one to five digits (e.g. `jwql.stsci.edu/fgs/archive/327/` is now valid), and update `data_containers.py` accordingly 
- Updates the JavaScript search function on the instrument archive pages to search by integer value rather than string value (i.e. now searching "1" will show not only proposals like "12334" but also like "143")


jwql/website/apps/jwql/data_containers.py
```diff
@@ -378,16 +378,17 @@ def get_filenames_by_proposal(proposal):
...
-        The five-digit proposal number (e.g. ``88600``).
+        The one- to five-digit proposal number (e.g. ``88600``).
 
     Returns
     -------
     filenames : list
         A list of filenames associated with the given ``proposal``.
     """
 
+    proposal_string = '{:05d}'.format(int(proposal))
```
bourque: @laurenmarietta Clever!

jwql/website/apps/jwql/data_containers.py
```diff
@@ -896,14 +901,14 @@ def thumbnails_ajax(inst, proposal=None):
...
             # Temporary workaround for noncompliant files in filesystem
-            filename_dict = {'activity': file_id[17:19],
-                             'detector': file_id[26:],
-                             'exposure_id': file_id[20:25],
-                             'observation': file_id[7:10],
-                             'parallel_seq_id': file_id[16],
-                             'program_id': file_id[2:7],
-                             'visit': file_id[10:13],
-                             'visit_group': file_id[14:16]}
+            filename_dict = {'activity': rootname[17:19],
```
bourque: @laurenmarietta Good call, `rootname` is a more consistent name to use here rather than `file_id`.

jwql/website/apps/jwql/urls.py
```diff
@@ -75,20 +75,20 @@
     re_path(r'^(?P<inst>({}))/unlooked/$'.format(instruments), views.unlooked_images, name='unlooked'),
     re_path(r'^(?P<inst>({}))/(?P<file_root>[\w]+)/$'.format(instruments), views.view_image, name='view_image'),
     re_path(r'^(?P<inst>({}))/(?P<file>.+)/hdr/$'.format(instruments), views.view_header, name='view_header'),
-    re_path(r'^(?P<inst>({}))/archive/(?P<proposal>[\d]{{5}})/$'.format(instruments), views.archive_thumbnails, name='archive_thumb'),
+    re_path(r'^(?P<inst>({}))/archive/(?P<proposal>[\d]{{1,5}})/$'.format(instruments), views.archive_thumbnails, name='archive_thumb'),
```
bourque: @laurenmarietta The power of `regex`!

bourque:
I will let @Johannes-Sahlmann comment if he would like before I merge this

jwql/website/apps/jwql/urls.py
laurenmarietta: 💫

Johannes-Sahlmann:
I can have a look at this early next week. If you want to merge before that, that is fine with me.

bourque:
We've got a bit of a catch-22 situation here.  The Jenkins builds for this PR are failing because the dev web server is not able to handle a test case with no-leading zeros (e.g. `/api/00308/filenames/`) yet, and won't be able to until this PR is merged.  So I think the best course of action here is to just merge this despite the 3 failed API tests.  If for some reason those tests keep failing in future PRs, we can deal with those issues then.

Johannes-Sahlmann:
Are you still expecting a review from me?

bourque:
@Johannes-Sahlmann I wanted to leave it up to you, but it is probably not necessary at this point.  I will go ahead and merge.

PR #422, opened 2019-05-17
Smarter handling of auth errors 

### LINKED ISSUE #408: More informative errors for auth failure
As @Johannes-Sahlmann discovered in #406, the errors that the web app throws if you don't have your `client` or `auth_mast` keywords in the `config.json` file defined properly are not very helpful.

We should add custom error messages there so it is easier for users to tell if something isn't working because of their configuration file.

(Related philosophical question: how do we balance informative error messages for developers with secrecy to maintain security?)

laurenmarietta:
This PR closes #408 and partially addresses #406.

Please test it rigorously - I could not fully recreate the original problems from either of those issues, so I am not sure I have implemented a proper fix.

Changes include: 
- Adding errors if a user tries to run anything while the `auth_mast`, `client_id`, or `client_secret` fields in `config.json` are blank
- Adding a validation error to the EDB forms if the user is not logged into MAST (rather than showing an error page)
- Moving @Johannes-Sahlmann's code to get the MAST token from the top of `engineering_database.py` into `utils.get_mast_token()`
- Updating `data_containers.log_into_mast()` to use the new `utils.get_mast_token()` function

Also, I am specifically concerned about how I re-wrote `data_containers.log_into_mast()`. This function used to just try to find the MAST token with `request.POST.get('access_token')`. I replaced that with a new `utils` function taken from `engineering_database.py` that goes through not just the POST request but also `config.json` and the user's environment variables to get the MAST token. Is there a reason that token would _need_ to come from the POST request?


Johannes-Sahlmann:
I am testing this out on a local branch and I think a problem is the `from astroquery.mast import Mast` import in `utils.py`. Because this is called before the customized import in `data_containers.py` it leads to problems with setting the MAST server that astroquery uses. The customization has to take place before the first import. I am working on it and will open a PR against @laurenmarietta 's fork when ready.

Johannes-Sahlmann:
An independent complication is that the tokens have to match the server that is being used. While testing this I found myself in a situation where I was providing a MAST token that I had created on the operational server, which failed because the `astroquery.mast` module was configured to access the dev server. So until we switch to fully use the MAST ops server, one has to ensure to use a correctly generated token (possibly this should be documented temporarily). I opened a PR against the fork with some tweaks I saw were necessary. 

Something weird also happened: When testing this branch and performing a successful mnemonic query on the web app, the returned bokeh plot was empty (blank panel, no axes or buttons). Not sure what is causing this.

laurenmarietta:
It's definitely tricky to keep track of where the tokens are being generated and under which version of MAST, I agree... Maybe we can try to address that with better docs in a future issue?

I take it that laurenmarietta/jwql#1 fixed the local error you were running into with the servers?

I have been seeing that the mnemonic query Bokeh plot has been empty for a while now, but I had no idea why so I thought I would ignore it and hope it goes away... clearly that is not working. Are you only seeing that behavior on this branch?

Also, I updated this PR to include and implement a new function `utils.utils.check_config()` to reduce duplication of the try/except clauses.

jwql/website/apps/jwql/oauth.py
```diff
@@ -162,8 +164,11 @@ def user_info(request, **kwargs):
 
         # If user is authenticated, return user credentials
         if cookie is not None:
+            check_config('auth_mast')
+            auth_mast = get_config()['auth_mast']
+
             response = requests.get(
```
Johannes-Sahlmann: Just as a sidenote: I tried this with the MAST ops server (specified via `auth_mast`) and it failed, probably because JWQL is not registered there yet.

jwql/utils/credentials.py
```diff
@@ -0,0 +1,49 @@
+"""Utility functions related to accessing remote services and databases.
```
bourque: Good idea creating a separate module for this!

jwql/utils/credentials.py
```diff
@@ -0,0 +1,49 @@
...
+
+Authors
+-------
+
+    - Johannes Sahlmann
+
+Use
+---
+
+    This module can be imported as such:
```
bourque: Small issue but `sphinx` docs will render better if this is written like this:

```python
This module can be imported as such:
::

    import credentials
    token = get_mast_token()
```

jwql/utils/credentials.py
```diff
@@ -0,0 +1,49 @@
...
+ """
+import os
+
+from astroquery.mast import Mast
+
+from jwql.utils.utils import get_config, check_config
+
+
+def get_mast_token(request=None):
+    """Return MAST token."""
```
bourque: Please provide a `Parameters` section to the docstring that describes `request`

jwql/utils/utils.py
```diff
@@ -395,6 +395,32 @@ def get_config():
     return settings
 
 
+def check_config(key):
```
bourque: This is really useful!

jwql/utils/utils.py
```diff
@@ -395,6 +395,32 @@ def get_config():
...
+    """Check that the config.json file contains the specified key
+    and that the entry is not empty
+
+    Returns
```
bourque: I think you mean `Parameters` here, not `Returns`

jwql/website/apps/jwql/oauth.py
```diff
@@ -63,6 +63,8 @@ def register_oauth():
     """
 
     # Get configuration parameters
+    for key in 'client_id client_secret auth_mast'.split():
```
bourque: That is an interesting way to iterate through a list! But I wonder if it would read easier if you just did:

```python
for key in ['client_id', 'client_secret', 'auth_mast']:
```

jwql/utils/credentials.py
laurenmarietta: Whoops, I completely forgot to add this docstring. Thanks for the reminder!

bourque:
I think the Jenkins build failed for this because we didn't restart the dev web server after #420 was merged.  I restarted it just now so hopefully it will pass now.

bourque:
This is working well when testing on my end.  Nice work!

PR #429, opened 2019-06-03
Update web app bokeh version

### LINKED ISSUE #425: EDB mnemonic query plot blank on local server
@Johannes-Sahlmann and I have both noticed that, when the web app is being run on a local server, the EDB mnemonic query `bokeh` plot does not appear:
Screen Shot 2019-05-22 at 9 25 05 AM
I personally see this behavior on every branch - feature branches, `develop`, and `master`.

Good news is, I don't see this bug on the development server (dljwql.stsci.edu). This suggests to me that it has something to do with running the local Django server? 

If anyone else has seen this, please share.

bourque:
This PR updates the web app to use bokeh version `1.2.0`, which is the current supported version in the `jwql` environment.  This also fixes the issue described in #425 

Closes #425


bourque:
@Johannes-Sahlmann Could you review this?  Everything seems to be working just fine for me locally.

I think it is rather annoying that we will have to keep these lines in `base.html` updated with every update to the `bokeh` version, but currently I can't think of a better solution.

bourque:
Since we are waiting on this for merging #432 and its a simple one-line change, I am going to make an executive merge here

PR #432, opened 2019-06-04
Add bokeh version test

laurenmarietta:
I was inspired (frustrated?) by #425 to write up a quick test that will fail if the Bokeh versions in our HTML templates don't match our environment.

(This will fail Jenkins until #429 is merged.)

### DIFF
--- a/jwql/tests/test_plotting.py
+++ b/jwql/tests/test_plotting.py
@@ -21,9 +21,11 @@ Use
 import glob
 import os
 import re
+import sys
 
 import bokeh
 from pandas import DataFrame
+import pytest
 
 from jwql.utils.plotting import bar_chart
 
@@ -46,6 +48,8 @@ def test_bar_chart():
     assert str(type(plt)) == "<class 'bokeh.plotting.figure.Figure'>"
 
 
+@pytest.mark.skipif(sys.version_info[:2] != (3, 6),
+                    reason="Web server run on Python 3.6")
 def test_bokeh_version():
     """Make sure that the current version of Bokeh matches the version being
     used in all the web app HTML templates.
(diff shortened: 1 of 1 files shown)


jwql/tests/test_plotting.py
```diff
@@ -36,3 +46,28 @@ def test_bar_chart():
...
+
+    template_paths = os.path.join(JWQL_DIR, 'website/apps/jwql/templates', '*.html')
+    all_web_html_files = glob.glob(template_paths)
+
+    for file in all_web_html_files:
+        with open(file) as f:
+            content = f.read()
+
+        # Find all of the times "bokeh-#.#.#' appears in a template
+        html_versions = re.findall(r'(?<=bokeh-)\d+\.\d+\.\d+', content)
```
bourque: @laurenmarietta Would this also include instances of `bokeh-widgets-#.#.#` as well?  I think we have a few of those scattered around our templates.

bourque:
When testing this, I noticed that we have a fair amount of `href`s pointing to `bokeh` amongst our various templates.  I wonder if those are actually necessary, since `bokeh` is imported in `base`, and each template extends `base`.  It might be worth seeing if removing the non-`base` `bokeh` imports actually affects anything.  If not, better to remove them I think.

jwql/tests/test_plotting.py
laurenmarietta: Good idea!

laurenmarietta:
I think I will try to address the issue of where `bokeh` is getting imported as part of #438, rather than here.

bourque:
The Jenkins builds for this are failing for this (see #443), but since it appears to be an issue unrelated to these code changes, I will go ahead and merge this.

PR #439, opened 2019-06-05
[WIP] Web app log pages

catherine-martlin:
New PR for logging monitor web app updates. 

- [x] Make landing page for links to logs
- [x] Make pull down for logs on that page for all past logs
- [ ] Update the url to allow for viewing those past logs when clicked on
- [ ] Update pages for monitors with link to log pages


bourque:
@catherine-martlin I pulled this branch to have a look.  This is a good first implementation; feel free to proceed with the remaining changes to get things rendered in a readable way.  

I think in a future PR we could consider moving towards drop down menus instead of separate buttons for all of the different monitors, as buttons might become overwhelming when we have many monitors implemented.

catherine-martlin:
@bourque - maybe we can chat this week? This is very close - I'm just struggling with activating the buttons in the dropdown menus at the top of each type/monitor page to go to individual logs instead of the default last log available. I need some help with the urls.py update needed to create the proper url paths and with formatting the dropdown button in view_log.html to link to those urls properly.

Issue #400, opened 2019-05-06
Update Django to 2.2


laurenmarietta:
There are aspects of the web app that prevent us from upgrading `django` to version 2.2. We should investigate what those are, fix them, and update our environment files.

The change log from the 2.2 release might prove helpful for troubleshooting there:
https://docs.djangoproject.com/en/2.2/releases/2.2/#backwards-incompatible-2-2

PR #448, opened 2019-06-21
Update web app to use django 2.2

bourque:
This PR provides various changes to get the web app working with `django` version 2.2. 

The new version requires the existence of a define `django.template.backends.django.DjangoTemplates` `TEMPLATE`.  Also, some `not_found` requests now also require additional parameters such as `exception`, which have been wrapped up in a `*args` parameter.


bourque:
Closes #400

PR #450, opened 2019-06-24
Updated README to include pip install instructions

### LINKED ISSUE #441: Update docs for pip installation
Now that JWQL is on PyPI (#419), we should update our docs to explain how to `pip install jwql`!

bourque:
This PR updates the `README` to include a section on installing the `jwql` package for users (which is simply a `pip install jwql`).

Even with the capability to `pip install jwql`, the install instructions for contributors remains the same, as to encourage a `python setup.py develop` so that `jwql` is installed to a `$PATH` that makes it easy to test code changes.

Also included is a small update to the list of current dev members and acknowledgements. 

Closes #441

### DIFF
--- a/README.md
+++ b/README.md
@@ -23,13 +23,23 @@ Official API documentation can be found on [ReadTheDocs](https://jwql.readthedoc
 
 The `jwql` application is currently under heavy development.  The `1.0` release is expected in 2019.  Currently, a development version of the web application can be found at [https://dljwql.stsci.edu](https://dljwql.stsci.edu).
 
-## Installation
+## Installation for Users
+
+To install `jwql`, simply use `pip`:
+
+```
+pip install jwql
+```
+
+The section below describes a more detailed installation for users that wish to contribute to the `jwql` repository.
+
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -64,25 +74,31 @@ Following the download of the `jwql` repository, contributors can then install t
 conda update conda
 ```
 
-Next, activate the `base` environment:
+Next, activate the `base` or `root` environment (depending on your version of `conda`):
 
 ```
-source activate base
+source activate base/root
```
laurenmarietta: Is this just here to make sure that users are starting in a clean environment? Should the command be `source deactivate` instead?

bourque: @laurenmarietta If I recall correctly when I first made these instructions, this command doesn't necessarily have to be executed if the user already had `conda` installed, but if a user has just installed `conda` for the first time, a new environment can't be made until the `base`/`root` is activated(?)

laurenmarietta: Oh, that would make sense then. Thanks for explaining!

Issue #454, opened 2019-06-25
Add observation time to dark monitor database tables


bhilbert4:
Currently the database tables used by the dark current monitor keep track of the time/date on which the monitor is run and each entry is added, but they do not keep track of the time/date associated with the observations being examined. We should add this to the database in order to facilitate the creation of plots.

What's the best way/place to add this information? We need it in the dark_dark tables. It may also be useful in the dark_pixel_stats tables?

PR #456, opened 2019-06-27
Add observation times to the database tables

bhilbert4:
This PR adds code to the `dark monitor` such that it will collect the observation time from all input files. It will then save the earliest, latest, and (calculated) mean observation time for the dataset into the appropriate database tables.

This will help with plotting the results of the dark monitor.


bhilbert4:
Closes #454

bhilbert4:
@bourque I think this is ready to review. It ran successfully in my local tests.

jwql/utils/instrument_properties.py
```diff
@@ -234,3 +235,49 @@ def calc_frame_time(instrument, aperture, xdim, ydim, amps, sample_time=1.e-5):
...
+    ----------
+    filename : str
+        Name of fits file
+
+    Returns
+    -------
+    obs_time : datetime.datetime
+        Observation date and time
+    """
+    with fits.open(filename) as h:
```
bourque: @bhilbert4 I don't have much of an opinion on this one way or the other, but I wonder if it would be better to use `jwst.datamodels` here to extract this information?  Especially if it is faster than FITS IO.  What do you think?

bhilbert4: In general `jwst.datamodels` are much slower than `fits.io` because the datamodels do a bunch of metadata validation when the file is opened.

Issue #222, opened 2018-12-14
Integrate Bokeh into the Django backend


gkanarek:
The current monitor scripts serve a dual purpose: both that of actually performing the relevant monitor task, and also of creating the html and Javascript output to be embedded in the relevant webpages, as required. I propose a new architecture, where the Bokeh code is integrated with the Django backend, and which reduces the monitor scripts to standalone monitoring tasks. This would improve on the existing setup in a few ways:

1. It will be easier for users to make use of the monitor scripts on their own for various analysis tasks, rather than having to deal with Bokeh output.
2. Making the Bokeh code part of the Django backend will allow us to leverage the power of Bokeh more effectively (e.g., streaming data to plots, not having to re-create html code every time a monitor script is run, etc).
3. People who develop and maintain the monitor scripts don't have to be versed in Bokeh or Django.

Before we can make this change, we need to decide on a standardized methodology for storing the results of the various monitors which can be read by the Bokeh code. Then, each script will need to be split in two: the actual monitoring task (which will be run via a cron job and store its output in the standardized format), and the web interface, which will be incorporated into the Django backend.

bourque:
FYI, We have come up with a generic way to store instrument monitor data (#304) in the form of database tables.

bourque:
@gkanarek Could you open a WIP pull request for your work on this?  It will help us reference and coordinate the work in other issues and pull requests.

bhilbert4:
We have moved away from Bokeh templating.

Issue #114, opened 2018-07-31
Integrating FITSViewer with JWQL web app


gkanarek:
Continuing discussion with @laurenmarietta from #45.

So as it happens, the static files should be unnecessary - I just included them so I could do things like center the viewer on the page and control some other CSS stuff while using Bokeh server. We can hopefully use the Bokeh `embed` module instead - there's a user guide [here](https://bokeh.pydata.org/en/latest/docs/user_guide/embed.html) on how to do it. Other resources I've found are [here](https://www.hackerearth.com/practice/notes/bokeh-interactive-visualization-library-use-graph-with-django-template/) and [here](https://bokeh.github.io/blog/2018/6/13/release-0-13-0/) (specifically the part on custom templates). Are we using Jinja templating? If so, that might make things a little easier to embed.

An alternative is to run the FITSViewer off of Bokeh server, using the latter as a library, embedded in the larger framework of the Django app. There are some examples of that [here](http://bokeh.pydata.org/en/latest/docs/user_guide/server.html#embedding-bokeh-server-as-a-library), though not for Django.

I don't know enough about how Django works to know the best way to proceed here. Thoughts?

bourque:
@bourque Reach out to individuals regarding support for Ginga

bourque:
@gkanarek Have you heard of/used LSST's `firefly` tool?  Perhaps something we could possibly use?

https://firefly.lsst.io/index.html

bhilbert4:
JWQL currently uses bokeh to interactively view fits images. While we could certainly do with some improved performance on this tool, that can be captured in a new issue. Closing this issue for now.

PR #459, opened 2019-06-28
Bokeh templating

gkanarek:
This branch is WIP, but we're eventually going to want to merge it into develop.


laurenmarietta:
This PR addresses, but I don't think completely closes, #114 and #222.

PR #460, opened 2019-06-28
Add dark monitor Bokeh templating files

### LINKED ISSUE #369: Specify Dark Monitor Plots
I thought there was already an issue open for this, but I couldn't find it. I just wanted to write down a list of dark monitor-related plots to be displayed in the web app.

1. Histograms of dark current rates. Currently there are separate histograms generated for each amplifier, as well as for the full detector, for all instruments. This should change a bit, as MIRI's readout pattern interleaves all the amplifiers. The code as-is won't be able to handle that. 
2. Mean dark current versus time (across multiple dark monitor runs)
3. Hot/dead/noisy pixels: Maybe display the mean dark rate image and mark the locations of any new hot/dead/noisy pixels with different colored dots? If nothing new, still display the mean dark rate image as that can still be useful.
4. Display tables of new hot/dead/noisy pixels? How will potential users get this information, assuming they won't have access to the JWQL database? Add to API?

Bonus: 
I think it would be interesting and useful to be able to plot the mean dark rate versus certain EDB mnemonics. For NIRCam we would want to watch dark rate versus the ASIC voltage and the FPA temperature.

In the case where the mean dark rate differs from the historical mean by some amount, overplot current and historical histograms?

bhilbert4:
This PR adds Bokeh templating files for creating dark monitor plots.

At the time of this initial commit, this is not yet functional.

Closes #369.


laurenmarietta:
I'm going to play around with this a bit to see if I can get a small demo working for our TIPS talk.

laurenmarietta:
(Just documenting that the current state of this branch is what we used for the TIPS web app demo in July.)

bourque:
@bhilbert4 #459 is merged, and I've updated this branch to include those changes, so now this can be finished up/tested with dark monitor plots.  Could you take a look?

bhilbert4:
Will do.

bhilbert4:
I think I have the issue related to the dark_monitor pre_init and BokehTemplate pre_init resolved after some conversation with @gkanarek. There are still issues to work on though. Currently when I try to create the dark_monitor plot page, the app thinks for a couple minutes, and then crashes with the statement that nodejs is required for rendering custom models....which is odd because I installed nodejs.

bhilbert4:
Re-tried everything this morning. `conda install nodejs` was providing a very old version of nodejs. I had to do `conda install nodejs==10.15.3`. With that in place, the dark monitor webpage is rendered, but none of the figures are there. I'm wondering if this is something similar to #467 ? I'm not seeing quite the same error messages in the Firefox developer console tab, but similar-ish? I'm testing with Bokeh 1.3.4, which is what we have in the requirements file as well as the python 3.6 env file. I tried downgrading to Bokeh 1.3.2, but got the same errors. 
jwql_error_firefox
jwql_error_terminal

bhilbert4:
Current status: making slow progress. 

Histogram plots: 10 histograms now appear on the page. They are correctly labeled, but looking at the histograms themselves, I'm worried that the same data is being plotted for all detectors.

Images: Can get a single image to be shown on the page. Code for showing images for all detectors is not working.

Dark rate vs time: This plot is the one that was the source of the bad array length errors that I was seeing. It turns out that the cause of this is that all entries currently in the database have the same time stamp. This was throwing off the code as it tried to set x-axis bounds. I've added a buffer to the x and y ranges so that this error won't happen again. As with the histograms, it appears that the same dark data is being plotted for all apertures. Somehow the new data is not being loaded from the database after the instrument and aperture are set using the setter.

bourque:
@bhilbert4 I was able to get things working on my end!  However, I could only do so by removing `NRCA5` and `NRCB5` from the list of apertures to plot.  I think it may have something to do with a lack of data available in the database for those apertures (I recently recent the database to test the bias monitor).  I will investigate that further, but it seems promising.

bhilbert4:
That's good (if a little strange)! How would a lack of data for 2 detectors cause all the rest to be plotted with the data from one of the other detectors?

bourque:
Finally getting back to this:  I made a few commits to clean things up just a bit to appease `flake8` and our style guide.  Again, I can only get things working by removing `NCRA5_FULL` and `NCRB5_FULL` from the list of apertures to plot.

bourque:
@bhilbert4 After some digging, I think I found the root cause of the issue with `NRC[A|B]5_FULL`.  Looking at this bit of code from `jwql.website.apps.jwql.monitor_dark_bokeh.load_data()`:

```python
        self.pixel_table = session.query(self.pixel_table) \
            .filter(self.pixel_table.detector == self.detector) \
            .all()
```

This code expect `self.detector` to be e.g. `NRCA5_FULL`, however the `dark_pixel_stats` table in the database stores the `detector` as `NRCALONG`, which means that no results are returned from this query, which then breaks the rest of the plotting code.

Is there a reason for storing things as `NRCALONG` as opposed to `NRCA5_FULL`?  Do they mean different things?  Is there a way we could avoid an inconsistency in naming conventions across code and database tables?

bhilbert4:
Ugh, what a pain. Good catch. NRCA5_FULL is an aperture name. (You can also have e.g. NRCA5_SUB160). In pipeline outputs the detector is stored as NRCALONG. Most people change this to NRCA5 in order to be easier to work with, and to be consistent with aperture names.

PR #463, opened 2019-07-03
Fix pytest-cov dependency in requirements.txt

catherine-martlin:
Updating so we can get RtD running again

Edit (Bourque): We will address RTD concerns in the `update-docs` branch and use this PR to fix the `pytest-cov` dependency.


catherine-martlin:
I forgot how picky RtD can be - it won't let me test on a forked version of the project so I will likely have to merge this with the one update. If it doesn't solve all the issues I'll need to make a branch off the master instead of off my fork so we can test it on RtD while working on it.

bourque:
@catherine-martlin Since this small fix is one we should make regardless of RTD, I'm happy to merge this in and we can go from there.

The `upstream` of this repo has a branch called `update-docs` -- Feel free to pull that branch, push commits to it, and use that branch to test out on RTD -- you just have to go to the `Admin` > `Advanced Settings` button and switch the `Default Branch` from `master` to `update-docs` and the RTD builds with use that branch.  When everything looks to be operational, we can switch that setting back to `master`.  In this way, you can use `update-docs` as a sandbox and be able to test things more easily.

bourque:
@catherine-martlin I'm gonna go ahead and merge this.  We can address why RTD is failing using the `update-docs` branch in `upstream` so that we can easily test changes on RTD.

PR #466, opened 2019-07-11
Validate config.json while opening

### LINKED ISSUE #423: add config.json format verification?
Wrong formatting of `config.json` raises misleading error messages.

For instances, having a trailing comma in the last entry of the dictionary raises 
`json.decoder.JSONDecodeError: Expecting property name enclosed in double quotes: line 14 column 1 (char 594)`

Maybe we should add a verification function that checks the formatting of `config.json`?

laurenmarietta:
This PR:
- Adds a new function, `utils.utils._validate_config()`, that uses the `jsonschema` package to make sure the `settings` dictionary created by `get_config()` includes all of the needed keys, and that the entries are of the correct data type.
- Puts wrappers around the JSON errors in `get_config()` and `_validate_config()` to make them slightly more user-friendly.
- Adds `jsonschema` to the requirements and environment files.
- Adds a test for `_validate_config()`

I know we try to be careful about what we commit to the repo when it comes to the config file, but I took care here not to commit anything that we don't have posted on our wiki page. If I did miss something, of course, please tell me!!

Closes #423.


jwql/utils/utils.py
```diff
@@ -387,20 +389,31 @@ def get_config():
...
-        settings = json.load(config_file)
+    with open(config_file_location, 'r') as config_file_object:
+        try:
+            # Load it with JSON
+            settings = json.load(config_file_object)
+        except json.JSONDecodeError as e:
+            # Raise a more helpful error if there is a formatting problem
+            raise ValueError('Incorrectly formatted config.json file. '
+                             'Please fix JSON formatting: {}'
+                                       .format(e))
```
bourque: It looks like this line is indented too much?  Perhaps this can just be put back into line 407.

jwql/utils/utils.py
```diff
@@ -426,6 +439,69 @@ def check_config(key):
         )
 
 
+def _validate_config(config_file_dict):
+    """Check that the config.json file contains all the needed entries with
+    expected data types
+
+    Parameters
+    -------
```
bourque: I think `sphinx` will freak out if there are not enough dashes to cover all of the `Parameters` heading

jwql/tests/test_utils.py
```diff
@@ -344,3 +345,46 @@ def test_filesystem_path():
     location = os.path.join(get_config()['filesystem'], 'jw96003', filename)
 
     assert check == location
+
+
+def test_validate_config():
+    """Test that the config validator works.
+    """
```
bourque: If the docstring only spans one line, I think we usually just put the description and both`"""` all on one line

jwql/tests/test_utils.py
```diff
@@ -344,3 +345,46 @@ def test_filesystem_path():
...
+    with pytest.raises(Exception) as excinfo:
+        _validate_config(bad_config_dict)
+    assert 'Provided config.json does not match the required JSON schema' \
+           in str(excinfo.value), \
+        'Failed to reject incorrect JSON dict.'
+
+    # Make sure a good config does not!
+    good_config_dict = {
+        "connection_string": "",
+        "database": {
```
bourque: Good call on splitting this all into separate keys; it will be much easier to refer to individual components this way

jwql/utils/utils.py
laurenmarietta: Good catch, thank you!

laurenmarietta:
Thanks for the edits - easy changes!

PR #471, opened 2019-07-23
Release 0.21.0 [develop]

bourque:
This PR updates `setup.py` and `CHANGES.rst` for the `0.21.0` release.

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -1,3 +1,45 @@
+0.21.0 (2019-07-23)
+===================
+
+New Features
+------------
+
+Project & API Documentation
+~~~~~~~~~~~~~~~~~~~~~~~~~~~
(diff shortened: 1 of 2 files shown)


laurenmarietta:
Hmm. When I run the test suite on this branch locally, I am getting some unexpected failures in `test_api_views` and `test_monitor_mast` (even though they don't seem to have failed on Jenkins). @bourque are you seeing that?

There is truly no reason why those failures would be caused by this PR, but I wanted to mention them before I approve it.

bourque:
@laurenmarietta Good catch, I am also seeing that locally.  I can open up issues for those!   I suggest that we carry on with the release despite these, especially since things seem to be passing on Jenkins 🤷‍♂

PR #499, opened 2019-08-28
Update header display

catherine-martlin:
Updates to allow viewing more than the primary header.


Johannes-Sahlmann:
Hi @catherine-martlin . I am running your branch and see that all extension headers are displayed, but not in a table format. Would you agree if I branch off this and try to implement the header view using a table format like for the EDB?

catherine-martlin:
@Johannes-Sahlmann - Yep! Since you have experience with creating the tables that are searchable feel free to start working on this branch and I'll take a break while you do. You should be able to access a couple dictionaries through the work I already did that contain each extension, its keywords, and their values. Let me know if you have questions and please feel free to change the formatting I've already done if there is another way that would work better with your plans to make it searchable! (My main goal was to have and show the information so I may have come at it a different way.)

Johannes-Sahlmann:
@catherine-martlin  The warnings I got looked like popups in the browser.

bourque:
@catherine-martlin Is this ready for review, or is there still work to be done on this?

catherine-martlin:
@bourque  - when I checked, I didn't get the pop-up errors that Johannas did so I think that it's ready for review. Perhaps part of the review would be to check if your system gets that pop-up as well?

jwql/website/apps/jwql/templates/view_header.html
```diff
@@ -9,17 +9,22 @@
 {% block content %}
 
     <main role="main" class="container">
-    	<h4>{{ file }}</h5>
+    	<h4>{{ file }}</h4><h>
```
bourque: Nice catch

PR #506, opened 2019-09-12
Switch to using astroconda instead of astroconda-dev

### LINKED ISSUE #505: Avoid using astroconda-dev
As we approach our 1.0 release, we should not be relying on `astroconda-dev` for some of our dependencies, since that channel is for development purposes and may break from time to time.  We should go through our list of dependencies and make sure we can install them from default `conda` channels, `astroconda`, or `pip`.

bourque:
This PR modifies the `jwql` `conda` environments to use the `astroconda` channel instead of `astroconda-dev`.

Closes #505


environment_python_3_5.yml
```diff
@@ -25,7 +24,6 @@ dependencies:
 - setuptools=41.0.1
```
jbcurtin: looks like this is using python 3.6.4 instead of python 3.5.6?

bourque: @jbcurtin How can you tell this?

jbcurtin: Expanding the code-snippet doesn't allow me to comment on the exact line, here is the line number from the whole file
https://github.com/spacetelescope/jwql/blob/0a987316f7f94837ed47c1e33dde8e46af3875e3/environment_python_3_5.yml#L20

bourque: Ah yes, good catch!

bourque:
It turns out that the `jwst` pipeline does not support Python 3.5, which we didn't notice until now because we were accidentally using Python 3.6 in our `environment_python_3_5.yml` file (thanks @jbcurtin for finding that mistake!)

Since the instrument monitors of `jwql` depend on the JWST pipeline, I think the best course of action is to remove support for Python 3.5 altogether.  I don't believe anyone has been developing in Python 3.5 anyways.  As such, I will remove `environment_python_3_5.yml` and the supporting Jenkins build in this PR.  If there is a desire for Python 3.5 support moving forward, it will be easy enough to add these things back, but it will require us to fix a few of our tests and corresponding modules.

PR #517, opened 2019-10-21
Moved DEPENDENCY_LINKS to before setup() in setup.py to fix logging

### LINKED ISSUE #516: DEPENDENCY_LINKS in setup.py causing bug in logging_functions
With the introduction of the `DEPENDENCY_LINKS` variable in `setup.py`, the logging of monitors is now failing to log the versions of depenencies listed, since the `REQUIRES` is not immediately followed by `setup()`:

```python
for i, line in enumerate(data):
            if 'REQUIRES = [' in line:
                begin = i + 1
            elif 'setup(' in line:
                end = i - 2
```

The solution is so simple move `DEPENDENCY _LINKS` to be defined before `REQUIRES`.

bourque:
Closes #516

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -9,6 +9,7 @@ AUTHORS += 'Graham Kanarek, Catherine Martlin, Johannes Sahlmann, Ben Sunnquist'
 
 DESCRIPTION = 'The James Webb Space Telescope Quicklook Project'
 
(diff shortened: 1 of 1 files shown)


bourque:
@catherine-martlin Could you review this?  It's just a small fix.

PR #518, opened 2019-10-22
Fix typo in monitor table

bourque:
This PR fixes a small typo in the `status` field of the `monitor` database table.  This typo has been causing errors in the `dark_monitor` when it completes and tries to update this table.


bourque:
@bhilbert4 Kind of a silly review here, but could you hit the button?

Issue #143, opened 2018-09-11
Enable flipping through preview images with keyboard arrows


bourque:
Users are going to be using the web application to visually inspect many images.  It would be convenient if they could easily navigate to next and previous images with the keyboard arrow keys intend of clicking a button with a mouse.

bourque:
I believe @laurenmarietta has poked around at implementing this but doesn't quite have it working(?)

laurenmarietta:
I tried this before I had any understanding of JavaScript, and then haven't gotten back to it sense.

I have a bit more confidence that I could figure it out now - or I would be happy to delegate out to someone else who wants to learn a little JavaScript!

Issue #149, opened 2018-09-17
Scroll bar for navigating through preview images with many integrations


bourque:
Currently our webpage for displaying a preview image displays one integration at a time with the option to click a right or left arrow button to display the next or previous integration.  It would perhaps be more useful to turn this feature into some sort of scroll bar instead of having to click a button many times.

bhilbert4:
Also keep in mind that in some cases (time-series observations) there will be cases with hundreds or thousands of integrations per exposure. We may want to come up with a smart way to deal with this. The majority of observations will most likely have only a couple integrations per exposure.

PR #522, opened 2019-11-04
Allow arrow keys to change which integration image is showing

bsunnquist:
This allows the right/left arrow keys to increase/decrease the integration image displaying.

I believe this is what we had in mind for https://github.com/spacetelescope/jwql/issues/143, basically just letting the arrow keys take the place of clicking the buttons.

### DIFF
--- a/jwql/website/apps/jwql/templates/view_image.html
+++ b/jwql/website/apps/jwql/templates/view_image.html
@@ -109,6 +109,22 @@
 	    	<a>Lauren needs to figure out what to do with these: {{suffixes}}</a>
 	    {% endif %}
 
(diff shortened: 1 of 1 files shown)


bsunnquist:
Although, another option is to save the arrow keys for flipping between different groups of images (e.g. flipping through all 'unlooked' images), rather than using it for flipping through the integrations within a single image

bsunnquist:
Ok, I believe the intent was for the arrow keys to be used for flipping through images in a set (not the integrations within a single image), and the scrollbar/slider would be used to change integrations (based on https://github.com/spacetelescope/jwql/issues/149).

If this is true, I can go ahead and close this pull request, since it's using the arrow keys to change integrations.

PR #523, opened 2019-11-06
Add image slider to change integration

bsunnquist:
This adds a slider below the image that lets you change which integration is showing - I think this is what we had in mind for https://github.com/spacetelescope/jwql/issues/149. It's not too useful for images with a low number of integrations, but it will be helpful for those with many (e.g. if there are 100 integrations you won't have to click 90 times to see the 90th integration now).

I kept the button functionality as is; I just moved them next to the slider because I noticed that they were collapsing strangely when you shrunk the window size down.

Let me know if this is what you had in mind, or if you think there's a better way.


bsunnquist:
image

PR #545, opened 2020-03-06
Fix typo in jwst installation version tag

bhilbert4:
This PR fixes a small typo that was causing the wrong version of the ``jwst`` package to be installed when using the environment file. Replacing the # symbol with the @ symbol allows the specified version to be installed. I updated the version to the latest released version. I also updated the version of ``asdf`` to match that required by 0.15.0 of the pipeline.


bourque:
Thanks @bhilbert4, looks good

PR #557, opened 2020-03-31
Make Readnoise monitor

bsunnquist:
Make the JWQL Readnoise Monitor


bsunnquist:
@bourque This is all ready for testing. It works for both NIRCam and NIRISS data.

jwql/instrument_monitors/common_monitors/readnoise_monitor.py
```diff
@@ -0,0 +1,625 @@
...
+            logging.info('\tRunning pipeline on {}'.format(filename))
+            try:
+                processed_file = self.run_early_pipeline(filename, group_scale=group_scale)
+                logging.info('\tPipeline complete. Output: {}'.format(processed_file))
+            except:
+                logging.info('\tPipeline processing failed for {}'.format(filename))
+                continue
+
+            # Find amplifier boundaries so per-amp statistics can be calculated
+            _, amp_bounds = instrument_properties.amplifier_info(processed_file, omit_reference_pixels=True)
```
bourque, line 388: Nice use of `_`!

jwql/instrument_monitors/common_monitors/readnoise_monitor.py
```diff
@@ -0,0 +1,625 @@
...
+            The input ramp data. The data shape is assumed to be a 4D array in
+            DMS format (integration, group, y, x).
+
+        Returns
+        -------
+        readnoise : numpy.ndarray
+            The 2D readnoise image.
+        """
+
+        # Create a stack of CDS images (group difference images) using the input ramp data,
```
bourque, line 309: It is not clear to me what "CDS" means here, could you write out what the acronym means at least once somewhere in this function?

jwql/instrument_monitors/common_monitors/readnoise_monitor.py
```diff
@@ -0,0 +1,625 @@
...
+        # combining multiple integrations if necessary.
+        logging.info('\tCreating stack of CDS difference frames')
+        n_ints, n_groups, n_y, n_x = data.shape
```
bourque, line 312: I would suggest to use slightly more explicit variable names here, maybe something like `num_ints`, `num_groups`.  It is small, but could help with the readability.

jwql/instrument_monitors/common_monitors/readnoise_monitor.py
```diff
@@ -0,0 +1,625 @@
...
+                if isinstance(amp_stats[key], (int, float)):
+                    readnoise_db_entry[key] = float(amp_stats[key])
+                else:
+                    readnoise_db_entry[key] = amp_stats[key].astype(float)
+
+            # Add this new entry to the readnoise database table
+            self.stats_table.__table__.insert().execute(readnoise_db_entry)
+            logging.info('\tNew entry added to readnoise database table: {}'.format(readnoise_db_entry))
+
+            # Remove the raw and calibrated files to save memory space
```
bourque, line 469: Good idea!

jwql/instrument_monitors/common_monitors/readnoise_monitor.py
```diff
@@ -0,0 +1,625 @@
...
+                             'entries_found': len(new_entries),
+                             'files_found': len(new_files),
+                             'run_monitor': monitor_run,
+                             'entry_date': datetime.datetime.now()}
+                self.query_table.__table__.insert().execute(new_entry)
+                logging.info('\tUpdated the query history table')
+
+        logging.info('Readnoise Monitor completed successfully.')
+
+    def run_early_pipeline(self, filename, group_scale=False):
```
bourque, line 577: I'm thinking it might be better to move this to the `pipeline_tools` module? In case it can eventually be used by other monitors.

jwql/instrument_monitors/common_monitors/readnoise_monitor.py
```diff
@@ -0,0 +1,625 @@
...
+        and the counts in each bin.
+
+        Parameters
+        ----------
+        data : numpy.ndarray
+            The input data.
+
+        Returns
+        -------
+        n : numpy.ndarray
```
bourque, line 271: I would suggest to rename this variable to something like `counts`, which is much more explicit than `n`.

jwql/instrument_monitors/common_monitors/readnoise_monitor.py
```diff
@@ -0,0 +1,625 @@
...
+        Returns
+        -------
+        output_filename : str
+            The full path to the output png file.
+        """
+
+        output_filename = os.path.join(self.data_dir, '{}.png'.format(outname))
+
+        # Get image scale limits
+        z = ZScaleInterval()
```
bourque, line 223: I would suggest to rename this variable something like `zscale`.  It is good to avoid one-letter variables when you can.

jwql/instrument_monitors/common_monitors/readnoise_monitor.py
```diff
@@ -0,0 +1,625 @@
...
+        # Loop over all instruments
+        for instrument in ['nircam', 'niriss']:
+            self.instrument = instrument
+
+            # Identify which database tables to use
+            self.identify_tables()
+
+            # Get a list of all possible apertures for this instrument
+            siaf = Siaf(self.instrument)
+            possible_apertures = list(siaf.apertures)
```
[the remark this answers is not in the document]
bsunnquist: Yeah it's really handy - I know at least the Bias Monitor uses it as well, and I think the Dark Monitor too

jwql/instrument_monitors/common_monitors/readnoise_monitor.py
bsunnquist: I changed the variable name in this function to counts - I wasn't sure if you were also wanting all the references to `n` changed as well, e.g. the `amp1_n` database column, but let me know if so and I'll make those changes as well.

jwql/instrument_monitors/common_monitors/readnoise_monitor.py
bsunnquist: I changed to using the one that exists already in pipeline_tools - previously, I didn't use that one because I wanted to run custom settings for some steps (e.g. odd_even_rows=False for the refpix step), but I think this will be taken care of now that the pipeline is using these new parameter reference files to correctly set the defaults (i.e. NIRCam files will run the refpix step with that setting by default now)

jwql/instrument_monitors/common_monitors/readnoise_monitor.py
```diff
@@ -0,0 +1,625 @@
...
+
+        # Calculate the readnoise by taking the clipped stddev through the CDS stack
+        logging.info('\tCreating readnoise image')
+        clipped = sigma_clip(cds_stack, sigma=3.0, maxiters=3, axis=0)
+        readnoise = np.std(clipped, axis=0)
+        readnoise = readnoise.filled(fill_value=np.nan)  # converts masked array to normal array and fills missing data
+
+        return readnoise
+
+    def most_recent_search(self):
```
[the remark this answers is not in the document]
bsunnquist: Hmm maybe, I hesitate though because I'm not sure how generalized the other monitors are for how/if they handle their query history entries, column names/end time units, etc.

bsunnquist:
@bourque I resolved most of these and did another test run and everything looked good. Let me know if you want anything else from those remaining comments.

jwql/instrument_monitors/common_monitors/readnoise_monitor.py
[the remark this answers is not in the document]
bourque: Ok.  Perhaps something to keep in mind for a future PR.

bourque:
Thanks @bsunnquist.  The code looks great.  I will test things out now.

PR #574, opened 2020-05-29
Anomaly results web app view

tnking97:
Preliminary forms, templates, and views for querying anomalies.


jwql/website/apps/jwql/api_views.py
```diff
@@ -206,6 +207,28 @@ def preview_images_by_rootname(request, rootname):
...
 
+def thumbnails_all_instruments(request, inst_list):
+    """Return a list of available thumbnails in the filesystem for the
+    given request
+
+    Parameters
+    ----------
+    request : HttpRequest object
+        Incoming request from the webpage
+    inst : list of strings
```
bourque, line 218: Looks like this should be `inst_list` instead of `inst`.  And could you change this to something like `instrument_list` (as it is a bit more explicit)?

jwql/website/apps/jwql/data_containers.py
```diff
@@ -790,6 +790,58 @@ def get_proposal_info(filepaths):
...
+    -------
+    thumbnails : list
+        A list of thumbnails available in the filesystem for the
+        given instrument.
+    """
+
+    # Make sure instruments are of the proper format (e.g. "Nircam")
+    thumbnail_list = []
+    for inst in instruments:  # JWST_INSTRUMENT_NAMES:
+        instrument = inst[0].upper()+inst[1:].lower()
```
bourque, line 807: It is unfortunately that MAST requires this naming convention, it can get messy!

jwql/website/apps/jwql/forms.py
```diff
@@ -45,18 +45,38 @@ def view_function(request):
...
 
 from astropy.time import Time, TimeDelta
 from django import forms
 from django.shortcuts import redirect
 from jwedb.edb_interface import is_valid_mnemonic
 
+# from data_containers import get_thumbnails_all_instruments
 from jwql.database import database_interface as di
-from jwql.utils.constants import ANOMALY_CHOICES, JWST_INSTRUMENT_NAMES_SHORTHAND
+from jwql.utils.constants import ANOMALY_CHOICES, FILTERS_PER_INSTRUMENT
```
bourque, line 57: Importing two things per line here seems a bit arbitrary and may make things difficult if we are trying to measure what/how many things we are importing from a repository-level, so I would suggest just importing one thing per line here. I think this is what PEP8 suggests too.

jwql/website/apps/jwql/forms.py
```diff
@@ -88,6 +108,69 @@ def update_anomaly_table(self, rootname, user, anomaly_choices):
...
+    def clean_anomalies(self):
+        anomalies=self.cleaned_data['anomaly_choices']
+        return anomalies
+
+
+class ApertureForm(forms.Form):
+    """Creates an ``ApertureForm`` object that allows for ``aperture``
+    input in a form field.
+    """
+    aperturelist = []
```
bourque, line 120: PEP 8 suggests variable names like these use an underscore, so something like `aperture_list`

jwql/website/apps/jwql/forms.py
```diff
@@ -380,3 +532,19 @@ class MnemonicExplorationForm(forms.Form):
...
+    modelist=[]
+    for instrument in OBSERVING_MODE_PER_INSTRUMENT.keys():  # Add AND in instruments chosen
+        modes_per_inst = OBSERVING_MODE_PER_INSTRUMENT[instrument]
+        for mode in modes_per_inst:
+            modelist.append([mode, mode]) if [mode, mode] not in modelist else modelist
+    mode = forms.MultipleChoiceField(required=False, choices=modelist, widget=forms.CheckboxSelectMultiple)
+
+    def clean_modes(self):
+        modes=self.cleaned_data['mode']
+        return modes
```
bourque, line 550: Great work on adding all of these forms!  I'm still wondering if we should be using one gigantic form, or many smaller ones as you have here ... Something that we can research and figure out for the next iteration updates.

jwql/website/apps/jwql/urls.py
```diff
@@ -74,6 +74,10 @@
     path('about/', views.about, name='about'),
     path('dashboard/', views.dashboard, name='dashboard'),
     path('edb/', views.engineering_database, name='edb'),
+    path('query_anomaly/', views.query_anomaly, name='query_anomaly'),
+    path('query_anomaly_2/', views.query_anomaly_2, name='query_anomaly_2'),
+    path('query_anomaly_3/', views.query_anomaly_3, name='query_anomaly_3'),
+    path('query_submit/', views.query_submit, name='query_submit'),
```
bourque, line 80: These all look and work great in the browser, but this again points out to me that there is likely a single form/HTML page solution out there, instead of having to divide things up in multiple pages/forms.  Something to think about in the future 🤔

jwql/website/apps/jwql/views.py
```diff
@@ -47,16 +48,27 @@
...
+from .data_containers import get_thumbnails_all_instruments
 from .data_containers import get_proposal_info
 from .data_containers import random_404_page
 from .data_containers import thumbnails_ajax
 from .data_containers import data_trending
 from .data_containers import nirspec_trending
-from .forms import AnomalySubmitForm, FileSearchForm
+from .forms import AnomalyForm, AnomalySubmitForm, ApertureForm, EarlyDateForm
+from .forms import ExptimeMaxForm, ExptimeMinForm, FileSearchForm, FiletypeForm
+from .forms import FilterForm, InstrumentForm, LateDateForm, ObservingModeForm
```
bourque, line 59: Would also suggest to import one thing per line here

jwql/website/apps/jwql/views.py
```diff
@@ -47,16 +48,27 @@
...
 from .oauth import auth_info, auth_required
-from jwql.utils.constants import JWST_INSTRUMENT_NAMES, MONITORS, JWST_INSTRUMENT_NAMES_MIXEDCASE
+from jwql.utils.constants import ANOMALIES_PER_INSTRUMENT, FILTERS_PER_INSTRUMENT
+from jwql.utils.constants import FULL_FRAME_APERTURES, JWST_INSTRUMENT_NAMES
+from jwql.utils.constants import MONITORS, JWST_INSTRUMENT_NAMES_MIXEDCASE
+from jwql.utils.constants import OBSERVING_MODE_PER_INSTRUMENT
```
bourque, line 64: and here

jwql/website/apps/jwql/views.py
```diff
@@ -402,6 +414,209 @@ def not_found(request, *kwargs):
     return render(request, template, context, status=status_code)
 
 
+def query_anomaly(request):
```
bourque, line 417: Great work on these views.  Very nicely constructed!

jwql/website/apps/jwql/api_views.py
tnking97: This function is not currently being used, so I removed it (at least temporarily), but I will make those changes if it is added back in!

jwql/website/apps/jwql/forms.py
tnking97: Thank you! I'm working on adding in the multi-widget form; I'll email you with more detailed updates on that.

jwql/website/apps/jwql/views.py
tnking97: Thank you!

PR #601, opened 2020-07-24
Add bad pixel monitor tests

### LINKED ISSUE #592: Build tests for Bad Pixel Monitor

bhilbert4:
This PR adds tests for the bad pixel monitor.

Resolves #592


bhilbert4:
Hmm. Tests failed because 'no module named jwst_reffiles'. I'm not sure why it wasn't installed.

bhilbert4:
From the Jenkins log:
```
Reading git+https://github.com/spacetelescope/jwst#0.16.2
Download error on git+https://github.com/spacetelescope/jwst#0.16.2: unknown url type: git+https -- Some packages may not be found!
Reading git+https://github.com/spacetelescope/jwst_reffiles
Download error on git+https://github.com/spacetelescope/jwst_reffiles: unknown url type: git+https -- Some packages may not be found!
```

bhilbert4:
I'm not sure why there is a failing test for this. The error statement says that when trying to import `jwst_reffiles` that there is no package by that name. But I also  see in the environment creation process where jwst_reffiles is successfully installed.

bhilbert4:
I don't see the `git+https` error now when I look, unless I was just looking in the wrong place.

bourque:
@bhilbert4 Is this ready for review or still a WIP?

bhilbert4:
This is ready for review. The only issue was the strange Jenkins failures with the environment creation. Looking at the latest test results, it appears to construct the environment with no problems, but then complains that `jwst_reffiles` is not installed, which is causing the new tests to fail.

bourque:
@bhilbert4 Ok thanks.  Maybe those issues will fix themselves when I update the branch!

bourque:
Sadly it didn't ...

bhilbert4:
No such luck. It looks like it's still tripping on `git+https`:
```Reading git+https://github.com/spacetelescope/jwst@0.16.2
Download error on git+https://github.com/spacetelescope/jwst@0.16.2: unknown url type: git+https -- Some packages may not be found!
Reading git+https://github.com/spacetelescope/jwst_reffiles
Download error on git+https://github.com/spacetelescope/jwst_reffiles: unknown url type: git+https -- Some packages may not be found!
```

bhilbert4:
The Jenkins log for #615 shows the same error statement, but somehow it wasn't a problem then. Maybe because at that point there were no test that call any part of `jwst_reffiles`? This would also imply that we don't have any tests that call `jwst`.

bhilbert4:
This looks like an old question and answer, but maybe this is what we need?
https://en.it1352.com/article/58557b86c2ec4b5eba17e357caca3b57.html

bhilbert4:
The previous commit seems to have gotten rid of the `git+https` error in the Jenkins log, but the tests are still failing, saying that `jwst_reffiles` is not installed.

jwql/tests/test_badpix_monitor.py
```diff
@@ -0,0 +1,216 @@
...
+    mask[bad2_only] = True
+
+    diff = bad_pixel_monitor.exclude_crds_mask_pix(bad2, bad1)
+    assert np.all(diff[mask] == 4)
+    assert np.all(diff[~mask] == 0)
+
+
+@pytest.mark.skipif(ON_JENKINS,
+                    reason='Requires access to central storage.')
+def test_locate_rate_files():
```
tnking97, line 117: Perhaps these test_locate_* functions could be moved down a bit in order to alphabetize all of the function definitions?

jwql/tests/test_badpix_monitor.py
```diff
@@ -0,0 +1,216 @@
...
+nis_list = ['NIS_CEN']
+nrs_list = ['NRS1_FULL', 'NRS2_FULL']
+miri_list = [('MIRIMAGE', 'MIRIM_FULL'), ('MIRIFULONG', 'MIRIM_FULL'), ('MIRIFUSHORT', 'MIRIM_FULL')]
+fgs_list = ['FGS1_FULL', 'FGS2_FULL']
+
+
+@pytest.mark.parametrize("instrument,expected_list", [("nircam", nrc_list), ("niriss", nis_list),
+                                                      ("nirspec", nrs_list), ("miri", miri_list), ("fgs", fgs_list)])
+def test_get_possible_apertures(instrument, expected_list):
+    """Make sure the correct apertures are returned for the given instrument
```
tnking97, line 188: Maybe a parameters section could be helpful here?

tnking97 (commented):
Hi @bhilbert4, this looks great! It seems very clear and addresses a variety of tests. I added a few comments.

jwql/tests/test_badpix_monitor.py
bhilbert4: Oops. Fixed this.

jwql/tests/test_badpix_monitor.py
bhilbert4: I'm not sure what you mean here? I figured it was best to keep the creation of e.g. `nrc_list` close to `test_possible_apertures` since that is the only test that uses it.

tnking97: Okay, yes that makes sense!

tnking97 (approved):

bourque:
Thanks for reviewing this @tnking97!

@bhilbert4 Could you also add the appropriate sphinx docs for this PR?

bhilbert4:
@bourque I don't understand why the API docs aren't working for this. I added those lines in to common_monitors.rst in the original bad pixel monitor PR. But it looks like, other than the section header, the documents aren't being created. The code looks identical to that from the other monitors. Can you see anything that's missing in there?

bourque:
I am at a loss here.  I played around with the Jenkins build a bit and explicitly did a clone and install of the `jwst_reffiles` package.  I even confirmed that the installation worked and that the software environment can import it.  But the test still complains that the module is not found. 🤷 

I think for now we should just skip this test when running on Jenkins.  We are pouring too many hours into solving this currently.  I will add a skip decorator and we can get this PR merged in.

bourque:
@bhilbert4 

> @bourque I don't understand why the API docs aren't working for this. I added those lines in to common_monitors.rst in the original bad pixel monitor PR. But it looks like, other than the section header, the documents aren't being created. The code looks identical to that from the other monitors. Can you see anything that's missing in there?

Hmm, it could be related to the `jwst_reffiles` problem.  If the module is unable to be imported by the sphinx build process, then the docs will remain blank.  But I was actually referring to adding the API docs for the tests themselves (i.e. `tests.rst`).  It was a simple addition so I just took care of it while I was trying to get the CI builds to work, so don't worry about it.  We can address the blank documentation in a different PR.  I imagine there are others that are like that too, so it would be good to do a 'fix all RtD documentation' PR at some point before out 1.0 release.

PR #602, opened 2020-07-24
Add newly-created apertures from pysiaf 0.9.0

### LINKED ISSUE #524: Dark Monitor crashing with IndexError
See traceback from recent log file:

```
11/12/2019 15:40:32 PM INFO: Working on aperture NRCA5_TAGRISMTS32_F405N in nircam
11/12/2019 15:40:32 PM INFO: 	No query history for NRCA5_TAGRISMTS32_F405N. Beginning search date will be set to 57357.0.
11/12/2019 15:40:32 PM INFO: 	Query times: 57357.0 58799.8603127774
11/12/2019 15:40:32 PM INFO: 	Aperture: NRCA5_TAGRISMTS32_F405N, new entries: 0
11/12/2019 15:40:32 PM CRITICAL: Traceback (most recent call last):
  File "/home/jwqladm/repositories/jwql/jwql/utils/logging_functions.py", line 279, in wrapped
    func(*args, **kwargs)
  File "/home/jwqladm/repositories/jwql/jwql/utils/logging_functions.py", line 244, in wrapped
    func(*args, **kwargs)
  File "dark_monitor.py", line 797, in run
    len(new_entries), instrument, aperture, file_count_threshold[0]))
  File "astropy/table/_column_mixins.pyx", line 74, in astropy.table._column_mixins._ColumnGetitemShim.__getitem__
  File "astropy/table/_column_mixins.pyx", line 57, in astropy.table._column_mixins.base_getitem
  File "astropy/table/_column_mixins.pyx", line 69, in astropy.table._column_mixins.column_getitem
IndexError: index 0 is out of bounds for axis 0 with size 0
```

bhilbert4:
This PR adds several newly-created apertures from the latest version of pysiaf to the dark_monitor_file_thresholds.txt file. When JWQL encountered a dark current file using one of these apertures, it crashed with an index error because there was no entry for it in the thresholds file. 

I'm not sure how best to keep this up-to-date in the future. I don't think new apertures will be added very frequently going forward, but I'm sure there will be some from time to time. Currently the thresholds file basically specifies that an full frame aperture needs 10 files in order for the dark monitor to run, while subarray apertures need 30. These values are just best guesses on what is needed to get decent signal-to-noise in a mean dark. We could replace the threshold file with a general rule similar to what's described above. Then aperture names would be grabbed from pysiaf and we shouldn't miss any new names any more. But this then limits the ability of teams to customize how many darks they think should be present before running the monitor. On NIRCam for example, subarrays range in size from 640x640 to 64x64. With 1/100th the exposure time, the 64x64 array probably should wait for many more files than the 640x640 before triggering the dark monitor.

In the mean time, this PR brings the threshold file up to date. There were several missing NIRCam apertures, and one missing NIRISS aperture. All other instruments were up to date.

Resolves #524


bourque:
Thanks @bhilbert4 for the change and providing some background information.  This reads to me like: even if we decide on a rule/path forward right now, there is a decent change of things changing once JWST is in operations and we understand the darks more.  So I am thinking that it is not worth worrying about at this point in time(?).

Issue #526, opened 2019-11-12
Add support for Python 3.7 and 3.8


bourque:
The `jwst` pipeline is currently developing in Python 3.7, and since the JWQL instrument monitors are highly dependent on the pipeline, perhaps we should also be supporting Python 3.7.

PR #603, opened 2020-07-25
Add support for new python versions

bhilbert4:
This PR adds support for new python versions (3.7, and hopefully 3.8).

- [x] Add new environment file for 3.7
- [x] Update Jenkins file to run tests under 3.7
- [x] Address any failing tests under 3.7
- [x] Add new environment file for 3.8
- [x] Update Jenkins file to run tests under 3.8
- [x] Address any failing tests under 3.8


bourque:
Closes #526

bhilbert4:
Jenkins is failing when it installs jsonschema 2.6 (which is quite old. 3.2 is current), and then runs into a problem when ASDF requires a more recent version. However, there is another issue which may be more important. Jenkins does not install `jwst` because: `unknown url type: git+https -- Some packages may not be found!`

bhilbert4:
Also, pysiaf currently does not support python 3.8, which is why I was getting unsolvable conflicts for that.

bhilbert4:
That previous commit is mostly just to restart Jenkins and see if we get the same `git+https` error

bhilbert4:
@bourque all tests are now passing. I tweaked some versions in setup.py and requirements.txt. The `git+https` lines are still present. I'm not entirely sure why they didn't give an error this time around.

bourque:
I tested the 3.7 env by running all of our unit tests locally.  There were some failures (e.g. #619), but they are unreleated to the environment I believe.  Will try the same for 3.8 next.

bourque:
Alright this should be good to go pending a successful Jenkins build!

Issue #504, opened 2019-09-10
Get tests running on Internal Jenkins server


bourque:
We currently have several tests that require access to central storage and thus cannot be run on our existing Jenkins server, since it does not have access (for example, [this test](https://github.com/spacetelescope/jwql/blob/develop/jwql/tests/test_utils.py#L276)).  Currently, we avoid running these tests altogether if they are executed from the Jenkins server (i.e. if `ON_JENKINS` is `True`).

DATB has set up a separate Jenkins server that does have access to central storage (#144), however currently I don't believe we actually have anything building/testing on it.

We should try to get tests running (and passing!) on this internal Jenkins server, and make the results available somehow. I believe the [`JenkinsfileRT`](https://github.com/spacetelescope/jwql/blob/develop/JenkinsfileRT) file contains the code to run things on there.

jbcurtin:
We have a few different JENKINS runtimes, I'd like to change ON_JENKINS to an enum. What do you think?

```
class TestingEnviornemnt:
  CentralStorage: str = 'central'
  Dev: str = 'dev'
  Nightly: str = 'nightly'

@pytest.mark.skipif(ON_JENKINS == TestEnvironment.CentralStorage, reason='Requires access to central storage')
def test_make_log_file():
    ...logic...

```

bourque:
@jbcurtin Sounds good to me!

bourque:
We have now removed our use of Jenkins altogether and replaced with GitHub Actions (#698), so I don't think we need to explore this any further.

PR #615, opened 2020-08-12
Make readnoise tests

bsunnquist:
Added tests for the readnoise monitor


bourque:
This might be the first time I've ever seen @pep8speaks come back with no issues!  Congrats!

bsunnquist:
Thanks @mfixstsci! Yes I believe that's the case, since jenkins can't access central storage to get the files needed for those tests (I copied that strategy from `test_dark_monitor.py`).

mfixstsci:
Thanks @bsunnquist, just wanted to double check. Might be a better question for @bourque but, do we plan on adding these files to `Jenkins` in the future? Or do we have a regression testing suite set up?

bourque:
@mfixstsci @bsunnquist  Yeah, we do have plans to spin up an internal Jenkins server that would have access to central storage and thus be able to run all of those tests (https://github.com/spacetelescope/jwql/issues/504), but it is not yet implemented, so for now we just ignore those tests in the current Jenkins build.

mfixstsci:
@bourque sounds good. When we are ready to set that infrastructure up, I would be willing to help since I've done a fair bit of it for the JWST Pipeline unit/regression testing suite. Besides that @bsunnquist everything else looks good!

PR #631, opened 2020-09-18
Update top level instrument JDox links

bhilbert4:
This PR updates the JDox links on our instrument pages to point to the current top-level pages. The links we previously had in there were no longer correct.


bourque (approved):
Thanks @bhilbert4!

PR #635, opened 2020-09-23
Updates to bokeh-templating

gkanarek:
This is my current functional version of `bokeh_templating`.


gkanarek:
@bourque bokeh_surface is currently broken. should I just remove it? is anyone using it right now? we may or may not want a version that works in the future, but at the moment it's nonfunctional.

bourque:
@gkanarek I think chances are relatively low that people will want to use `bokeh_surface`, at least in the near term, so I endorse its removal!

bourque (approved):
Thanks for these updates @gkanarek!

PR #636, opened 2020-09-23
Cosmic Ray Monitor

mengesser:
Adding cosmic ray monitor, as well as unit test code and relevant database files. The code currently has MAST retrieval commented out for testing purposes. I can provide the test files and my copy of the config.json file for anyone doing review.

PR #638, opened 2020-10-01
[WIP] Building Cosmic Ray Monitor Plots

mengesser:
Work in progress. Includes all effort toward developing CR monitor bokeh plots. Needs debugging. Also includes changes from PR #636 because I didn't branch properly and can't figure out how to separate them. This is okay, because this fixes an issue with #636 in the database_interface.py file.


bourque:
Since this includes changes already in #636, we should merge in #636 first, then update this branch, and this this PR should reflect the changes specific to this PR.

bhilbert4:
I just did a quick comparison of the files in this PR to those in #700 and they look to be largely identical. So I think it's safe to consider #700 as a replacement for this PR, just as it is for #636. Any further edits to the cosmic ray monitor plots should be done in #700.

bhilbert4:
Now that #700 has been merged, I don't think we need this any more. Closing.

PR #612, opened 2020-08-11
 Add bad pixel monitor Bokeh plots

bhilbert4:
Resolves #588 

This PR includes code to create plots and tables for the web app from the results of the bad pixel monitor. 

Necessary file framework only at this point. Still lots of significant updates to be made.

PR #643, opened 2020-10-13
Fix dark monitor database query

### LINKED ISSUE #637: Fix most_recent_search() in dark monitor
I've just found that most_recent_search() in the bad pixel monitor is not functioning the way I anticipated. It seems to be not filtering correctly, and is not finding results when it should. This is leading to problems in the database tables, and later on, the Bokeh plots. I have a fix that I think should work for the bad pixel monitor. But since the original function in there was based on most_recent_search() in the dark current monitor, I should go fix the dark current monitor as well.

Somehow the subquery wasn't working, so this is the new code I used:

```
query = session.query(self.query_table).filter(self.query_table.aperture==self.aperture). \
                                                filter(run_field==True)

        dates = np.zeros(0)
        if file_type.lower() == 'dark':
            for instance in query:
                dates = np.append(dates, instance.dark_end_time_mjd)
        elif file_type.lower() == 'flat':
            for instance in query:
                dates = np.append(dates, instance.flat_end_time_mjd)
```

bhilbert4:
This PR fixes an issue with the most_recent_search() method that was first discovered when working on #612 . The database query used to find the date of the most recent run of the monitor for a given aperture was not working correctly. This fix updates the method to work in a similar way to the corresponding function in the bad pixel monitor. 

A quick test with a reset and custom-populated dark monitor database table showed that the new code returns the expected value.

Resolves #637

### DIFF
--- a/jwql/instrument_monitors/common_monitors/dark_monitor.py
+++ b/jwql/instrument_monitors/common_monitors/dark_monitor.py
@@ -441,31 +441,20 @@ class Dark():
             Date (in MJD) of the ending range of the previous MAST query
             where the dark monitor was run.
         """
+        query = session.query(self.query_table).filter(self.query_table.aperture==self.aperture). \
+                                                filter(self.query_table.run_monitor==True)
 
-        sub_query = session.query(self.query_table.aperture,
-                                  func.max(self.query_table.end_time_mjd).label('maxdate')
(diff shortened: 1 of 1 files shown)


bourque (approved):
Thanks @bhilbert4, looks good!

PR #644, opened 2020-10-15
Update django OAuth import path

lepture:
This path has been changed since v0.13.

Here is the compatible change guide:

https://gist.github.com/lepture/726e07a737b20f111d570d1d0437d5d4#file-rn-md


bourque:
This will be fixed through #645

Issue #625, opened 2020-09-02
test_utils failing


bsunnquist:
```
(jwql-3.6) bash-3.2$ pytest -vv test_utils.py
===================================================================================== test session starts ======================================================================================
platform darwin -- Python 3.6.4, pytest-4.6.2, py-1.8.0, pluggy-0.12.0 -- /Users/bsunnquist/miniconda3/envs/jwql-3.6/bin/python
cachedir: .pytest_cache
metadata: {'Python': '3.6.4', 'Platform': 'Darwin-17.7.0-x86_64-i386-64bit', 'Packages': {'pytest': '4.6.2', 'py': '1.8.0', 'pluggy': '0.12.0'}, 'Plugins': {'openfiles': '0.3.2', 'cov': '2.6.1', 'arraydiff': '0.3', 'metadata': '1.8.0', 'html': '1.19.0', 'doctestplus': '0.3.0', 'remotedata': '0.3.1'}}
rootdir: /Users/bsunnquist/Documents/repos/jwql, inifile: pytest.ini
plugins: openfiles-0.3.2, cov-2.6.1, arraydiff-0.3, metadata-1.8.0, html-1.19.0, doctestplus-0.3.0, remotedata-0.3.1
collected 24 items                                                                                                                                                                             

test_utils.py::test_copy_files PASSED                                                                                                                                                    [  4%]
test_utils.py::test_get_config PASSED                                                                                                                                                    [  8%]
test_utils.py::test_filename_parser[/test/dir/to/the/file/jw90002/jw90002001001_02102_00001_nis_rateints.fits-solution0] PASSED                                                          [ 12%]
test_utils.py::test_filename_parser[jw00327001001_02101_00002_nrca1_rate.fits-solution1] PASSED                                                                                          [ 16%]
test_utils.py::test_filename_parser[jw00327001001_02101_00002_nrca1-solution2] PASSED                                                                                                    [ 20%]
test_utils.py::test_filename_parser[jw94015002002_02108_00001_mirimage_o002_crf.fits-solution3] PASSED                                                                                   [ 25%]
test_utils.py::test_filename_parser[jw90001001003_02101_00001_nis_o001-solution4] PASSED                                                                                                 [ 29%]
test_utils.py::test_filename_parser[jw80600-o009_t001_miri_f1130w_i2d.fits-solution5] PASSED                                                                                             [ 33%]
[...]

PR #654, opened 2020-10-26
Fix test_utils

bsunnquist:
This fixes https://github.com/spacetelescope/jwql/issues/625.

The problem was one of the files in our filesystem (jw04192001001_01101_00001_MIRIMAGE_uncal.fits) had an uppercase detector name. I made a simple edit to `utils.py` to make it robust to this when searching for the matching instrument.


bsunnquist:
@bourque There is also this random hidden file the causes test_utils to fail:
`/grp/jwst/ins/jwql/filesystem/jw04192/._jw04192001001_01101_00001_MIRIMAGE_uncal.fits`

I was going to remove it, but figured I'd check with you first in case it was important.

cracraft:
As the creator of the file (using a script written in SCSB to convert ground test data to pipeline format), I can't see any need for a hidden version of the file. I assume it's the same as the non-hidden version. But since I haven't developed code, also good to check with Matt.

bourque:
@bsunnquist @cracraft Yeah I agree, we should be fine to remove this!

bourque:
Alright I went ahead and removed that file.  I am also going to redact the filepath from the comments for security purposes.

bourque (approved):
Thanks @bsunnquist!

Issue #623, opened 2020-09-02
test_data_containers failing


bsunnquist:
```
(jwql-3.6) bash-3.2$ pytest -vv test_data_containers.py
===================================================================================== test session starts ======================================================================================
platform darwin -- Python 3.6.4, pytest-4.6.2, py-1.8.0, pluggy-0.12.0 -- /Users/bsunnquist/miniconda3/envs/jwql-3.6/bin/python
cachedir: .pytest_cache
metadata: {'Python': '3.6.4', 'Platform': 'Darwin-17.7.0-x86_64-i386-64bit', 'Packages': {'pytest': '4.6.2', 'py': '1.8.0', 'pluggy': '0.12.0'}, 'Plugins': {'openfiles': '0.3.2', 'cov': '2.6.1', 'arraydiff': '0.3', 'metadata': '1.8.0', 'html': '1.19.0', 'doctestplus': '0.3.0', 'remotedata': '0.3.1'}}
rootdir: /Users/bsunnquist/Documents/repos/jwql, inifile: pytest.ini
plugins: openfiles-0.3.2, cov-2.6.1, arraydiff-0.3, metadata-1.8.0, html-1.19.0, doctestplus-0.3.0, remotedata-0.3.1
collected 18 items                                                                                                                                                                             

test_data_containers.py::test_get_acknowledgements PASSED                                                                                                                                [  5%]
test_data_containers.py::test_get_all_proposals PASSED                                                                                                                                   [ 11%]
test_data_containers.py::test_get_dashboard_components XPASS                                                                                                                             [ 16%]
test_data_containers.py::test_get_expstart PASSED                                                                                                                                        [ 22%]
test_data_containers.py::test_get_filenames_by_instrument PASSED                                                                                                                         [ 27%]
test_data_containers.py::test_get_filenames_by_proposal PASSED                                                                                                                           [ 33%]
test_data_containers.py::test_get_filenames_by_rootname PASSED                                                                                                                           [ 38%]
test_data_containers.py::test_get_header_info FAILED                                                                                                                                     [ 44%]
test_data_containers.py::test_get_image_info PASSED                                                                                                                                      [ 50%]
test_data_containers.py::test_get_instrument_proposals PASSED                                                                                                                            [ 55%]
test_data_containers.py::test_get_preview_images_by_instrument PASSED                                                                                                                    [ 61%]
test_data_containers.py::test_get_preview_images_by_proposal PASSED                                                                                                                      [ 66%]
[...]

PR #655, opened 2020-10-26
Fix test_data_containers

bsunnquist:
This fixes https://github.com/spacetelescope/jwql/issues/623.

`get_header_info` expects the filename without the fits extension, and the returned header info is in a dictionary (not a string). I've updated the docstring in `get_header_info` to reflect this as well.


bourque (approved):
Thanks @bsunnquist!  Looks good.  I used this PR as an opportunity to add the `wtforms` dependency too.

PR #662, opened 2020-11-06
Switch EDB MAST token authentication to use config.json

### LINKED ISSUE #661: Switch EDB to use MAST token from config.json always
Currently the EDB feature of the JWQL web app attempts to authenticate users with whichever MAST token is currently cached (https://github.com/spacetelescope/jwql/blob/develop/jwql/utils/credentials.py#L45), and if that doesn't succeed, _then_ it uses the `mast_token` key in the `config.json` file.  This seems problematic if users are creating new tokens but then attempting to perform EDB queries in the same browser.  We should probably switch this to just use the `mast_token` key in `config.json` file always.

bourque:
Closes #661

### DIFF
--- a/jwql/utils/credentials.py
+++ b/jwql/utils/credentials.py
@@ -41,11 +41,6 @@ def get_mast_token(request=None):
         print('Authenticated with Astroquery MAST magic')
         return None
     else:
-        if request is not None:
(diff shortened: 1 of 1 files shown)


bhilbert4 (approved):
This update worked for me! I can now log in and run the queries on the EDB page

PR #664, opened 2020-11-12
Datatable initialization error when viewing headers in web app

### LINKED ISSUE #531: Fix DataTables warning when viewing headers in web app
When viewing a header for a file in the web app, we encounter this error:

Screen Shot 2020-01-28 at 12 08 27

bhilbert4:
Resolves #531 

This PR makes changes to eliminate the pop-up warning that appears when viewing fits file headers in the web app.


bhilbert4:
At commit 7f92d98, I've replaced the original method for making the header html (astropy table -> write to temp file in jsviewer format -> read in table) with the recently added table viewer code.

At the moment it appears to be working pretty well. The one issue I'd like to solve is that the new method alphabetizes the header keywords. It would be nicer to keep them in the original order.

bhilbert4:
This is ready for review, modulo whatever is causing the doc build to fail. The header pop-up warnings have disappeared, although in the course of this work I've discovered some deeper issues with the header display code, that will be addressed in a separate PR.

bourque (approved):
Thank you @bhilbert4, this looks good!  And I see no issues with displaying the headers now, and no signs of the pop-up/warning!

Issue #648, opened 2020-10-20
Configure new dev, test, and ops service accounts


bourque:
We now have separate dev, test, and ops service accounts for our various VMs (#585).  All of these accounts need to be configured such that:

- [x] The `jwql` environment(s) and package are all installed in the accounts

- [x] The web app is using the `jwql-3.7` conda environment

- [x] The web app works without errors from the dev server

- [x] Each account has a working `crontab`
    - [x] The `dev` account will run monitors from whatever the current branch is (by default this will be `develop` but could be any branch that people are developing on)
    - [x] The `test` account will run monitors from the `develop` branch always
    - [x] The `ops` account will run monitors from the `master` branch always
    - [x] The scripts that the cron jobs are completing successfully

- [x] Each account uses a separate `config.json` file
    - [x] The `dev` account will point to the dev database and `auth.masttest` authentication
    - [x] The `test` account will point to the test database and `auth.masttest` authentication 
    - [x] The `ops` account will point to the ops database and `auth.mast` authentication

- [x] Each account saves output products in separate `outputs` directory
   - [x] The `dev` account will save output products to `outputs/dev/`
   - [x] The `test` account will save output products to `outputs/test/`
   - [x] The `ops` account will save output products to `outputs/ops/`

- [x] Each account's log files will be owned by their appropriate service account and written to a separate `logs` directory
   - [x] The `dev` account will save logs to `logs/dev/` and will be owned by the `dev` service account
   - [x] The `test` account will save logs to `logs/test/` and will be owned by the `test` service account
   - [x] The `ops` account will save logs to `logs/ops/` and will be owned by the `ops` service account

- [x] Each account will store data in the appropriate database instance
    - [x] The `dev` account will store data in the `dev` database on the `dev` database server
    - [x] The `test` account will store data in the `test` database on the `dev` database server
    - [x] The `ops` account will store data in the `ops` database on the `ops` database server

- [x] Ask IT to open up read permissions for the read only account on the dev database
- [x] Ask IT to open up read permissions for the read only account on the test and ops databases
- [x] Ask IT to set up a test web server

bourque:
I did a check on the cron jobs running on the dev server.  The `jwql` monitors (i.e. filesystem, mast, cron job, and preview images) are all working, as well as the Bad Pixel Monitor (for all instruments).  Here are the current issues:

- Dark monitor is crashing (should be addressed by #618 and #600 )
[...]

PR #670, opened 2020-12-07
Fix issues caused by readnoise monitor webpage

bsunnquist:


bourque (approved):
Looks good, thanks for the prompt fix @bsunnquist!

bourque:
Looks like the Jenkins build is only failing because of the current work going on with the dev server (#648), so I will go ahead and merge.

PR #671, opened 2020-12-08
Fix readnoise webpage hanging issue

bsunnquist:
This simple edit should fix the issue of the readnoise monitor accessing the database when just starting the web app, which was resulting in "hanging" database processes when running the web app. 

Now, it only accesses the database when you actually go to the readnoise monitor webpage and the database processes go away when you navigate to another monitor webpage, which I verified is the same behavior of other monitor webpages (e.g. the bad pixel webpage).


bourque:
Thanks for this fix @bsunnquist!

bourque (approved):

PR #674, opened 2020-12-11
Fix MIRI amp bounds calculation when omitting refpix

bsunnquist:
Previously, the command
`_, amp_bounds = instrument_properties.amplifier_info('miri_image.fits', omit_reference_pixels=True)`

returned equal boundaries for all amps for full-frame MIRI images, e.g.:

```
{'1': [(4, 1028, 4), (0, 1024, 1)],
 '2': [(4, 1028, 4), (0, 1024, 1)],
 '3': [(4, 1028, 4), (0, 1024, 1)],
 '4': [(4, 1028, 4), (0, 1024, 1)]}
```

This PR updates this function to ensure the x start values remain interleaved for MIRI when omitting refpix, and the results of that command are now as expected, e.g.:

```
{'1': [(4, 1028, 4), (0, 1024, 1)],
 '2': [(5, 1028, 4), (0, 1024, 1)],
 '3': [(6, 1028, 4), (0, 1024, 1)],
 '4': [(7, 1028, 4), (0, 1024, 1)]}
```

These changes only affect the results for MIRI images with refpix, and only when running with `omit_reference_pixels=True`.


bsunnquist:
@bhilbert4 would you have time to review this PR? I think you're most familiar with this function. Otherwise, I can check with the slack channel if you're too busy.

My changes assume all MIRI subarrays are readout with the interleaving amp regions (i.e. the same as the full-frame), which I believe is the case but wasn't 100% sure.

bhilbert4:
Maybe @cracraft can confirm the subarray readout details?

bhilbert4 (approved):
Looks good to me!

cracraft:
Our amplifiers are interleaved such that it reads amp1, amp2, amp3, amp4, so the change looks correct to me.

cracraft (approved):
This code seems to correctly make sure the amps are interleaved for MIRI data.

bourque:
Thank you all for chiming in on this!  I will merge this pending a successful Jenkins build.

PR #677, opened 2020-12-14
Update bokeh to latest version

### LINKED ISSUE #632: Update Bokeh to latest version
I remember there was some reason that we were holding off on upgrading Bokeh from 1.3.4. However, Bokeh is now up to version 2.2.1 I believe. We should look into upgrading the version used for JWQL in order to take advantage of new features and so that we minimize the number of plots created under 1.3.4 which may need to be tweaked to work under the new version.

For example, one difference I ran into today was that the keyword "legend", which is used in 1.3.4 to denote the string printed in the legend for a particular element, has been changed to "legend_label" in version 2.2.1.

bourque:
Closes #632

### DIFF
--- a/environment_python_3_6.yml
+++ b/environment_python_3_6.yml
@@ -4,7 +4,7 @@ channels:
 dependencies:
 - astroquery=0.4
 - authlib=0.15.2
-- bokeh>=1.0,<1.4
+- bokeh>=2.2.3
 - django=2.2.5
 - drizzle=1.13
 - flake8=3.7.9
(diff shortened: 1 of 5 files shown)


bourque:
I tested the web app locally and I don't see any errors pertaining to these changes.  I'm just running all of the cron jobs on the dev server to make sure there are no errors in the scripts related to this.  Assuming everything is good there, this should be ready to merge.

bourque:
This is good to go

PR #678, opened 2020-12-14
Updated django to latest version

### LINKED ISSUE #626: Upgrade Django to 3.0
Django 3.0 is out, and since it is a major release, we should consider upgrading to this.

bourque:
Closes #626


bourque:
I tested this locally and there appears to be no issues!

Issue #538, opened 2020-03-04
Specify Bias Monitor plots


bourque:
We need to use the bokeh templating language to build up the plots for the bias monitor.

PR #680, opened 2020-12-22
Make Bias Monitor plots/webpage

bsunnquist:
This PR creates the Bias Monitor plots/webpage (currently only active for NIRCam).

### DIFF
--- a/jwql/website/apps/jwql/monitor_pages/yaml/monitor_bias_interface.yaml
+++ b/jwql/website/apps/jwql/monitor_pages/yaml/monitor_bias_interface.yaml
@@ -260,14 +260,14 @@
 - !ColumnDataSource: &cal_data
     ref: "cal_data"
     data:
-        dh: [1]
-        dw: [1]
+        dh: [0]
+        dw: [0]
         image: [[[0,0], [0, 0]]]
 - !Figure: &cal_image
     ref: "cal_image"
     title: 'Calibrated Zeroth Group of Most Recent Dark'
-    height: 700
-    width: 700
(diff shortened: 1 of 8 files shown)


bsunnquist:
@bourque this is all ready for review - here is what the webpage should look like:
Screen Shot 2020-12-22 at 6 40 38 PM

bourque:
Closes #538

jwql/website/apps/jwql/monitor_pages/yaml/monitor_bias_interface.yaml
```diff
@@ -0,0 +1,336 @@
+# - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - -
```
bourque, line 1: There seems to be a lot of repetition in this.  Is there no way to make a more generalized interface file?  Perhaps a question for @gkanarek 

Not a dealbreaker for this PR, but if there is a way to improve this, we should look into it in the future.

bourque (approved):
Thanks @bsunnquist, this looks excellent!

jwql/website/apps/jwql/monitor_pages/yaml/monitor_bias_interface.yaml
bsunnquist: yeah definitely - originally I had a general bias vs time figure that I updated in a loop for each amp/odd/even region, but it ended up being ~twice as slow as separating them like what's in this PR. But yeah, I'm sure there's a better way.

PR #688, opened 2021-01-28
Prevent dark monitor crash when database is empty

bhilbert4:
This is a small PR that adds some code to create dummy data for the dark current monitor in cases where the dark current databases are empty. Prior to these changes, if the databases were empty, JWQL would crash as it attempted to use the non-existent data to create the Bokeh plots.


bourque (approved):
Thanks @bhilbert4, this looks good to me!

Issue #676, opened 2020-12-14
Address speed issue for loading thumbails/preview images


bourque:
Navigating certain thumbnail/preview image pages can be painstankingly slow.  I suspect this is due to the fact that the web app has to look over a network in order to retrieve thumbnails/preview information, as everything is stored in the JWQL central storage area.  If that is indeed the cause, we should explore storing the thumbnails/preview images directly on the dev server on which the web app runs.

bourque:
I have created a support ticket to request internal storage space on the JWQL VMs to house preview images and thumbnails.

bourque:
We now have the internal storage space on the dev, test, and ops servers, and the preview images and thumbnails now reside there.  I notice a slight speedup in the web app in rendering the images, however, the loading of the archive web app pages is still quite slow.  We should look into if the algorithm for identifying/loading preview images/thumbnails can be improved.

PR #694, opened 2021-02-09
Generate all preview images in filesystem

tnking97:
Currently we do not generate preview images for all filenames. This change allows us to generate preview images for additional file types such as 'guider' or 'time_series'.


tnking97:
This particular bit of PR is ready for review. However, I think there is still some sort of permission error on the local host with viewing the preview image link that was mentioned in the linked issue. The images have been generated (and thus this issue should be fixed), but we may need to additionally test on the dev server.

bourque:
@tnking97 Perfect, thank you.  I will go ahead and test things out on the dev/tests servers.  It will also help me to test #676 at the same time!

bourque (approved):
Thanks again for this @tnking97!

I Just also wanted to note here that I (re)built all of the preview images & thumbnails on the dev server (in our new internal storage space).  It took nearly three days(!) but the preview image generation script completed successfully!  And I do notice a slight speedup in loading the images on the web app, though we will still need to address the speed of the algorithm itself (#676)

PR #696, opened 2021-02-11
Fix filename parsing for WFS&C and AMI products

bsunnquist:
This PR adds support for stage 3 WFS&C and additional AMI products in the filename parser. 

These types of products don't currently live in our JWQL filesystem (and thus, `test_filename_parser_whole_filesystem` passes regardless of this PR), but they will be eventually once we sync up with MAST. 

I've successfully tested it on the example names given [here](https://innerspace.stsci.edu/pages/viewpage.action?spaceKey=SCSB&title=JWST+Science+Data+Products) for these product types.


bourque (approved):
@bsunnquist Thank your for your updates on this!  This looks good, and works well on my end when I test it.

bourque:
The Jenkins build fails because there seems to be an issue with ST Jenkins and support for python 3.8.  Otherwise, the envs are installed ok and the tests pass.  So I will merge this.

PR #699, opened 2021-02-17
Adding Github Actions to JWQL

mfixstsci:
Adding github actions template with python 3.8. I plan on extending to python 3.7 and updating build matrix to loop through files to perform unit testing.


mfixstsci:
I currently have a successful build using github actions for JWQL. The most recent build can be seen here:

https://github.com/mfixstsci/jwql/actions/runs/579307416

One of the reasons I had to rearrange all of the imports in the tests and skip a majority of them is because we don't have access to central store for the monitor tests nor the development server for the web tests. Any imports/use of `get_config` in the monitors/tests would immediately stop the builds, which a majority of the monitors use. We skipped a lot of the tests prior to this PR but I am assuming the tests that I had to doctor to get github actions to pass were some that got through the cracks because Jenkins is able to access certain private facing parts of STScI's filesystem.

To Do's Before Merging
------------------------
- Pep8 Compliance
- Fix pytest skip comments
- Squash testing commits

mfixstsci:
@bourque this one is ready to go, there is a conflict because I removed the `Jenkinsfile` but that is expected :sweat_smile:

jwql/tests/test_api_views.py
```diff
@@ -73,6 +74,7 @@
     urls.append('api/{}/thumbnails/'.format(rootname))  # thumbnails_by_rootname
 
 
+@skipIf(ON_GITHUB_ACTIONS, "Can't access webpage without VPN access")
```
bourque, line 77: Could this be changed to use `@pytest.mark.skipif()` instead of `unittest.skipIf()`? Just so that we are not adding another dependency if we don't have to?

jwql/tests/test_bad_pixel_monitor.py
```diff
@@ -23,20 +23,18 @@
...
-
-if not ON_JENKINS:
+ON_GITHUB_ACTIONS = '/home/runner' in os.path.expanduser('~')
+
+if not ON_GITHUB_ACTIONS:
+    from jwql.database.database_interface import NIRCamBadPixelQueryHistory, NIRCamBadPixelStats
+    from jwql.database.database_interface import NIRISSBadPixelQueryHistory, NIRISSBadPixelStats
+    from jwql.database.database_interface import MIRIBadPixelQueryHistory, MIRIBadPixelStats
+    from jwql.database.database_interface import NIRSpecBadPixelQueryHistory, NIRSpecBadPixelStats
+    from jwql.database.database_interface import FGSBadPixelQueryHistory, FGSBadPixelStats
```
bourque, line 33: Is there a reason why these imports need to be under the `if not ON_GITHUB_ACTIONS` conditional?

jwql/tests/test_bad_pixel_monitor.py
```diff
@@ -23,20 +23,18 @@
...
     from jwql.instrument_monitors.common_monitors import bad_pixel_monitor
 
 
-@pytest.mark.skipif(ON_JENKINS, reason='Jenkins server currently unable to import jwst_reffiles')
+@pytest.mark.skipif(ON_GITHUB_ACTIONS, reason='Github Actions server currently unable to import jwst_reffiles')
```
bourque, line 37: Things have changed since we first wrote this, and `jwst_reffiles` is now `pip` installable and should be part of the nominal 3.7 and 3.8 environments, so perhaps these can be removed.  It would be a good thing to test out.

jwql/tests/test_pipeline_tools.py
```diff
@@ -6,7 +6,7 @@
 -------
 
     - Bryan Hilbert
-
+``
```
bourque, line 9: I think this may have been added accidentally?

bourque (changes requested):
Thanks @mfixstsci for this initial implementation!  Looks like this was pretty tricky, so thanks for this effort.  I just have a few questions/suggestions, let me know what you think.

jwql/tests/test_bad_pixel_monitor.py
mfixstsci: Hey @bourque! We have to add them under the conditional because they require the VPN to import and access that information. I believe we avoided this prior with Jenkins because it has access on the ST side of the firewall.

jwql/tests/test_bad_pixel_monitor.py
mfixstsci: Hey @bourque  sounds good. Do you mean, remove the tests that mimic JWST ref files and just import `jwst_reffiles`?

mfixstsci: Hey @bourque, I noticed that even though you can use `jwst_reffiles` in 3.7 and 3.8, these tests still call monitors which we can't currently import. So, I just went ahead and changed the reason message.

bourque (approved):
Thanks for implementing the initial solution on this @mfixstsci!  I think between your changes and mine, this is now good to go!

RIP Jenkins ⚰️

mfixstsci:
> Thanks for implementing the initial solution on this @mfixstsci! I think between your changes and mine, this is now good to go!
> 
> RIP Jenkins ⚰️

We love to see it :sunglasses:

PR #700, opened 2021-02-24
Cosmic ray monitor (via mengesser:develop)

bourque:
Creating a new PR for the cosmic ray monitor, on a new branch called `mengesser_develop`.  This will help avoid issues with `develop` branches as was seen in #636


jwql/database/database_interface.py
```diff
@@ -428,6 +428,10 @@ class : obj
 MIRIReadnoiseStats = monitor_orm_factory('miri_readnoise_stats')
 FGSReadnoiseQueryHistory = monitor_orm_factory('fgs_readnoise_query_history')
 FGSReadnoiseStats = monitor_orm_factory('fgs_readnoise_stats')
+NIRCamCosmicRayQueryHistory = monitor_orm_factory('nircam_cosmic_ray_query_history')
+NIRCamCosmicRayStats = monitor_orm_factory('nircam_cosmic_ray_stats')
```
bourque, line 432: I added files and tables for NIRCam here, as the code appears to support NIRCam too.  Hopefully that is ok!

jwql/instrument_monitors/miri_monitors/cosmic_ray_monitor.py
```diff
@@ -0,0 +1,557 @@
...
+        jump_locations_pre : list
+            List of matching coordinates one group before
+            ``jump_locations``.
+
+        rateints : ndarray obj
+            Array in ``DN/s``.
+
+        jump_data :
+
+        jump_head :
```
bourque, line 134: These parameters were missing in the docstring, and I can't quite deduce what they are for based on the code itself.  Would you mind filling them in?

jwql/tests/test_cosmic_ray_monitor.py
```diff
@@ -0,0 +1,175 @@
+#! /usr/bin/env python
```
bourque, line 1: I made some tweaks to get these tests working, and now most of them pass, however a few do not:

```
============================================================= short test summary info ==============================================================
FAILED jwql/tests/test_cosmic_ray_monitor.py::test_magnitude - assert -0.15904 == -2.77504
FAILED jwql/tests/test_cosmic_ray_monitor.py::test_get_cr_mags - assert [-0.15904] == [-2.77504]
```

Do you know what is going on here?

bourque (commented):
@mengesser Nice work on the updates here!  This is looking really good.  I just have a few more questions and there are some minor tweaks that still need to be made -- nothing major, so hopefully it won't take much effort.

Really nice work on the API docs and tests too!

I will spend some time early next week testing the code out on the test server and report back with any issues.

jwql/instrument_monitors/miri_monitors/cosmic_ray_monitor.py
```diff
@@ -0,0 +1,566 @@
...
+                jump_locations = self.get_jump_locations(jump_dq)
+                jump_locations_pre = self.group_before(jump_locations)
+
+                eff_time = jump_head['EFFEXPTM']
+
+                # Get observation time info
+                obs_start_time = jump_head['EXPSTART']
+                obs_end_time = jump_head['EXPEND']
+                start_time = julian.from_jd(obs_start_time, fmt='mjd')
+                end_time = julian.from_jd(obs_end_time, fmt='mjd')
```
bourque, line 435: @mengesser It seems like there is some logic that breaks down here.  I am seeing a lot of errors when testing things out in this part of the code.  Namely, the `pipeline_tools.calwebb_detector1_save_jump` step does not seem to complete sucessfully, which means that no `ramp_fit` or `jump` files are produced, and so the later parts of the code cannot be completed (they crash since variables like 'jump_head' are referenced before assignment).

Could you re-work this part of the algorithm to ensure that parts of the code either crash gracefully or get skipped appropriately?

bourque:
Things are looking good, and testing this module out on the dev server is now completing succesffully.  There are still a few lingering questions/concerns:

- [ ] Currently the line `cosmic_ray_rate = get_cr_rate(jump_locs, eff_time)` is commented out.  Should this be un-commented? or can it be removed?
- [ ] The `<instrument>_cosmic_ray_stats` database tables appear to not be populated.  Is this expected?
- [ ] Since this module supports both MIRI and NIRCam, it should be moved to the `common_monitors` directory
- [ ] We should make sure the API docs are added to the `sphinx` docs in this PR.
- [ ] I need to test the repeatability of this module (i.e. does it complete successfully when it is run two times in a row? when there is no new data?)

mengesser:
Just an update: We plan to start seeing MIRI data in MAST in the middle of April, so I will begin testing this again once that data is available.

jwql/instrument_monitors/common_monitors/cosmic_ray_monitor.py
```diff
@@ -0,0 +1,608 @@
...
+        dq: ndarray
+            Data Quality array containing jump flags.
+
+        Returns:
+        -------
+        jump_locs: list
+            List of coordinates to a pixel marked with a jump.
+        """
+
+        temp = np.where(dq == 4)
```
bhilbert4, line 213: Are you explicitly looking for groups flagged as containing a jump and nothing else? With this line as-is, if you have a pixel that contains any other flag and then has a jump flag added on top, it won't be found.

If you do want to catch jumps in pixels that contain other flags, I usually use this:
from jwst.datamodels import dqflags
jump = np.where(dq & dqflags.pixel["JUMP_DET"] > 0)

bhilbert4:
I'm a little concerned about the dependence on jwedb. It looks like a package to interact with the EDB, but its last release was in 2019.

jwql/instrument_monitors/common_monitors/cosmic_ray_monitor.py
```diff
@@ -0,0 +1,608 @@
...
+                    logging.info('Working on aperture {} in {}'.format(aperture, instrument))
+
+                    self.aperture = aperture
+
+                    # We start by querying MAST for new data
+                    self.query_start = self.most_recent_search()
+
+                    logging.info('\tMost recent query: {}'.format(self.query_start))
+
+                    new_entries = self.query_mast()
```
bhilbert4: I think this search needs to be refined. I'm running this on the test server and it seems to be looking for all kinds of level 2 and level 3 products. This doesn't make sense when it is then going to search these for CRs, which have been removed by the end of stage 1. I think the best bet would be to limit the search to uncal.fits files, and then run the pipeline up through the jump step.

jwql/website/apps/jwql/monitor_pages/monitor_cosmic_rays_bokeh.py
```diff
@@ -0,0 +1,165 @@
...
+    ::
+
+        from jwql.website.apps.jwql import monitor_pages
+        monitor_template = monitor_pages.CosmicRayMonitor()
+"""
+
+import datetime
+import os
+
+from astropy.io import fits
```
bhilbert4: Looks like the fits import can be removed. I don't think it's used in here.

jwql/instrument_monitors/common_monitors/cosmic_ray_monitor.py
```diff
@@ -0,0 +1,608 @@
...
+                                           'obs_end_time': end_time,
+                                           'jump_count': cosmic_ray_num,
+                                           'magnitude': cosmic_ray_mags
+                                           }
+                    self.stats_table.__table__.insert().execute(cosmic_ray_db_entry)
+
+                    logging.info("Successfully inserted into database. \n")
+                except:
+                    logging.info("Could not insert entry into database. \n")
+
```
bhilbert4, line 614: It would be helpful here to put a few lines of code in order to delete the data files that were used (uncal, jump, rate, rateits, etc). This will help save disk space on the server. As commissioning has gone on, we've seen the leftover products from the other monitors can quickly fill up our available space.

bhilbert4 (changes requested):
A few comments from an initial round of running on the test server using real data. It isn't done yet, so I haven't had a chance to look at the results. It looks like the run may have died due to lack of memory.

jwql/instrument_monitors/common_monitors/cosmic_ray_monitor.py
bhilbert4: Filtering using something like this might work:
https://github.com/spacetelescope/jwql/issues/935
Filter down to just unique file base names, then look for only the uncal and rate files.

bhilbert4:
We may also want to make the jump threshold in the call to calwebb_detector1 a user-settable parameter. At the moment it is hard-coded to 15. It might be better to have it default to 15 but be able to be set by the user. Analysis of early data will inform us about what a reasonable value for this parameter is. It may even be different for each instrument.

https://github.com/spacetelescope/jwql/blob/develop/jwql/instrument_monitors/pipeline_tools.py#L357

bhilbert4:
Trying to deal with bokeh templating here and remember all the ins and outs is frustrating. Given that bokeh templating is no longer maintained and we want to move away from it anyway, I'm working on a straight bokeh implementation of the plots.

bhilbert4:
Initial test on the test server looks good!

bhilbert4:
Initial test on the test server looks good!

bhilbert4:
@mfixstsci I think this is ready for review. I ran it twice on the dev server and it completed and the plots look ok. There are definitely a few minor improvements that can be made to make the plots look even better, but it might be better to see some more real data in the plots before that happens. It's also just be nice to finally get this in and working since it's been languishing for so long.

jwql/instrument_monitors/common_monitors/cosmic_ray_monitor.py
```diff
@@ -0,0 +1,727 @@
...
+                # Identify which tables to use
+                self.identify_tables()
+
+                # Get a list of possible apertures
+                #possible_apertures = list(Siaf(instrument).apernames)
+                possible_apertures = self.possible_apers(instrument)
+
+                #XXXFOR TESTING
+                possible_apertures = ['NRCA1_FULL']
```
mfixstsci: Add other full frame detectors/apertures

jwql/instrument_monitors/common_monitors/cosmic_ray_monitor.py
```diff
@@ -0,0 +1,727 @@
...
+                    new_entries = self.pull_filenames(new_entries)
+
+                    # Filter new entries so we omit stage 3 results and keep only base names
+                    new_entries = self.filter_bases(new_entries)
+                    logging.info(f'\tAfter filtering to keep only uncal files, we are left with {len(new_entries)} files')
+
+                    for fname in new_entries:
+                        logging.info(f'{fname}')
+
+                    # for testing purposes only
```
mfixstsci: Maybe we remove these lines?

jwql/website/apps/jwql/monitor_pages/monitor_cosmic_rays_bokeh.py
```diff
@@ -0,0 +1,196 @@
...
+
+        # Create the histogram plot
+        self.histogram_figure = self.histogram_plot()
+
+    def get_histogram_data(self):
+        """Get data required to create cosmic ray histogram from the
+        database query.
+        """
+
+        #self.mags = {}   # For testing
```
mfixstsci: more testing lines to remove here.

jwql/instrument_monitors/common_monitors/cosmic_ray_monitor.py
```diff
@@ -0,0 +1,727 @@
...
+
+                    logging.info('')
+                    logging.info('Working on aperture {} in {}'.format(aperture, instrument))
+
+                    self.aperture = aperture
+
+                    # We start by querying MAST for new data
+                    self.query_start = self.most_recent_search()
+
+                    #XXXXFOT TESTING
```
mfixstsci: Use the most recent query date as start date here.

jwql/instrument_monitors/common_monitors/cosmic_ray_monitor.py
```diff
@@ -0,0 +1,727 @@
...
+through the pipeline, and for which the output is stored in a new
+directory for each observation. Each observation is then analyzed for
+jumps due to cosmic rays, of which the number and magnitude are
+recorded. This information is then inserted into the stats database
+table.
+
+Authors
+-------
+
+    - Mike Engesser
```
mfixstsci, line 19: @bhilbert4 don't forget to add yourself here!!

setup.py
```diff
@@ -25,6 +25,7 @@
     'ipython',
     'jinja2',
     'jsonschema',
+    'julian',
```
mfixstsci: Remove and test with `astropy.time`

requirements.txt
```diff
@@ -11,6 +11,8 @@ inflection==0.5.1
 ipython==8.2.0
 jinja2==3.0.3
 jsonschema==4.4.0
+julian==0.14
+jwedb>=0.0.6
```
mfixstsci: Remove dependency on `jwedb`.

jwql/instrument_monitors/common_monitors/cosmic_ray_monitor.py
```diff
@@ -0,0 +1,727 @@
...
+import datetime
+import logging
+import os
+import re
+import shutil
+
+# Third-Party Imports
+from astropy.io import fits
+from astropy.time import Time
+import julian
```
mfixstsci: Remove and test with `astropy.time`

mfixstsci (commented):
Hey @bhilbert4 and @mengesser Thanks for the code review, there, I added the comments we discussed and once those are done, I will approve. Thanks!

jwql/instrument_monitors/common_monitors/cosmic_ray_monitor.py
bhilbert4: Done

jwql/website/apps/jwql/monitor_pages/monitor_cosmic_rays_bokeh.py
bhilbert4: Done

jwql/instrument_monitors/common_monitors/cosmic_ray_monitor.py
bhilbert4: Done

setup.py
bhilbert4: Done

requirements.txt
bhilbert4: Done

jwql/instrument_monitors/common_monitors/cosmic_ray_monitor.py
bhilbert4: Done

jwql/instrument_monitors/common_monitors/cosmic_ray_monitor.py
bhilbert4: Done

jwql/instrument_monitors/common_monitors/cosmic_ray_monitor.py
bhilbert4: Done

bhilbert4:
@mfixstsci I think all comments have been addressed. Looks like a couple tests are failing though.

jwql/website/apps/jwql/data_containers.py
```diff
@@ -43,7 +43,7 @@
 
 from jwql.database import database_interface as di
 from jwql.database.database_interface import load_connection
-from jwql.edb.engineering_database import get_mnemonic, get_mnemonic_info
+from jwql.edb.engineering_database import get_mnemonic, get_mnemonic_info, mnemonic_inventory
```
mfixstsci, line 47: Cool, thanks for updating this

jwql/instrument_monitors/common_monitors/cosmic_ray_monitor.py
```diff
@@ -0,0 +1,727 @@
...
+                     'MIRIFU_CHANNEL2B'
+                     'MIRIFU_CHANNEL2C',
+                     'MIRIFU_CHANNEL3A',
+                     'MIRIFU_CHANNEL3B'
+                     'MIRIFU_CHANNEL3C',
+                     'MIRIFU_CHANNEL4A',
+                     'MIRIFU_CHANNEL4B',
+                     'MIRIFU_CHANNEL4C']
+
+        if inst.lower() == 'niriss':
```
mfixstsci, line 479: Awesome, all of the instruments are included thank you for the updates.

jwql/tests/test_cosmic_ray_monitor.py
```diff
@@ -0,0 +1,187 @@
...
+    result = cr.most_recent_search()
+
+    assert isinstance(result, float)
+
+
+@pytest.mark.skipif(ON_GITHUB_ACTIONS, reason='Requires access to central storage.')
+def test_query_mast():
+    """Test the ``query_mast`` function"""
+
+    cr = CosmicRay()
```
mfixstsci, line 232: @bhilbert4 looks like we have tests, but it only covers miri. I think parameter-izing the tests to include the other instruments shouldn't be too difficult.

jwql/instrument_monitors/common_monitors/cosmic_ray_monitor.py
mfixstsci: @bhilbert4 are we going to implement these changes?

jwql/website/apps/jwql/monitor_pages/monitor_cosmic_rays_bokeh.py
mfixstsci: @bhilbert4 are we going to implement these changes?

jwql/instrument_monitors/common_monitors/cosmic_ray_monitor.py
mfixstsci: @bhilbert4 are we going to implement these changes?

jwql/website/apps/jwql/monitor_pages/monitor_cosmic_rays_bokeh.py
bhilbert4: This has already been removed.

jwql/instrument_monitors/common_monitors/cosmic_ray_monitor.py
bhilbert4: This has already been implemented.

jwql/instrument_monitors/common_monitors/cosmic_ray_monitor.py
bhilbert4: This has been implemented, in filter_bases()

bhilbert4:
I tested this on the test server yesterday, running it twice back to back. The results look good.

Tasks left:
- [x] Remove lines used for testing (i.e. open up to all apertures. Do we open up to all instruments, or just miri and nircam?)
- [x] Tests are failing because they refer to jwedb package, which we no longer import
- [x] Should we wait until the celery PR is merged and then update the pipeline calls in here to follow that methodology?

bhilbert4:
@mfixstsci based on @york-stsci 's response in slack, I think this is ready to be merged. I'm not sure how deeply you want to review it. Bourque reviewed it quite a while ago, but I think the number of updates made since that review are pretty small.

mfixstsci:
> @mfixstsci based on @york-stsci 's response in slack, I think this is ready to be merged. I'm not sure how deeply you want to review it. Bourque reviewed it quite a while ago, but I think the number of updates made since that review are pretty small.

Sounds good to me @bhilbert4 I am updating the branch but we need to address some of the pip issues.

bhilbert4:
> > @mfixstsci based on @york-stsci 's response in slack, I think this is ready to be merged. I'm not sure how deeply you want to review it. Bourque reviewed it quite a while ago, but I think the number of updates made since that review are pretty small.
> 
> Sounds good to me @bhilbert4 I am updating the branch but we need to address some of the pip issues.

The bare `except`s bring up an interesting question. What should the monitor do if it fails to open and read in data from a file? The way it's written at the moment it would log the failure, but then on the next line it would crash because the variables that are referenced wouldn't be defined. I'm tempted to instead have the code just skip that file and move on to the next one. Does that make sense to you?

york-stsci:
I agree that, if you're looping through files, and fail to find a given file, then it makes sense to do one of:

- log the exception and re-raise it so that the monitor will crash
- log the exception and call `continue` to proceed to the next file
- log the exception and call `break` to end this particular loop

Which it makes sense to do depends on what the effect of a single file missing should be. If you should make a product (or database entry, or whatever) with what you could find, then `continue` is the most sensible. If you should fail to make this particular product, and then go back and try again the next time the monitor is run, use `break`. If failing to find a file leaves the monitor in a place where it can no longer do anything useful in this particular run, then re-raise and crash.

bhilbert4:
@mfixstsci I've taken care of most of the PEP8 issues. The few that are left are:
1) multiline import statements. I feel like indenting the second line to match the first makes things much more readable. But I can switch to the official 4 spaces if you think we should be strict.
2) A character used for regex that PEP8 doesn't realize
3) a " == True" comparison used in a database, so I think we need to keep it, right?
4) line breaks before a series of 'and' statements. The PEP8 documentation takes no official stand on this, so I'm not sure why the linter is. The way we have it seems the most readable, and in fact matches an example in the PEP8 documentation.
5) On last bare except. I don't know what specific exceptions to catch here. The try/except is around a database insert command. I checked, and other monitors don't have a try/except here. It seems like it could actually be useful to have it, but I'm not sure how to find the specific exceptions that may happen.

york-stsci:
For cases where pep8 is complaining, but there is a valid reason to do it the way that you're doing it, is it worth using the '#noqa X" comment style (which will cause pep8 to ignore that issue on that line)?

mfixstsci:
@bhilbert4 here are my thoughts:

1. I agree with the indenting, I recall in multiple files we have in the package we just repeat lines like this:
```
from jwql.module_name import a, b, c
from jwql.module_name import d, e, f
```
2. I think we can ignore this one
3. Here we will want to use `a is True` instead of `a == True` here
4. I tried to follow docs but it was giving me trouble too, I think we should ignore it.
5. Most packages like `sqlalchemy` have their own defined errors. Could be something like `DatabaseInsertionError` and normally can import it `from sqlalchemy.errors import DatabaseInsertionError`.

@york-stsci that is the way we can ignore pep issues, I think we should add it to the line with the `regex` issue.

bhilbert4:
@mfixstsci 
I don't think we want to change "== True" to "is True" in 3). That line is going into a database query. See the note above that line [here](https://github.com/spacetelescope/jwql/blob/1436932cbd92896fd5dbb08e812b7e7529e70432/jwql/instrument_monitors/common_monitors/cosmic_ray_monitor.py#L400).

I'm not sure what to do with the remaining errors:
[Line 1092:17](https://github.com/spacetelescope/jwql/blob/93bbfad74d87bcf7bb1a12f08ad6b3d8a4c0ee67/jwql/website/apps/jwql/data_containers.py#L1092): [W503](https://duckduckgo.com/?q=pep8%20W503) line break before binary operator
If I put a `noqa` on each line, it results in a syntax error, I'm assuming because the lines are continuations of one another.

mfixstsci:
@bhilbert4 thats okay, the binary operator was giving me issues too. As for number 3 I found this:
https://stackoverflow.com/questions/18998010/flake8-complains-on-boolean-comparison-in-filter-clause

Solution is to add `#noqa` to that line.

bhilbert4:
@mfixstsci I think I've finally addressed everything. I noticed #1034 while checking the log this morning, but I think that is either a pipeline issue or an issue with the JWQL code for calling the pipeline, rather than a problem with the cosmic ray monitor. So I think this is ready for re-re-re-reviewing, or merging. :)

jwql/instrument_monitors/common_monitors/cosmic_ray_monitor.py
```diff
@@ -254,13 +254,11 @@ def get_jump_data(self, jump_filename):
...
 
         """
-
         try:
-            hdu = fits.open(jump_filename)
-            head = hdu[0].header
-            data = hdu[1].data
-            dq = hdu[3].data
-            hdu.close()
+            with fits.open(jump_filename) as hdu:
```
mfixstsci, line 258: Awesome, this is a good catch.

mfixstsci (approved):

mfixstsci:
This has been a long time coming @bhilbert4 @mengesser, merging.

PR #705, opened 2021-03-01
Update pipeline rscd step call

bsunnquist:
This PR updates the call to the JWST pipeline RSCD step, which changed naming formats to be consistent with the other pipeline steps.

I didn't see any other codes directly calling this step, so only `pipeline_tools` was updated.


bourque:
@bsunnquist Thank you for the quick fix!  I will use this PR as an opportunity to update the environments with the latest `jwst` version; hopefully the tests pass!

bourque:
It looks like the tests are passing here (though the Jenkins builds fail due to lingering python 3.8 issue, but that it not relevant), so merging this now.

PR #709, opened 2021-03-18
Improve loading times for web app archive pages

### LINKED ISSUE #676: Address speed issue for loading thumbails/preview images

bourque:
This PR aims to improve the loading times to the web app archive pages.  Closes #676

### DIFF
--- a/jwql/jwql_monitors/generate_proposal_thumbnails.py
+++ b/jwql/jwql_monitors/generate_proposal_thumbnails.py
@@ -1,0 +1,60 @@
+#! /usr/bin/env python
+
+"""Generate thumbnails for each proposal to be used for display in
+the archive page of the web app.
+
+In the web app, each instrument has its own 'archive' page where users
+may view JWST images for the particular instrument.  The page displays
+buttons for each proposal with a thumbnail image from each particular
+proposal.  This script goes through the thumbnail filesystem and
+creates the thumbnail to display based on the first existing rate file.
+The thumbnail is saved as ``<proposal_id>.thumb``.
+
+Authors
+-------
+
+    - Matthew Bourque
+
+Use
+---
+
+    This script is intended to be executed as such:
+
+    ::
+
+        python generate_proposal_thumbnails.py
+"""
+
+import glob
+import logging
+import os
+import shutil
+
+from jwql.utils.logging_functions import configure_logging, log_info, log_fail
+from jwql.utils.utils import get_config
+
+
+@log_fail
+@log_info
+def generate_proposal_thumbnails():
+    """The main function of the ``generate_proposal_thumbnails`` module.
+    See module docstring for further details."""
+
+    proposal_dirs = glob.glob(os.path.join(get_config()['thumbnail_filesystem'], '*'))
+
+    for proposal_dir in proposal_dirs:
+        rate_thumbnails = glob.glob(os.path.join(proposal_dir, '*rate*.thumb'))
+        if rate_thumbnails:
+            thumbnail = rate_thumbnails[0]
+            proposal = os.path.basename(thumbnail)[0:7]
+            outfile = os.path.join(proposal_dir, '{}.thumb'.format(proposal))
+            shutil.copy2(thumbnail, outfile)
+            logging.info('Copied {} to {}'.format(thumbnail, outfile))
+
+
+if __name__ == '__main__':
+
+    module = os.path.basename(__file__).strip('.py')
+    configure_logging(module)
+
+    generate_proposal_thumbnails()
--- a/jwql/utils/utils.py
+++ b/jwql/utils/utils.py
@@ -74,9 +74,9 @@ def _validate_config(config_file_dict):
                     "password": {"type": "string"},
                     "host": {"type": "string"},
                     "port": {"type": "string"}
-                    },
-                    "required": ['engine', 'name', 'user', 'password', 'host', 'port']
-                 },
+                },
+                "required": ['engine', 'name', 'user', 'password', 'host', 'port']
+            },
             "filesystem": {"type": "string"},
             "preview_image_filesystem": {"type": "string"},
             "thumbnail_filesystem": {"type": "string"},
@@ -105,7 +105,7 @@ def _validate_config(config_file_dict):
         jsonschema.validate(instance=config_file_dict, schema=schema)
     except jsonschema.ValidationError as e:
         raise jsonschema.ValidationError(
-            'Provided config.json does not match the ' + \
+            'Provided config.json does not match the '
             'required JSON schema: {}'.format(e.message)
         )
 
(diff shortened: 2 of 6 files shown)


bourque:
@tnking97 I was wondering if you would be interested and able to review this PR sometime next week (or this week, depending on when you read this)?  You are pretty familiar with how the web app pages dealing with thumbnails work, so it would be really helpful if you could look it over.  Let me know if you would have time to do that.  Thanks!

tnking97:
Hi @bourque, Sure I can review this PR early next week!

bourque:
A few things to note about this PR:

- Putting the thumbnails in the internal storage of the dev/test/ops server already helps to speed up the web page loading times (see #694), but even with that the loading times were quite slow, especially for hefty instruments such as NIRCam.
- I determined that the major bottleneck for rendering the `archive_proposals` view was due to having to locate suitable thumbnails in the thumbnail filesystem (with `glob.glob`).  This seems unnecessary since we only really need one thumbnail to represent a given proposal, so in theory they should be easy to locate.  So to get around this, I created a new `jwql_monitor` monitor (called `generate_proposal_thumbnails.py`) that goes through the `thumbnail_filesystem`, chooses one thumbnail to represent the proposal, and copies the thumbnail to a `<proposal_id>.thumb` file.  Once that exists, the web app can now (essentially) hard-code a path to it, so no searching is necessary
- Another major bottleneck was because we were scraping the filesystem to discover files display in the web app, instead of using `astroquery.Mast` (database queries tend to be much faster than file I/O and/or pinging a network).  This is understandable, as early in the development of JWQL, `astroquery.Mast` was still being developed and was not ready to be used yet.  But I switched to using `astroquery.Mast` in this PR and it seems to be working well (and faster!)
- Another bottleneck was in the `filesytem_path` function in `utils.py`, which was checking if the file exists in the filesystem, which takes I/O & network effort.  While this is a nice check to have, it was slowing things down too much, so I created a `check_existence` parameter to optionally perform this.  For the purposes of the web app, we now skip this step.
- Another bottleneck was pinging the `utils.get_config()` function many times within various functions.  To get around this, I moved the `get_config()` file towards the beginning of the `utils` module, so that I could create a global `FILESYSTEM` variable that needs only to be defined/imported once.

bourque:
Awesome, thanks so much @tnking97!

jwql/jwql_monitors/generate_proposal_thumbnails.py
```diff
@@ -0,0 +1,60 @@
...
+def generate_proposal_thumbnails():
+    """The main function of the ``generate_proposal_thumbnails`` module.
+    See module docstring for further details."""
+
+    proposal_dirs = glob.glob(os.path.join(get_config()['thumbnail_filesystem'], '*'))
+
+    for proposal_dir in proposal_dirs:
+        rate_thumbnails = glob.glob(os.path.join(proposal_dir, '*rate*.thumb'))
+        if rate_thumbnails:
+            thumbnail = rate_thumbnails[0]
```
tnking97, line 48: Is there any case where we would expect the first thumbnail to not be representative of the set of thumbnails? This seems like a very good way to be consistent, though, and I'm not sure that it matters if it's not a representative thumbnail.

jwql/utils/utils.py
```diff
@@ -43,6 +43,109 @@
...
+                "properties": {
+                    "engine": {"type": "string"},
+                    "name": {"type": "string"},
+                    "user": {"type": "string"},
+                    "password": {"type": "string"},
+                    "host": {"type": "string"},
+                    "port": {"type": "string"}
+                    },
+                    "required": ['engine', 'name', 'user', 'password', 'host', 'port']
+                 },
```
tnking97, line 79: It looks like there are a few little things that pep8speaks already mentioned, such as the over-indent of this bracket, but I didn't find any additional formatting issues.

jwql/utils/utils.py
```diff
@@ -43,6 +43,109 @@
...
+            raise ValueError('Incorrectly formatted config.json file. '
+                             'Please fix JSON formatting: {}'.format(e))
+
+    # Ensure the file has all the needed entries with expected data types
+    _validate_config(settings)
+
+    return settings
+
+if not ON_GITHUB_ACTIONS:
+    FILESYSTEM = get_config()['filesystem']
```
tnking97, line 147: Nice way to make this run more smoothly! Minor note- I think pep8 wants two spaces before the if statement.

jwql/website/apps/jwql/data_containers.py
```diff
@@ -1161,23 +1158,25 @@ def thumbnails_ajax(inst, proposal=None):
...
-    detectors = [data_dict['file_data'][rootname]['filename_dict']['detector'] for
-                 rootname in list(data_dict['file_data'].keys())]
-    proposals = [data_dict['file_data'][rootname]['filename_dict']['program_id'] for
-                 rootname in list(data_dict['file_data'].keys())]
+    # (Don't include the proposal as a sorting parameter if the proposal has already been specified)
+    detectors, proposals = [], []
+    for rootname in list(data_dict['file_data'].keys()):
+        proposals.append(data_dict['file_data'][rootname]['filename_dict']['program_id'])
+        try:  # Some rootnames cannot parse out detectors
+            detectors.append(data_dict['file_data'][rootname]['filename_dict']['detector'])
```
tnking97, line 1176: Great, this is a good catch and should still be easy to add additional thumbnail sorting options.

tnking97 (commented):
I was excited to see these changes with lots of clever adjustments. Everything is running smoothly locally in the tests that I did. I added just a few minor comments. Did you use a profiler to figure out where the bottlenecks were?

jwql/jwql_monitors/generate_proposal_thumbnails.py
bourque: Great question.  I did not re-work that logic as part of this PR, so using the first `rate` file as a thumbnail is how we currently do it, but perhaps this could be improved.  My hunch is that (at least for now), this is not a big deal, since the thumbnails are really just for aesthetics and are not intended to be used for any sort of meaningful way.

bourque:
@tnking97 Thank you for the review!  I pushed some edits to address your comments, and this looks like it is good to go!

jwql/jwql_monitors/generate_proposal_thumbnails.py
tnking97: Okay, that seems very reasonable!

tnking97 (approved):

PR #712, opened 2021-03-24
Adding Bandit Security Check to Github Actions

mfixstsci:


mfixstsci:
The `bandit` check from https://github.com/marketplace/actions/python-security-check-using-bandit does not work with ubuntu. Trying a different action https://github.com/marketplace/actions/bandit-check

mfixstsci:
I went ahead and added `continue-on-error: true` to the Bandit step of the build to avoid build failures. There are a lot of potential risks raised by Bandit and I have investigated a few and found that although there are risks, we are using the offending methods or libraries responsibly. The suggested libraries or methods suggested by bandit don't allow the same functionality so there will need to be some further investigation.

bourque (approved):
Thanks @mfixstsci!  This is a nice initial implementation!

Issue #532, opened 2020-01-28
Build "all" option for viewing headers in web app


bourque:
Currently users may view a header for a specific extension of a file.  It would be nice to have a "view all" option in case users want to ctrl+f for something, or otherwise view all extension headers for other reasons.

PR #721, opened 2021-04-07
WIP Display All Header Content in Single Table.

mfixstsci:
Addresses #532


mfixstsci:
@bourque if you know anyone with some more JS experience could take a look at this. It was a little buggy before my changes. The tables don't display properly when the headers are loaded.

Issue #634, opened 2020-09-23
Improved config file handling


gkanarek:
Currently, whenever I pull from the repo, my config file gets overwritten. (Also, there's no conceptual reason why `config.json` should be in `jwql.utils`.)

Here are my suggestions:
- As part of `setup.py`, create the `~\.jwql`  directory and create a default `config.json` file there. This will be the new location of the config file, so any code that looks for the config file needs to point there.
- Include a `jwql-version` keyword in the config file. Whenever the config file is accessed, check that version against the runtime JWQL version; if there's a mismatch, raise a warning, and also make sure to add default values for any missing keywords (unless the omission breaks the run, in which case it should throw an error as normal).
- It might be useful to have a flag for "development" vs "production" or something similar, which would point to different directories for output files, logs, etc., so we wouldn't need to edit the paths in the config file when we're developing code that might dump a bunch of junk in the logs directory, or need dummy output that we have locally, or whatever.

Issue #711, opened 2021-03-22
Ensure there is full API docs coverage prior to 1.0 release


bourque:
We should make sure we do not have any missing API docs before we make our 1.0 release.

bourque:
With #745 completed, I think this should be mostly done.  We just need to make sure there are not any modules that exist in the `jwql` package that are not currently listed in the `docs/source/*.rst` files.  @bhilbert4 Could you do a pass and make sure nothing is missing?

PR #724, opened 2021-04-13
Config file handling

tnking97:
Improve config file handling as described in #634


bourque:
@tnking97 FYI, as I was doing some work with #726, I realized that we have `utils._validate_config` and `test_utils.test_validate_config` functions that make sure the `config.json` file is structured in an expected way.  This PR would be a good opportunity to make sure those functions are up-to-date with the current config file.  I think there may have been some keys added since those functions were originally written.

bhilbert4:
Here's a possibly-related issue with the config file relating to #711. When I looked into the log of the [latest documentation build on readthedocs](https://readthedocs.org/api/v2/build/13747438.txt), I noticed there are a bunch of FileNotFoundError, where the config file isn't being found by readthedocs when building the environment. This is preventing some JWQL modules from being imported, and it looks like these are the modules for which the API documentation pages are not being built.

example_config.json
```diff
@@ -0,0 +1,28 @@
+{
```
bourque, line 1: Great idea to name this something different than `config.json` so that we avoid committing personal config files!

bourque (approved):
Thank you @tnking97, this looks great!  And thank you for updating the wiki page too!

Issue #727, opened 2021-04-29
Update pysiaf


bourque:
When importing parts the `jwql` package, this message comes up:

```
Please consider updating pysiaf, e.g. pip install --upgrade pysiaf or conda update pysiaf
```

PR #726, opened 2021-04-27
Switch filesystem to MAST isilon storage

bourque:
This PR makes necessary changes to get JWQL to properly handle data stored in MAST isilon storage.

PR #732, opened 2021-05-10
Update pysiaf

mfixstsci:
Updating `pysiaf` versioning for conda and pip requirements. This PR addresses #727


bourque (approved):
Thank you for this @mfixstsci!

Since our filesystem is in a state of disarray right now (see #726), I can't really run all of the unit tests on this PR.  Let's just assume this will not break anything (what could go wrong???), and if it does, we can address it in a future PR.

PR #740, opened 2021-06-08
Patch for ReadTheDocs to find empty example config file

tnking97:
ReadTheDocs is currently failing because it is looking for the `config.json` file which does not exist. An `example_config.json` file is used as an empty placeholder and description of what contents are needed in `config.json`. However, this file is updated by each user as it contains private information. Using the `example_config.json` naming convention, in addition to placing `config.json` in `.gitignore`, also prevents users from accidentally committing the complete configuration file to the JWQL repository. This patch uses the `READTHEDOCS` environment variable to determine whether JWQL should look for `example_config.json` or `config.json`.

### DIFF
--- a/jwql/utils/utils.py
+++ b/jwql/utils/utils.py
@@ -123,12 +123,17 @@ def get_config():
     settings : dict
         A dictionary that holds the contents of the config file.
     """
-    config_file_location = os.path.join(__location__, 'config.json')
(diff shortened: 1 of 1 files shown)


bhilbert4 (approved):
Looks good to me!

bourque (approved):
Thanks @tnking97, this is very clever!  Looks good!

PR #745, opened 2021-06-11
Fix RTD build failures

bhilbert4:
This PR is being used to investigate readthedocs build failures related to the config file.


bhilbert4:
Woo hoo! @bourque @tnking97 this looks like it's working! The hidden failures in the RTD build have all been eliminated, RTD now finds the example config file, and as a result, all the API docs are now being created. 

Note that this moves the location of the config file from where it was placed in #724. In this case, it is moved into the top-level directory of the package (as opposed to the top level directory of the repository), so that it is included in the build on RTD. I changed the assumed location for both config.json and example_config.json in order to be consistent, but I don't see any real reason why they can't be in different places if we decide that's best (it's a one-line change). 

Anyway, I think this is ready for review.

tnking97:
This looks great to me! I just have one question: in jwql/website/apps/jwql/data_containers.py and jwql/tests/test_preview_image.py.get_test_fits_files(), would we expect to run into cases in which ON_GITHUB_ACTIONS is false and ON_READTHEDOCS is false, for instance, that might require other if/then options?

bhilbert4:
The lines within the if clauses are only performed when not on github actions or readthedocs, and all the lines are basically get_config() calls. When we are on github actions or readthedocs I don't think we need any of the information contained in the config file, so I think it's ok to not have an else statement tacked on. I think all the passing tests and successful RTD build confirms that it's ok.

bourque (approved):
Thank you for your hard work on this @bhilbert4!  The code changes look great, and it is really nice having this RTD check built into the other GitHub checks.  It is nice to see all of the API docs finally available!

Also thank you @tnking97 for also taking a look at this!

bourque:
BTW, I used this PR as an excuse to update the `MANIFEST.in` file

Issue #763, opened 2021-07-19
Dark Monitor crashes on data from latest DMS build


bourque:
**Describe the bug**
The `dark_monitor` crashes on NIRCam data from the latest DMS build (#748)

**To Reproduce**
Run the `dark_monitor.py` script on the dev server which points to the latest DMS build

Here is some of the log file that contains the traceback (some info has been clipped for redundancy and not exposing directory structures)

```
07/19/2021 11:50:56 AM INFO: 	Working on file: outputs/dev/dark_monitor/data/nircam_nrca1_full/jw00759141001_03101_00001_nrca1_dark.fits
07/19/2021 11:50:56 AM INFO: 	Pipeline steps that remain to be run:
07/19/2021 11:50:56 AM INFO: 		group_scale: False
07/19/2021 11:50:56 AM INFO: 		dq_init: False
07/19/2021 11:50:56 AM INFO: 		saturation: False
07/19/2021 11:50:56 AM INFO: 		ipc: False
07/19/2021 11:50:56 AM INFO: 		refpix: False
07/19/2021 11:50:56 AM INFO: 		superbias: False
07/19/2021 11:50:56 AM INFO: 		persistence: False
07/19/2021 11:50:56 AM INFO: 		dark_current: False
07/19/2021 11:50:56 AM INFO: 		linearity: False
07/19/2021 11:50:56 AM INFO: 		firstframe: False
07/19/2021 11:50:56 AM INFO: 		lastframe: False
07/19/2021 11:50:56 AM INFO: 		rscd: False
07/19/2021 11:50:56 AM INFO: 		jump: True
07/19/2021 11:50:56 AM INFO: 		rate: True
07/19/2021 11:50:56 AM INFO: 	Slope file outputs/dev/dark_monitor/data/nircam_nrca1_full/jw00759141001_03101_00001_nrca1_dark_rate.fits already exists. Skipping call to pipeline.
07/19/2021 11:50:56 AM INFO: 	Slope images to use in the dark monitor for nircam, NRCA1_FULL:
07/19/2021 11:50:56 AM INFO: 		outputs/dev/dark_monitor/data/nircam_nrca1_full/jw00759101001_02101_00001_nrca1_dark_rate.fits
07/19/2021 11:50:56 AM INFO: 		outputs/dev/dark_monitor/data/nircam_nrca1_full/jw00759103001_02101_00001_nrca1_dark_rate.fits
[...]

Issue #761, opened 2021-07-19
Bad Pixel Monitor crashes on data from latest DMS build


bourque:
**Describe the bug**
The `bad_pixel_monitor` crashes on MIRI data from the latest DMS build (#748)

**To Reproduce**
Run the `bad_pixel_monitor.py` script on the dev server which points to the latest DMS build

Here is some of the log file that contains the traceback (some info has been clipped for redundancy and not exposing directory structures)

```
07/19/2021 11:35:45 AM INFO: Working on aperture ('MIRIMAGE', 'MIRIM_FULL') in miri
07/19/2021 11:35:45 AM INFO: 	No query history for MIRIMAGE. Beginning search date will be set to 57357.0.
07/19/2021 11:35:45 AM INFO: 	No query history for MIRIMAGE. Beginning search date will be set to 57357.0.
07/19/2021 11:35:45 AM INFO: 	Flat field query times: 57357.0 59414.649822773325
07/19/2021 11:35:45 AM INFO: 	Dark current query times: 57357.0 59414.649822773325
07/19/2021 11:35:45 AM WARNING: 		Unable to locate jw99002001002_03102_00001_mirimage_uncal.fits in filesystem. Not including in processing.
07/19/2021 11:35:45 AM WARNING: 		Unable to locate jw99002001002_03102_00001_mirimage_uncal.fits in filesystem. Not including in processing.
07/19/2021 11:35:45 AM INFO: 	Bad pixels from flats skipped. 3 new flat files for miri,('MIRIMAGE', 'MIRIM_FULL') found. 10 new files are required to run bad pixelsfrom flats portion of monitor.
07/19/2021 11:35:45 AM WARNING: 		Unable to locate jw00619044001_02101_00001_mirimage_uncal.fits in filesystem. Not including in processing.
07/19/2021 11:35:45 AM WARNING: 		Unable to locate jw00619051001_02101_00001_mirimage_uncal.fits in filesystem. Not including in processing.
07/19/2021 11:35:45 AM WARNING: 		Unable to locate jw00619058001_02101_00001_mirimage_uncal.fits in filesystem. Not including in processing.
07/19/2021 11:35:45 AM WARNING: 		Unable to locate jw00619079001_02101_00001_mirimage_uncal.fits in filesystem. Not including in processing.
07/19/2021 11:35:45 AM WARNING: 		Unable to locate jw00619086001_02101_00001_mirimage_uncal.fits in filesystem. Not including in processing.
[...]

Issue #747, opened 2021-06-16
Add test for monitor webpages


bsunnquist:
It would be nice to test that the various instrument monitor pages load as part of our unit tests. This could probably be accomplished by modifying `test_loading_times.py`.

PR #765, opened 2021-07-23
Add tests for monitor webpages

bsunnquist:
This PR adds all available monitor URLs to `test_loading_times.py`, to make sure each monitor webpage is able to load as part of our unit tests.


bsunnquist:
@bourque this is ready for review whenever. I thought it would be a good idea to add these tests (https://github.com/spacetelescope/jwql/issues/747), especially if we ever start modifying bokeh versions, but let me know if not.

bourque (approved):
Thank you @bsunnquist, the code changes here look great.

It looks like `test_monitor_mast` is failing for some reason not related to your changes.  Perhaps the MAST API has changed in some capacity.  I will investigate this in a separate issue.

As far as this PR is concerned, I think it is good to go.  There are a few of these tests are failing, but their traceback makes me think they are related to #761 and #763.

Issue #746, opened 2021-06-15
Remove need for Astroconda channel


bourque:
The support of the `astroconda` `conda` channel at the institute is being deprecated in the near future.  For packages that are distributed in this channel, they will become (or already are) available via PyPI. 

For JWQL, we need to:
1. Identify any packages in the JWQL list of dependencies that are currently being installed via the `astroconda` channel
2. Modify the JWQL `conda` environment `yaml` files to make sure said packages are being installed via `pip`.

PR #767, opened 2021-07-28
Remove need for astroconda channel

tnking97:
Remove need for astroconda channel by pip installing packages that were previously installed with astroconda.


bourque:
Closes #746

bourque (approved):
Thanks @tnking97, this looks great!

Issue #714, opened 2021-03-26
Ensure newly-created database tables are readable by JWQL database accounts


bourque:
The JWQLDB database is managed by two accounts, one is read-only, one is read & write.  After using `reset_database.py`, it appears that newly-created tables are not readable by the read-only account, as performing a `SELECT` query on a given table results in a permissions error.

We should figure out if there is any way to grant read permissions to tables through our current `database_interface.py`/`reset_database.py` `sqlalchemy` solution.

mfixstsci:
@bourque I can take this one, I don't have any `in progress` items for the `v1.0.0` project board.

bourque:
Thanks @mfixstsci!

PR #785, opened 2021-08-18
Add Read Permissions

mfixstsci:
Using `sqlalchemy` to grant read permissions to `public` tables in database for `jwqldb_user_read`.


mfixstsci:
Still need to test this!

mfixstsci:
This one is ready for review! 🎉

jwql/database/database_interface.py
```diff
@@ -388,6 +389,22 @@ class : obj
...
+    Parameters
+    ----------
+    None
+
+    Returns
+    -------
+    None
+    """
+    with engine.connect() as conn:
+        raw_sql = text("""GRANT SELECT ON ALL TABLES IN SCHEMA public TO jwqldb_user_read;""")
```
bourque, line 404: I would suggest to read in the database username from the `config.json` file, as to not expose usernames in our code base.  The problem though is that the `database`.`user` key may point to either the read or the write account, so one would have to do some finagling.  Perhaps something like this would work:

```python
db_username = get_config()['database']['user']
db_username = '_'.join(db_username.split('_')[:-1])
db_account = '{}_read'.format(db_username)
raw_sql = text(GRANT SELECT ON ALL TABLES IN SCHEMA public TO {};""".format(db_account)
```

bourque (changes requested):
@mfixstsci Thank you for implementing this!  It will be one less thing to worry about when dealing with our database.

I just have a few small suggestions

jwql/database/database_interface.py
mfixstsci: @bourque I tried this and it works, but then breaks. You never want to use `.format()` when inserting text into a sql query. So I went ahead and tried to do it "the right way"


```
def set_read_permissions(username):
    """Set read permissions for db tables"""

    with engine.connect() as conn:
        raw_sql = text("GRANT SELECT ON ALL TABLES IN SCHEMA public TO :user")
        conn.execute(raw_sql, user=username)
```

But `username` is passed as a string and the database isn't a very big fan of:

`GRANT SELECT ON ALL TABLES IN SCHEMA public TO 'jwqldb_user_read'`

and is fine with:

`GRANT SELECT ON ALL TABLES IN SCHEMA public TO jwqldb_user_read`

That extra set of `' '` seems to be whats breaking it.

bourque: Ahh, I see.  I think you had explained this to me earlier too, sorry I misunderstood.  I'll try to think if a way around this.

mfixstsci: Yeah, I find it strange that the command you sent me from the ticket also contained the `' '` around the username. I am unsure if it is a `postgres` or `sqlalchemy` thing.

bourque:
@mfixstsci After some digging, I found a solution that appears to work, which uses `engine.execute()` directly:

```python
def set_read_permissions():
    """Set read permissions for db tables"""

    db_username = get_config()['database']['user']
    db_username = '_'.join(db_username.split('_')[:-1])
    db_account = '{}_read'.format(db_username)
    command = 'GRANT SELECT ON ALL TABLES IN SCHEMA public TO {};'.format(db_account)
    engine.execute(command)
```

With that, I think this PR is good to go!

bourque (approved):

bourque:
closes #714

mfixstsci:
@bourque thank you!

Issue #788, opened 2021-08-26
Describe the process for submitting a bug report and/or feature request in About page


bourque:
Before we release the JWQL app to the users, we should add a section in the About page that describes how to provide the team with feedback

PR #798, opened 2021-09-22
Add info on bug/feature requests to about page

bsunnquist:
This PR makes changes to the about page to describe the bug/feature request instructions for users.

### DIFF
--- a/jwql/website/apps/jwql/templates/about.html
+++ b/jwql/website/apps/jwql/templates/about.html
@@ -60,7 +60,7 @@
 
         <br>
         <h4>Contact</h4>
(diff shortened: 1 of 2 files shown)


bsunnquist:
@bourque we had some info in our contact section already for this, so I just added a little more details to it instead of making a new section. But let me know if you were wanting more detailed instructions, e.g. like this: https://docs.github.com/en/issues/tracking-your-work-with-issues/creating-an-issue#creating-an-issue-from-a-repository

bourque:
Closes #788

bourque:
@bsunnquist These changes look great!

But now that I think about this some more, it would probably also be good to have something on our homepage, as users may not think to go over to the About page for this.  Could you add a little button towards the bottom of the homepage that directs users to https://github.com/spacetelescope/jwql/issues/new/choose?  Maybe something like:

```
Found a bug or have a feature request?

<Submit an issue>
```

bsunnquist:
Yeah that's a good idea - I just submitted that change

bourque (approved):
@bsunnquist Awesome, this looks perfect!  Thanks again!

Issue #800, opened 2021-09-23
Exclude ASIC tuning data from MAST search results


bhilbert4:
Currently the dark current monitor (and possibly other monitors) query MAST for dark current exposures and perform analyses on these exposures. One subset of data that are included in these queries are ASIC tuning data. These files are labeled with an EXP_TYPE of e.g. NRC_DARK, so they are indistinguishable from normal dark current exposures. However, the readout details can change during ASIC tuning, which can result in a change to the measured dark current signal. Therefore, ASIC tuning data should be ignored in terms of instrument monitor analyses. 

After some investigation, it turns out that there is a TEMPLATE keyword in the primary header of the science files. For ASIC tuning data, this keyword is set to be  'ISIM ASIC Tuning'. Knowing this, we can now adapt the instrument monitors to filter out the ASIC data by examining this keyword in the list of files returned by MAST queries. In the jwst datamodels, the TEMPLATE keyword corresponds to the model.meta.observation.template entry in the schema. I'm not sure how/if it is stored in the metadata returned by a MAST query.

PR #804, opened 2021-09-24
[WIP]: Add function to filter out ASIC tuning data from mast results

bhilbert4:
This PR adds code to the instrument monitors such that any ASIC tuning data returned by the MAST query will be ignored. ASIC tuning data will have suspect pixel values since readout properties will be non-nominal for these data. We don't want these pixel values contaminating measures of dark current, readnoise, etc.

Files are excluded by checking the 'template' keyword in the dictionaries returned by the MAST query.

- [ ] Find out if there are instrument-specific ASIC tuning template names. Currently the code only looks for the value that NIRCam uses.
- [ ] More testing required. Will need to reset the database tables in order to fully test.
- [ ] Add the filtering function to the readnoise, bias, and bad pixel monitors (cosmic ray monitor?)

### DIFF
--- a/jwql/instrument_monitors/common_monitors/dark_monitor.py
+++ b/jwql/instrument_monitors/common_monitors/dark_monitor.py
@@ -76,13 +76,36 @@ from jwql.database.database_interface import FGSDarkQueryHistory, FGSDarkPixelSt
 from jwql.instrument_monitors import pipeline_tools
 from jwql.jwql_monitors import monitor_mast
 from jwql.utils import calculations, instrument_properties
-from jwql.utils.constants import JWST_INSTRUMENT_NAMES, JWST_INSTRUMENT_NAMES_MIXEDCASE, JWST_DATAPRODUCTS, RAPID_READPATTERNS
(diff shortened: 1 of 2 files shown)


bourque:
Closes #800

bhilbert4:
According to the JWST keyword dictionary, there is only one TEMPLATE value that is related to ASIC tuning, so no need to look for instrument specific template values. Of course, if the instrument teams had other templates they wanted to ignore, those templates could be added to the new list in constants.py and they would be skipped.

bhilbert4:
Testing on the dev server looks good. The dark monitor failed, but due to NIRSpec darks with either bad metadata or in need of better filtering based on array size. This was all after successfully filtering out ASIC tuning data for multiple other apertures. Tests in the bad pixel, bias, and readnoise monitors were also successful.

bhilbert4:
@bourque I think this is ready for review. I am a little confused about the RTD failure though. From the raw RTD log, it looks like the test is installing Django version 4.0 (which, when I look around, seems not to exist?) See the Django references here: https://readthedocs.org/api/v2/build/14836853.txt

bourque:
Thanks @bhilbert4!  I think the RTD failures were fixed in a recent PR; updating the branch appears to have fixed it.

bourque:
Thanks @bhilbert4 for the updates here!  I'm going to merge this now.

PR #821, opened 2021-10-07
[WIP]: Display correct number of files in thumbnails on archive page

### LINKED ISSUE #801: Archived Images page showing the wrong number of files for each program
Working on the dev server:
On the Archived Images page for a given instrument, there is a square for each program ID. Along with the ID it says how many files are in that program. Rarely does that number seem to match the number of files shown when you click on the program and open the page for it though. I'm not sure if it's getting that count by including all versions of each file (i.e. uncal, rate, etc), or if this is just a symptom of this test filesystem and the numbers are correct.

bhilbert4:
Resolves #801 

This is the start of an attempt to get the correct number of thumbnails/activities to be displayed in each of the proposal ID thumbnail images when on an instrument's Archived Images page. Currently the counts are too large. I think this is because it is counting each filetype (uncal, rate, etc) for each exposure.

### DIFF
--- a/jwql/website/apps/jwql/data_containers.py
+++ b/jwql/website/apps/jwql/data_containers.py
@@ -853,6 +853,31 @@ def get_proposal_info(filepaths):
     for filepath in filepaths:
         proposal = filepath.split('/')[-1][2:7]
         if proposal not in proposals:
+
+            # Get a list of all integ0.thumb files for the instrument
+            # Note that this assumes all filepaths will be for the same
+            # instrument.
+            instrument = os.path.basename(filepath).split('_')[3]
+            base_dir = os.path.dirname(filepath)
+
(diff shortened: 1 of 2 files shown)


mfixstsci:
Hey @bhilbert4 is this issue fixed in #927?

bhilbert4:
This is separate from #927. Prior to commissioning there were cases where the thumbnail images gave the wrong number of exposures in a given program. Now that we have commissioning data I can check again and see if JWQL is still mis-counting.

bhilbert4:
It looks like this is still an issue. I spot checked some commissioning programs. The MIRI and NIRCam programs I checked were ok, but the NIRISS programs (01145, 01079) had more files reported in the program thumbnail than were actually present. I wonder if parallels are screwing up the count somehow.

bhilbert4:
This issue was solved in #942.

PR #825, opened 2021-10-14
[WIP] Build Grating Wheel Sensor Telemetry Monitor

tnking97:


jwql/tests/test_grating_monitor.py
```diff
@@ -0,0 +1,75 @@
...
+
+    assert grating_val == 'MIRROR'
+    assert adcmgain == 2.5001527415
+    assert adcmoffset == -0.0010681153
+    assert motor_vref == -0.0009155273999999999
+    assert x_pos == 173.1699688
+    assert y_pos == 98.6046408
+
+
+# def test_process():
```
bourque, line 65: Testing functions like these with unit test are quite difficult since the functions do a lot of stuff and touch external subsystems like the database as you mention. I would submit that you don't need to test this.  We will know if something is wrong in the `process` function if the monitor itself fails when running on the dev and test servers.

bourque (commented):
@tnking97 I just did an initial review of this and it looks really good!  All of the major pieces are there.  It is well organized, documented, and tested, and I don't see any major changes that need to be made.  Excellent work!

Issue #764, opened 2021-07-20
Add new MIRI readout patterns: [Bug]


cracraft:
Issue:
This is not a bug yet, but there may be some test data in the latest DMS builds that could show this up to be a bug quickly. The MIRI team has added a set of new readout patterns that will be used in flight. These need to be added to the software so that they will be recognized and not crash the system.

The current patterns are FAST, SLOW and FASTGRPAVG.
The new readout patterns that need to be added are: FASTR1, SLOWR1, FASTGRPAVG8, FASTGRPAVG16, FASTGRPAVG32 and FASTGRPAVG64. The first two are going to be the standard default values, while the rest are going to be for TA, but all need to be recognized by the system. FASTR100 should also be an allowed mode, even though it will only be used in specific cases.

At this point, we're not asking that FAST and SLOW be removed, since we will be taking those in commissioning, as well as that being what has been used as the default values up to this point, so any test data currently in the system would crash.

Hopefully this is just a simple matter of adding these to a dictionary file as valid MIRI readout patterns.

cracraft:
Also, for the current list of APERTURES in the constants file, this is what is listed for MIRI.
'MIRI': ['MIRIM_FULL', 'MIRIM_MASKLYOT', 'MIRIM_SLITLESSPRISM', 'MIRIM_SUB256',
                                     'MIRIM_SUB128', 'MIRIM_SLIT']

If this is meant to cover all subarrays, there are a few missing: MIRIM_BRIGHTSKY, MIRIM_SUB64, MIRIM_MASK1065, MIRIM_MASK1140, MIRIM_MASK1550

PR #826, opened 2021-10-15
Updated MIRI readout patterns

cracraft:
Updated MIRI readout patterns as specified in Issue #764.


cracraft:
Ok, while we don't have it in our reference files yet, it can apparently be selected in engineering modes in APT, so we should add FASTR100 just to cover all bases.

PR #828, opened 2021-10-21
Update EDB

mfixstsci:
- Removing reliance on `jwst-dms-edb`
- Adding infrastructure to obtain `base_url` for MAST API calls
- updating `engineering_database.py` and adding needed code from `jwst-dms-edb`


bhilbert4:
What happens if you input a mnemonic that doesn't exist?

mfixstsci:
> What happens if you input a mnemonic that doesn't exist?
Screen Shot 2021-10-29 at 4 56 48 PM
@bhilbert4 it looks like there is a form validation rule set for that! Same if `start_date >= end_date`

bhilbert4:
Huh. RTD still used docutils 0.18. Maybe it needs to be added to rtd_requirements.txt also?

mfixstsci:
> Huh. RTD still used docutils 0.18. Maybe it needs to be added to rtd_requirements.txt also?

It was the `rtd_requirements` that needed to be updated!

mfixstsci:
🎊 all checks are passing now!

bourque:
Apologies for letting this sit around @mfixstsci .  I'm hoping to review this tomorrow or early next week.

mfixstsci:
Hey @bourque, I went ahead and merged in some changes/resolved some conflicts from the dependency updates.

bourque (commented):
The code changes here look great, thanks @mfixstsci!  I just want to make sure this works on the dev web server before merging.  But assuming all is well there, this should be good to go.

bhilbert4:
Trivial detail, but it might be helpful to tweak the wording in the bottom panel of the EDB page where it reports how many matches were found for a particular search. The text is, for example: "Query returned 16 matching rows in the EDB inventory containing 14947 mnemonics." It took me a few reads to realize that this is saying the entire EDB contains 14947 mnemonics, and not that it was miscounting the number of returned entries. I think mixing the terms "rows" and "mnemonics"  is misleading. It might be more clear to have something like, "Query returned 16 matching mnemonics from the EDB inventory of 14947 mnemonics." Or just don't bother saying the total number of mnemonics in the EDB.

jwql/utils/credentials.py
```diff
@@ -23,6 +23,40 @@
...
+
+    Returns
+    -------
+    token : str or None
+        Base url for MAST JWST EDB API.
+    """
+    if Mast.authenticated():
+        print('Authenticated with Astroquery MAST magic')
+        return None
+    else:
```
bourque, line 42: It appears that these lines are causing the second panel of the EDB page to break on the dev, test, and ops web servers.  On those servers, this `if` statement passes, and so the function returns `None`, which causes the `base_url` in `engineering_database.py` to be the default one, which doesn't currently work.  Thus, we need this function to return the `base_url` provided either in the `config.json` file or the environment variable.

As such, I think the only changes necessary here are to simply remove these lines that I've highlighted (and properly indent the rest of the function).

bourque (changes requested):

bourque:
Great suggestions @bhilbert4.  @mfixstsci, could you also address Bryan's comment in this PR?

jwql/utils/credentials.py
mfixstsci: Went ahead and removed the logic. We are getting some test failures. The last run failure was because of a timeout. I Just kicked them off again so we will see what happens 🤞

bourque (approved):
Thanks for all the changes @mfixstsci!  This is ready to jetty pending successful CI builds.

mfixstsci:
@bourque everything is passing :fire:

mfixstsci:
Thank you @bourque 😎

PR #829, opened 2021-10-25
Cross-origin request fix

### LINKED ISSUE #822: Update cross-origin request in test/ops web servers
Currently the test and ops servers do not render the instrument archive or proposal webpages.  I think it is related to this error that shows up on the web console when loading those pages:

```
Cross-Origin Request Blocked: The Same Origin Policy disallows reading the remote resource at https://<dev server>/ajax/FGS/archive/. (Reason: CORS header ‘Access-Control-Allow-Origin’ missing)
```

I'm thinking that maybe there is something hard-coded in the web app code that is pinging the dev server for requests from the test/ops servers.  This should be fixed.

bourque:
Closes #822


bourque:
I was hoping to fold in any fixes related to the dev/test/ops `ajax` issue in this PR, but since that is an ongoing issue and not necessarily related to #822, I think this is ready to merge.

bhilbert4:
See here for a possible explanation of RTD build failures. RTD installs multiple versions of setuptools. It does remove the previous version each time it installs a new one, so perhaps this is not our issue, but it seems to be worth mentioning. https://github.com/pypa/setuptools/issues/2957

bourque:
Per @bhilbert4's above comment, I am going to merge this in despite the RTD failures in hope that the issue will soon be solved via pypa/setuptools#2957 and related PRs.  If this failure continues to happen towards the end of the `v1.1` sprint, we can revisit this.

bourque:
Nevermind my previous comment.  I was able to get the RTD build passing by unchecking the "Use system packages" settings in the RTD advanced settings 🤷

PR #833, opened 2021-11-16
Add anomaly msg

### LINKED ISSUE #758: Add 'submit sucessful' message when submitting anomalies
Currently, when a user submits an anomaly on the `view_image` view, the page is refreshed and nothing else happens.  It would be nice to have some sort of success message pop up if the anomaly submission was successful so that the user knows that something actually happened.

shanosborne:
Adding a message when an anomaly is submitted.
* Check box(es) and hit submit -> get successful pop up
* Check no boxes and hit submit -> get "This field is required” warning (This was already set up, but the coded errored  before it got to displaying the warning)
* If the form isn’t valid -> get an error pop up

Pop ups can be closed manually, otherwise they close automatically after 5 seconds

Closes #758


bourque:
Thank you @shanosborne!  I plan to review this sometime today or tomorrow!

jwql/website/apps/jwql/templates/base.html
```diff
@@ -174,6 +174,23 @@
 
 		<div id="main_content"></div>
 
+		<!--Handle incoming messages-->
```
bourque, line 177: Neat!  I like that this is concise and we can put it in `base.html`

bourque (approved):
Awesome work here @shanosborne, looks great! And works great when I test it out locally.  This is good to go pending a successful build.

Issue #834, opened 2021-12-02
Bias Monitor plots crashing in web app due to None data


bsunnquist:
The NIRCam Bias Monitor plots fail to render on the web app due to some NoneType data in the bias stats arrays.
![image](https://user-images.githubusercontent.com/16088329/144502879-5fccb825-1c19-4b91-9710-f466445154ee.png)

PR #836, opened 2021-12-03
Made plotting limits robust to missing data

bsunnquist:
This PR should fix https://github.com/spacetelescope/jwql/issues/834 by making the scale limits calculation for the bias and readnoise monitors robust to missing data.

The data that caused this issue shouldn't exist in flight - it was a subarray with incorrect header info calling it full frame, which doesnt have multiple output channels so those database columns were blank for it, causing the crash.


bsunnquist:
@bourque this is ready for review whenever

bourque (approved):
Thanks @bsunnquist!

PR #839, opened 2021-12-13
[WIP] Include level 1b products in mast query; reorganize mast_utils.py

tnking97:


tnking97:
With recent MAST updates, it looks like we will have to avoid using the Ops cluster. I have been in communication with Dick Shaw about this PR, and have forwarded additional in-depth information regarding the PR/conversation with Dick Shaw to Matthew Bourque. The updated function should be implemented in modules including `test_monitor_mast`, `data_containers`, `monitor_template`, and `monitor_mast`.

bhilbert4:
[jwst_mast_query](https://github.com/spacetelescope/jwst_mast_query/blob/main/jwst_mast_query/jwst_download.py) is a package that queries and downloads data from MAST. In that case, users can specify the file suffixes to be downloaded (or by specifying just "fits" all suffixes will be downloaded). So it might be worth looking at how that calls astroquery and make any necessary tweaks to JWQL to assure the same behavior.

bhilbert4:
@mfixstsci do you agree that this no longer seems to be an issue? I haven't noticed any missing uncal files anywhere.

mfixstsci:
Agreed @bhilbert4 thanks for taking a look.

PR #841, opened 2021-12-13
Switch to test string for astroquery.mast queries

### LINKED ISSUE #840: Switch to astroquery.Mast test base url
The archive has removed rehearsal data from their OPS environment(s) in preparation for JWST commissioning.  This causes our nominal `astroquery.mast` queries to return no data.  There is a backend attribute to the `Mast` class that we can use to switch to a different test archive area that contains the rehearsal data.

bourque:
Closes #840


bourque:
Since this is a major roadblock for other work, I am going to bypass a review here and put this into production promptly.  There are not any major changes here, just a temporary few lines of code to set an attribute that we can remove when it comes time for JWST operations.

PR #843, opened 2021-12-20
Import initialize_instrument_monitor from the correct location

bhilbert4:
This PR fixes an error seen in the cron job logs recently:

File "jwql/jwql_monitors/generate_proposal_thumbnails.py", line 34, in <module>
    from jwql.utils.utils import get_config, initialize_instrument_monitor
ImportError: cannot import name 'initialize_instrument_monitor' from 'jwql.utils.utils' (jwql/utils/utils.py)

With this fix, initialize_instrument_monitor is now imported from the correct location. Looks like this must have been missed in #804.


bhilbert4:
ARG. Looks like setuptools version 60 was released earlier today. I'm betting this is why RTD is failing.

bhilbert4:
It looks to me like the `mock` package being installed is what is causing setuptools 60.0.1 to be installed in all cases. But version 1.0.1 of mock, which is the version being installed, is from 2012!!?

bhilbert4:
This RTD failure looks very similar to what happened when they released version 50. It looks like RTD is installing 60.0.1, then JWQL specifies version 58, but perhaps the path to version 60 is before the path to version 58?
https://github.com/pypa/setuptools/issues/2353

bhilbert4:
Also note that `jwst` requires setuptools < 58.

bhilbert4:
I suspect that this is the cause of our failing RTD builds: https://github.com/pypa/setuptools/issues/2957
RTD installs the latest version of setuptools, but then JWQL (and also `jwst`) request older versions of setuptools.

bourque (approved):
Thanks @bhilbert4, looks good!

PR #844, opened 2021-12-29
Fix archive webpage ajax issue

bourque:
This PR fixes an issue in which the archive webpages were not loading on the dev/test/ops servers.  It also updates `constants.py` with a newly-encountered suffix type for segmentation map files.


bourque:
I've tested this on the dev server and it appears to be working great!  The archive and thumbnail pages work for every instrument besides NIRCam.  For NIRCam, the servers experience some sort of timeout/502 error, but I believe this error is unrelated and should be solved in a separate PR.

PR #846, opened 2021-12-30
[WIP]: Speed up archive webpages

### LINKED ISSUE #816: Improve loading times throughout the web app
There are various pages in the web app (particularly the instrument archive and specific proposal webpages) that take a long time to load, even on the dev/test/ops servers.  We should investigate the reason(s) behind this and try to make improvements.

bhilbert4:
This PR contains edits designed to speed up the creation and display of the archive web pages. It seems like the main changes that need to be made are reducing the number of MAST queries being made. Timing tests on the test server recently showed that each MAST query performed within get_filenames_by_instrument() was taking 90 seconds.

The only change made so far is to cut the number of MAST queries in get_filenames_by_instrument() from 2 to 1. Most likely more changes will be necessary.

Resolves #816

### DIFF
--- a/jwql/website/apps/jwql/data_containers.py
+++ b/jwql/website/apps/jwql/data_containers.py
@@ -476,7 +476,18 @@ def get_expstart(instrument, rootname):
     return expstart
 
 
-def get_filenames_by_instrument(instrument, restriction='all'):
+def mast_query(instrument, colnames, filters=[]):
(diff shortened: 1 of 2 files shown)


bhilbert4:
I think this may be superseded by #867. Might be best to keep it around until that is merged, but I think this can be closed after that.

bhilbert4:
#867 has been merged. This PR is no longer needed.

PR #857, opened 2022-01-24
Create symlinks to filesystem in web application static directory.

### LINKED ISSUE #851: Add code to manage.py to create necessesary symlinks to run web app locally
In order to run the JWQL web app locally, one must create symbolic links to the `outputs`, `thumbnails`, `preview_images`, and `filesystem` directories.  We can add some code in `website.manage.py` in order to do this automatically.  Something like this:

```python
from jwql.utils.utils import get_config()

if __name__ == "__main__":

    os.environ.setdefault("DJANGO_SETTINGS_MODULE", "jwql_proj.settings")

    # Create symbolic links here (if they don't already exist)
    for directory in ['filesystem', 'outputs', 'preview_image_filesystem', 'thumbnails_filesystem']:
        path = get_config()[directory]
        # code to create symlink

    try:
        from django.core.management import execute_from_command_line
    except ImportError as exc:
        raise ImportError(
            "Couldn't import Django. Are you sure it's installed and "
            "available on your PYTHONPATH environment variable? Did you "
            "forget to activate a virtual environment?"
        ) from exc
    execute_from_command_line(sys.argv)
```

Credit @york-stsci for the suggestion!

york-stsci:
Closes #851

This adds code to the manage.py file that will look for the static filesystem links (to the filesystem itself, to the outputs directory, to the preview images, and to the thumbnails), and will create symlinks based on the config.json items of the same keyword if they don't already exist.


jwql/website/manage.py
```diff
@@ -36,10 +36,25 @@
 import os
 import sys
 
+from jwql.utils.utils import get_config()
```
bourque, line 39: I think this just needs to be `get_config` instead of a function call to `get_config()`

jwql/website/manage.py
```diff
@@ -36,10 +36,25 @@
...
 
+    directory_mapping = {
+        'filesystem': 'filesystem',
+        'outputs': 'outputs',
+        'preview_image_filesystem': 'preview_images',
+        'thumbnail_filesystem': 'thumbnails'
+        }
+
+    for directory in ['filesystem', 'outputs', 'preview_image_filesystem', 'thumbnail_filesystem']:
+        symlink_location = os.path.join(get_config()['jwql_dir'], directory_mapping[directory])
```
bourque, line 53: If I am understanding this correctly, `symlink_location` should be pointing to the location in which the symlinks will be written _within the local clone of the repo_.  If that is the case, then I think this line needs to be changed to be relative to the local clone's `static` path, instead of the `jwql_dir`, as `jwql_dir` points to the central storage location.

bourque (changes requested):
Thanks @york-stsci, this is looking pretty good, but I just have a few comments/suggestions.

jwql/website/manage.py
york-stsci: 100% correct. Change made.

jwql/website/manage.py
york-stsci: Yes, that's definitely correct, and I should have picked that up. The existing code areas that use the symlinks do things via the config file, and I expect it's exactly the same when it comes to the active server, but we should do it differently. Is there a standard way to refer to the local directory, or should I be building a path based on the location of __file__?

bourque: Right, the best way I can think to do it at the moment is building a path based on the file location, something like this:

```python
symlink_location = os.path.join(os.path.dirname(__file__), 'apps/jwql/static/', directory_mapping[directory)
```

york-stsci: Fixed I think.

bourque (approved):
Thanks for the changes @york-stsci, this looks and works great!

PR #867, opened 2022-02-04
[WIP]: Speed up app loading 

bhilbert4:
This PR modifies the code in an attempt to speed up webapp loading times.

### DIFF
--- a/jwql/jwql_monitors/generate_proposal_thumbnails.py
+++ b/jwql/jwql_monitors/generate_proposal_thumbnails.py
@@ -34,6 +34,7 @@ from jwql.utils.logging_functions import configure_logging, log_info, log_fail
 from jwql.utils.utils import get_config
 from jwql.utils.monitor_utils import initialize_instrument_monitor, update_monitor_table
 
+SETTINGS = get_config()
 
 @log_fail
 @log_info
@@ -41,7 +42,7 @@ def generate_proposal_thumbnails():
     """The main function of the ``generate_proposal_thumbnails`` module.
     See module docstring for further details."""
 
-    proposal_dirs = glob.glob(os.path.join(get_config()['thumbnail_filesystem'], '*'))
+    proposal_dirs = glob.glob(os.path.join(SETTINGS['thumbnail_filesystem'], '*'))
 
     for proposal_dir in proposal_dirs:
         rate_thumbnails = glob.glob(os.path.join(proposal_dir, '*rate*.thumb'))
--- a/jwql/website/apps/jwql/data_containers.py
+++ b/jwql/website/apps/jwql/data_containers.py
@@ -41,7 +41,7 @@ from jwql.database.database_interface import load_connection
 from jwql.edb.engineering_database import get_mnemonic, get_mnemonic_info
 from jwql.instrument_monitors.miri_monitors.data_trending import dashboard as miri_dash
 from jwql.instrument_monitors.nirspec_monitors.data_trending import dashboard as nirspec_dash
-from jwql.utils.utils import ensure_dir_exists, filesystem_path, filename_parser, get_config
+from jwql.utils.utils import check_config_for_key, ensure_dir_exists, filesystem_path, filename_parser, get_config
 from jwql.utils.constants import MONITORS
 from jwql.utils.constants import INSTRUMENT_SERVICE_MATCH, JWST_INSTRUMENT_NAMES_MIXEDCASE, JWST_INSTRUMENT_NAMES_SHORTHAND
 from jwql.utils.preview_image import PreviewImage
@@ -56,7 +56,6 @@ ON_READTHEDOCS = False
 if 'READTHEDOCS' in os.environ:
     ON_READTHEDOCS = os.environ['READTHEDOCS']
 
-from jwql.utils.utils import get_config, filename_parser, check_config_for_key
 if not ON_GITHUB_ACTIONS and not ON_READTHEDOCS:
     from .forms import MnemonicSearchForm, MnemonicQueryForm, MnemonicExplorationForm
     check_config_for_key('auth_mast')
@@ -663,6 +662,10 @@ def get_image_info(file_root, rewrite):
     filenames.extend(glob.glob(os.path.join(FILESYSTEM_DIR, 'proprietary', proposal_dir, observation_dir, '{}*.fits'.format(file_root))))
     image_info['all_files'] = filenames
 
+    # Determine the jph directory
+    prev_img_filesys = get_config()['preview_image_filesystem']
+    jpg_dir = os.path.join(prev_img_filesys, proposal_dir)
+
     for filename in image_info['all_files']:
 
         # Get suffix information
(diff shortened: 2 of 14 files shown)


bhilbert4:
@bourque based on the timing test results in #816, what do you think about maintaining files with lists of existing preview images/thumbnail files?

bhilbert4:
Local tests with just the get_config() fix seem to show the archived images pages are loading noticeably faster than before. Individual program pages are still a little slow. But I should repeat the tests on one of the servers to see what happens there.

bhilbert4:
@bourque assuming all tests pass here, I think this is ready for review. I think we'll need to manually create the initial versions of the preview image and thumbnail inventory files on the production server (as I did on the test server) when it is ready to deploy.

jwql/website/apps/jwql/data_containers.py
```diff
@@ -767,15 +789,18 @@ def get_preview_images_by_instrument(inst):
...
+    # images for a single instrument.
+    preview_list_file = f"{PREVIEW_IMAGE_LISTFILE}_{inst.lower()}.txt"
+    preview_images = retrieve_filelist(os.path.join(PREVIEW_IMAGE_FILESYSTEM, preview_list_file))
 
+    # DO WE NEED TO DO THIS??? I GUESS THIS GETS RID OF ANY PREVIEW IMAGES
+    # THAT DON'T MATCH UP TO OBSERVATION FILES.
     # Get subset of preview images that match the filenames
-    preview_images = [os.path.basename(item) for item in preview_images if
-                      os.path.basename(item).split('_integ')[0] in filenames]
+    #preview_images = [os.path.basename(item) for item in preview_images if
```
bhilbert4, line 800: I'm not sure about this. I guess we do still need to filter the instrument-specific list of preview image filenames by file roots from the list of observations?

mfixstsci (commented):
Hey @bhilbert4 this looks good. Just had a few questions and comments. When addressing the commented code, could you also look at the messages from PEP8? Thanks for submitting this!

jwql/jwql_monitors/generate_proposal_thumbnails.py
```diff
@@ -34,14 +34,15 @@
 from jwql.utils.utils import get_config
 from jwql.utils.monitor_utils import initialize_instrument_monitor, update_monitor_table
 
+SETTINGS = get_config()
```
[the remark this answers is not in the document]
bhilbert4: Yes. The idea behind this is that I found that running get_config() actually takes quite a while. In one instance there was a call to get_config() inside of a loop over a list of files, and this was adding a significant amount of run time to the script. So my thought was to call get_config() once and have the results as a  global so that all of the functions could then simply access the resulting dictionary.

jwql/website/apps/jwql/data_containers.py
```diff
@@ -514,12 +515,22 @@ def get_filenames_by_instrument(instrument, restriction='all'):
...
     """
 
-    service = INSTRUMENT_SERVICE_MATCH[instrument]
+    #service = INSTRUMENT_SERVICE_MATCH[instrument]
 
-    # Query for filenames
-    params = {"columns": "filename, isRestricted", "filters": []}
-    response = Mast.service_request_async(service, params)
-    result = response[0].json()
+    if not query_file and not query_response:
```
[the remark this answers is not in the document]
bhilbert4: Oops. Looks like I forgot to remove that commented code. I'll do that.

bhilbert4:
Regarding PEP8 issues above:
The issues in [organize_filesystem.py](https://github.com/spacetelescope/jwql/blob/237879d4d9095ead4ed0a003acee8e76eea1b74f/jwql/utils/organize_filesystem.py) are due to commented out commands. These were commented out before this PR. It's not clear to me why they are commented out or whether we should uncomment them. Might be worth some investigation.

The issues in preview_image.py are due to having import statements that depend on whether the module is being called as part of GitHub Actions or not. These lines didn't change in this PR, so I'm tempted to ignore PEP8 here.

bhilbert4:
In data_containers.py, the bare except issue predates this PR. It would take some investigation to see what error types need to be caught.

bhilbert4:
And for the line break issues in data_containers.py, I've tried several different ways while following examples given in the PEP8 documentation itself, and it has still complained for all of them. I'm not sure what else to do for it...

jwql/website/apps/jwql/data_containers.py
```diff
@@ -496,7 +497,7 @@ def get_expstart(instrument, rootname):
     return expstart
 
 
-def get_filenames_by_instrument(instrument, restriction='all'):
+def get_filenames_by_instrument(instrument, restriction='all', query_file=None, query_response=None):
     """Returns a list of filenames that match the given ``instrument``.
 
     Parameters
```
mfixstsci, line 503: @bhilbert4 what is this query file parameter? Would you be able to update the docstring here?

jwql/website/apps/jwql/data_containers.py
mfixstsci: This is interesting. I would say we could tie this up with the new issue we plan to open for the other commented code. Do you think this could be a hold over from the pre-launch filesystem? I know we had a lot of file mismatches prior to launch with old pre-flight data hanging around that wasn't in MAST, right?

jwql/website/apps/jwql/data_containers.py
```diff
@@ -907,11 +921,12 @@ def get_thumbnails_all_instruments(parameters):
         # Query MAST for all rootnames for the instrument
         service = "Mast.Jwst.Filtered.{}".format(instrument)
 
-        if (parameters['apertures'][inst.lower()] == []) and (parameters['detectors'][inst.lower()] == []) \
-            and (parameters['filters'][inst.lower()] == []) and (parameters['exposure_types'][inst.lower()] == []) \
-            and (parameters['read_patterns'][inst.lower()] == []):
-                params = {"columns": "*",
-                          "filters": []}
+        if ((parameters['apertures'][inst.lower()] == [])
```
mfixstsci, line 924: This is much more readable, thank you for doing this.

mfixstsci (approved):
Hey @bhilbert4 all of the new changes look good. Thanks for making certain parts more readable and I like the use of the `lambda` function to sort the files by `expstart`.

jwql/website/apps/jwql/data_containers.py
bhilbert4: Thinking about this more, I suspect we should keep it. This is going to remove any preview images that don't match up to a base name in the list of fits files in the filesystem. I think that makes sense to do in that, if we have a preview image file for a file that is no longer in the filesystem, or some non-standard preview image, it will be ignored. I think that's desirable. It's not really clear to me off-hand how a preview image like that would get into the filesystem to begin with barring manual creation.

bhilbert4: I can see this line taking a long time to run down the road when we have boatloads of data. I would put this towards the top of a list of code to revisit if we need to speed up the web app in the future.

jwql/website/apps/jwql/data_containers.py
bhilbert4: PEP8 still doesn't like it. But it does match an example given in the PEP8 documentation...

mfixstsci (approved):

PR #868, opened 2022-02-04
Plot non-numerical mnemonic dataset

mfixstsci:
Many of the mnemonic datasets used by the EDB contain `date` and non-numerical `euvalues`.

Current approach:
* If unique possible values of `euvalues` is `> 1 & <=4` then we will plot the data.
* If unique possible values of `euvalues` is `> 4` then we will display a table
* If values are numeric, just generate the expected figure.

### DIFF
--- a/jwql/edb/engineering_database.py
+++ b/jwql/edb/engineering_database.py
@@ -44,12 +44,9 @@ Notes
 """
 
 from collections import OrderedDict
+import copy
 from datetime import datetime
-from enum import unique
-# from msilib.schema import tables
 import os
-from select import POLLOUT
-from tokenize import PlainToken
 import warnings
 
 from astropy.table import Table
@@ -58,6 +55,7 @@ from astroquery.mast import Mast
 from bokeh.embed import components
 from bokeh.plotting import figure, show
 import numpy as np
+import tempfile
 
 from jwst.lib.engdb_tools import ENGDB_Service
 from jwql.utils.credentials import get_mast_base_url, get_mast_token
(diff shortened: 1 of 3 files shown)


mfixstsci:
Ready for review, I copied the code that "Explore the EDB mnemonic inventory" uses to generate the table. I am sure there is a way to synergize the code there and in the method I wrote to generate a table for mnemonic data with too many unique text values.

Examples to Test:
------------------
* Defaults will return expected plot
* Test parameters that only has 3 or 4 unique values (`SA_ZSLEWST`, `2021-09-01 00:00:00.000`, `2021-09-03 00:01:00.000`)
* Test parameters for text that has too many unique strings to plot (`ICTM_EVENT_MSG`, `2021-09-01 00:00:00.000`, `2021-10-01 00:01:00.000`)

Try it out locally and tell me what you think!

jwql/edb/engineering_database.py
```diff
@@ -142,6 +145,64 @@ def bokeh_plot(self, show_plot=False):
...
+            script, div = components(p1)
+
+            return [div, script]
+
+    def get_table_data(self):
+        """Get data needed to make interactivate table in template.
+        """
+
+        import tempfile
+        import copy
```
bourque, line 191: Could these be moved up to the top of the module with the other imports?

jwql/edb/engineering_database.py
```diff
@@ -45,7 +45,11 @@
 
 from collections import OrderedDict
 from datetime import datetime
+from enum import unique
+# from msilib.schema import tables
 import os
+from select import POLLOUT
+from tokenize import PlainToken
```
bourque, line 52: Could these be removed?  I don't see them being used in the code.

jwql/edb/engineering_database.py
```diff
@@ -45,7 +45,11 @@
 
 from collections import OrderedDict
 from datetime import datetime
+from enum import unique
```
bourque, line 48: I see `numpy.unique` being used in the code, but I don't see `enum.unique` being used.  Could this be removed?

bourque (changes requested):
@mfixstsci Thanks for these changes, they look and work great when I test them out!  I just have a few questions in the review.

bourque (approved):
Thanks for the updates @mfixstsci, this looks great!

PR #871, opened 2022-02-14
Ignore original.fits files

### LINKED ISSUE #870: "original" filetypes are being picked up by the web app
The preview image pages of the web app show `original` as a filetype option, but JWQL doesn't support these filetypes as I believe they are some sort of copy of a file that the archive uses that is not relevant to the instrument teams.  We should fix the code as to ignore these filetypes.

bhilbert4:
Resolves #870 

Tweak the web app code to ignore "original.fits" files. It looks like we were already ignoring them when creating preview images. https://github.com/spacetelescope/jwql/blob/develop/jwql/jwql_monitors/generate_preview_images.py#L646
But I think we were getting around that exclusion with the call to get_image_info(). 

- [x] Any other suffixes we should ignore? I added "stream" as well, which I saw a handful of times in the test filesystem.
- [x] Update generate_preview_images to use the new constants as well, for consistency.
- [x] Search for other glob calls where we may need to explicitly ignore these suffixes


bourque:
Thanks for taking this on @bhilbert4!

I agree that `stream` files should probably be ignored.  Not sure what those are, and they don't appear to be documented in the [pipeline docs](https://jwst-pipeline.readthedocs.io/en/stable/jwst/introduction.html#pipeline-step-suffix-definitions)

jwql/utils/constants.py
```diff
@@ -215,6 +215,9 @@
 # Possible suffix types for guider exposures
 GUIDER_SUFFIX_TYPES = ['stream', 'stacked_uncal', 'image_uncal', 'stacked_cal', 'image_cal']
 
+# JWQL should ignore some filetypes in the filesystem.
+IGNORED_SUFFIXES = ['original', 'stream']
```
bourque, line 219: Great idea to store this in `constants`!

bhilbert4:
This seems to be working well. Tests on the test server show that 'original' files no longer show up in the list of filetypes when looking at a particular exposure.

My one remaining question is about the lines in data_containers.py in get_image_info() that are commented out. These lines would create a preview image for a given file if that preview image doesn't already exist. Those lines are already commented out (i.e. I'm not changing them to be commented out here). It seems like it would be a good thing to run. But when I uncomment the lines, I get an error about the filesystem being read-only. I think this is why they were commented out in the first place. It would be a nice feature to have though.

bhilbert4:
Outside of the question above, this is ready for review @bourque

jwql/utils/constants.py
mfixstsci, line 219: Agreed with @bourque, have we always ignored `stream` files as well?

mfixstsci (commented):
Hey Bryan, these changes look good thanks for submitting them. I am going to wait to merge once we get the dependency updates in!

jwql/utils/constants.py
bhilbert4: I thought I saw one on an observation's page at one point. Anyway, I figure this can help us be sure to ignore them.

mfixstsci:
> This seems to be working well. Tests on the test server show that 'original' files no longer show up in the list of filetypes when looking at a particular exposure.
> 
> My one remaining question is about the lines in data_containers.py in get_image_info() that are commented out. These lines would create a preview image for a given file if that preview image doesn't already exist. Those lines are already commented out (i.e. I'm not changing them to be commented out here). It seems like it would be a good thing to run. But when I uncomment the lines, I get an error about the filesystem being read-only. I think this is why they were commented out in the first place. It would be a nice feature to have though.

Maybe we can open a separate issue and figure this out? Get in touch with IT?

mfixstsci (approved):

Issue #873, opened 2022-02-16
Integration slider bugs/enhancements


bsunnquist:
Issues/suggestions for the integration slider from Jo's email:

1. Integration slider does not work (e.g. this file has 25 integrations https://jwql.stsci.edu/NIRISS/jw01078002001_02103_00001_nis/)
2. for the integration slider, show how many integrations there are as you go through, e.g. 1/10, 2/10. Or just show 1/1 if only 1
3. Integration number in slider is different than integration in image title & filename (0 vs 1 indexed)

bsunnquist:
https://github.com/spacetelescope/jwql/pull/876 fixes 1 and 2

PR #876, opened 2022-02-17
Fix/modify integration slider

bsunnquist:
This PR fixes the issue of the integration slider not working on preview images. It also adds the total ints to the output.


bsunnquist:
This addressed points 1 and 2 in https://github.com/spacetelescope/jwql/issues/873

This slider also had a bug for images with >10 ints, so I fixed that too

bsunnquist:
@bourque this is ready for review whenever

bsunnquist:
@bourque ok, it should work now no matter how many pngs/ints there are (and should also still work if we increase that ratio in the future). The only thing I couldn't get working quite right was the live update of the slider number as you drag it.

Issue #891, opened 2022-02-23
Check that database connections are being closed


bourque:
We recently had an issue come up in which the production database server was unable to support new connections because there were currently 100 open connections.  Perhaps there is something in our code that is not closing database connections after using features such as the `JWQLDB`

mfixstsci:
I believe I found the culprit. In `jwql.website.apps.jwql.data_containers.build_table` is what I use to obtain the data but I don't tear the `session` object down afterwards.

PR #892, opened 2022-02-23
Closing Session in build_table Function

mfixstsci:
Addresses #891. I am looking around the web application and noticed that we have a lot of sessions that are left open after we perform queries. I am unsure if this will solve our problems immediately, but it's worth a shot.


mfixstsci:
Went ahead and closed all of the sessions. I made sure to do it at the end of the class methods/functions to not interrupt any data collection.

bourque (approved):
@mfixstsci Thanks so much for this speedy fix, looks good!

PR #908, opened 2022-03-25
Reduce memory usage by the dark monitor

### LINKED ISSUE #905: Failing monitors - probably due to memory issues with large files
It looks like all instrument monitors except the bias monitor are failing. In most cases, the logs from the monitor runs are not helpful. They show that the cal pipeline or some other calculation was starting, and that's the end of the log, implying that they are failing silently. 

For the bad pixel monitor, there is a line in the log saying that there was not enough memory to create an array for the data in the file being worked on. (It needed 1.5 GB). I looked in the logs for the other monitors and in all cases, the monitors were working on especially large files at the time that they failed. e.g. 4 integrations x 50 groups, 20 integrations x 10 groups. 

We may need to adjust the instrument monitors to be able to handle these cases. One option would be for the monitors to skip these files entirely. Or perhaps we can create a module that checks the number of groups and integrations and if a file is too large, read in and work on only a portion of the data? Unfortunately for the dark monitor, the failure came while the ramp-fitting step of the pipeline was running. I'm not sure how to get around that other than skipping the file entirely. We should check and see how large the official commissioning and cycle 1 cal dark data will be, in order to know whether this is going to continue to be a problem.

@bsunnquist

bhilbert4:
This PR makes adjustments to the dark current monitor in order to try and reduce the overall memory usage. I'm afraid there isn't much we can do here. The highest memory usage comes when running the cal pipeline on the input files. 

One option would be to skip the cal pipeline and fall back to using (last group - first group) as an approximation of the slope image. 

- [ ] ~~Look into what effects using the difference image would make.~~ 
- [ ] ~~Would CRs throw off the statistics since we'd be skipping the jump step~~
- [ ] ~~Would the dark rate histograms still be useful~~

This is being done as part of the effort to resolve #905


bhilbert4:
I think this is all we can really do for now. I think we should review and merge this, and hopefully #921 will take care of residual memory errors.

jwql/instrument_monitors/common_monitors/dark_monitor.py
```diff
@@ -522,6 +522,9 @@ def process(self, file_list):
         mean_slope_file = self.save_mean_slope_image(slope_image, stdev_image, slope_files)
         logging.info('\tSigma-clipped mean of the slope images saved to: {}'.format(mean_slope_file))
 
+        # Free up memory
+        del slope_image_stack
```
mfixstsci, line 526: Nice, out of curiosity, how much memory does this free up?

mfixstsci (approved):

jwql/instrument_monitors/common_monitors/dark_monitor.py
bhilbert4: If depends on how many exposures were used in the dark current calculations. This deletes the stack of input images from which a mean image is constructed.

mfixstsci: Cool thanks! Merging.

PR #909, opened 2022-03-29
Add JWQL Python 3.9 Environment

mfixstsci:
- Created 3.9 environment
- Removed support for python 3.7
- Updated README docs

### DIFF
--- a/environment_python_3_9.yml
+++ b/environment_python_3_9.yml
@@ -1,32 +1,32 @@
 # This file describes a conda environment that can be to install jwql
 #
 # Run the following command to set up this environment:
-# $ conda env create -f environment_python_3_7.yml
+# $ conda env create -f environment_python_3_9.yml
 #
 # The environment name can be overridden with the following command:
-# $ conda env create -n <custom name> -f environment_python_3_7.yml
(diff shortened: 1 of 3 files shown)


mfixstsci:
@bhilbert4 I went ahead and updated the python 3.9 environment file to match 3.8. Since April is around the corner I figured we could just wait to see what pyup has to say. This is ready for review.

environment_python_3_9.yml
```diff
@@ -42,7 +42,7 @@ dependencies:
   - psycopg2=2.8.6
   - pytest=6.2.4
   - pytest-cov=3.0.0
-  - python=3.7.10
+  - python=3.8.5
```
bhilbert4, line 45: This should be 3.9.11?

bhilbert4 (approved):
Looks good to me. I'm guessing the next dependency update will probably include the same updates from #896, plus whatever is new since then.

environment_python_3_9.yml
mfixstsci: Looks like I missed this, I'll get on it and merge.

Issue #900, opened 2022-03-11
Readnoise Monitor not picking up new data


bsunnquist:
The Readnoise Monitor isn't running on new NIRCam darks, though the Bias Monitor is and they use very similar methods.

bsunnquist:
Update - it appear to be crashing due to a header keyword:
Failed validating 'enum' in schema:
    OrderedDict([('title', 'Quality of pointing information from EngDB'),
                 ('type', 'string'),
                 ('enum',
                  ['CALCULATED',
                   'CALCULATED_COARSE_TR_202107',
                   'CALCULATED_FULL',
                   'CALCULATED_FULLVA',
                   'CALCULATED_GSCMD',
                   'CALCULATED_GSCMD_V3PAGS',
                   'CALCULATED_ORIGINAL',
                   'PLANNED',
                   'CALCULATED_TR_202105',
                   'CALCULATED_TR_202105_VA',
                   'CALCULATED_TRACK_TR_202107']),
                 ('fits_keyword', 'ENGQLPTG'),
                 ('blend_table', True)])

On instance:
    'CALCULATED_COARSE_TR_202111'


I'm unable to reproduce the error locally and the pipeline runs fine (using and jwst 0.17.1), so this must be something new in the production jwst version 1.3.3

bsunnquist:
Yep that was it - thanks! I was able to reproduce the error on jwst1.3.3, but then it went away on jwst1.4.3.

Do you know how to get our ops environment to use jwst1.4.3 (i.e. the one jwql.stsci.edu uses)?

bhilbert4:
Ha. The pipeline, for versions >1.3.3, is supported only on python 3.8 and higher. So we need a whole new environment on the JWQL servers.

bsunnquist:
That's too bad. I do have a temporary workaround if needed - if I set `model.meta.visit.engdb_pointing_quality = 'CALCULATED'` before saving the model it runs fine (for some reason just removing that ENGQLPTG keyword from the header before running the pipeline doesn't work)

mengesser:
I don't see it noted anywhere so I thought I might add that the Dark monitor has also been failing on the same timeline for what I presume is the same reason.

bsunnquist:
Thanks @mengesser ! Yep, both the dark monitor and bad pixel monitor are also crashing due to this same issue (confirmed via their recent logs).

[...]

PR #922, opened 2022-04-08
Make Readnoise Monitor less memory-intensive 

bsunnquist:
This PR does the following
- switch to perform the readnoise calculation in smaller slices to reduce memory pressure
- makes the plot limits work better with the ranges in values we're seeing (applied same change to bias monitor too, while I was at it)
- less stringent minimum groups required to calculate readnoise, to maximize the number of data points
- increased the lookback time, so these changes would be applied to all of the new darks we received

The first point is the main one - it addresses the memory issues noted in  https://github.com/spacetelescope/jwql/issues/905 , and (after the jwst pipeline is switched to jwst1.4.3) should also solve https://github.com/spacetelescope/jwql/issues/900.

### DIFF
--- a/jwql/instrument_monitors/common_monitors/readnoise_monitor.py
+++ b/jwql/instrument_monitors/common_monitors/readnoise_monitor.py
@@ -548,13 +548,6 @@ class Readnoise():
             siaf = Siaf(self.instrument)
             possible_apertures = list(siaf.apertures)
 
-            # Get the minimum groups needed to calculate readnoise for this instrument. MIRI
-            # removes the first five and last group before calculating the readnoise, so needs extra.
-            if instrument == 'miri':
-                min_groups = 8
-            else:
-                min_groups = 2
-
             for aperture in possible_apertures:
 
                 logging.info('\nWorking on aperture {} in {}'.format(aperture, instrument))
@@ -611,7 +604,13 @@ class Readnoise():
                         else:
                             num_groups = fits.getheader(uncal_filename)['NGROUPS']
                             num_ints = fits.getheader(uncal_filename)['NINTS']
-                            if (num_groups >= min_groups) & (num_ints >= 2):  # skip processing if the file doesnt have enough groups/ints to calculate the readnoise
+                            if instrument == 'miri':
+                                total_cds_frames = int((num_groups-6)/2) * num_ints
+                            else:
+                                total_cds_frames = int(num_groups/2) * num_ints
+                            # Skip processing if the file doesnt have enough groups/ints to calculate the readnoise.
+                            # MIRI needs extra since they omit the first five and last group before calculating the readnoise.
+                            if total_cds_frames >= 10:
                                 shutil.copy(uncal_filename, self.data_dir)
                                 logging.info('\tCopied {} to {}'.format(uncal_filename, output_filename))
                                 set_permissions(output_filename)
--- a/jwql/instrument_monitors/common_monitors/readnoise_monitor.py
+++ b/jwql/instrument_monitors/common_monitors/readnoise_monitor.py
@@ -351,7 +351,7 @@ class Readnoise():
         readnoise = np.zeros((num_y, num_x))
         for idx in cols_idx:
             # Create a stack of correlated double sampling (CDS) images using input
-            # ramp data, combining multiple integrations if necessary. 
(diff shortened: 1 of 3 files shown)


bsunnquist:
@bhilbert4 this is ready for review whenever. It should solve all of the ongoing readnoise monitor issues (assuming also the jwst pipeline is updated to 1.4.3)

jwql/instrument_monitors/common_monitors/readnoise_monitor.py
```diff
@@ -342,28 +342,34 @@ def make_readnoise_image(self, data):
...
-            else:
-                # Omit the last group if the number of groups is odd
-                cds = data[integration, 1::2, :, :] - data[integration, ::2, :, :][:-1]
 
-            if integration == 0:
-                cds_stack = cds
-            else:
-                cds_stack = np.concatenate((cds_stack, cds), axis=0)
+        # Calculate the readnoise in slices to avoid memory issues on large files.
+        slice_width = 20
```
mfixstsci, line 349: Nice addition here, I think this should help with the memory pressure we have been experiencing.

jwql/instrument_monitors/common_monitors/readnoise_monitor.py
```diff
@@ -542,16 +548,23 @@ def run(self):
             siaf = Siaf(self.instrument)
             possible_apertures = list(siaf.apertures)
 
+            # Get the minimum groups needed to calculate readnoise for this instrument. MIRI
+            # removes the first five and last group before calculating the readnoise, so needs extra.
+            if instrument == 'miri':
```
mfixstsci: Were all minimum groups set to a higher value previously?

jwql/website/apps/jwql/monitor_pages/monitor_bias_bokeh.py
```diff
@@ -181,5 +181,11 @@ def update_mean_bias_figures(self):
                 if len(bias_vals) != 0:
                     self.refs['mean_bias_xr_amp{}_{}'.format(amp, kind)].start = expstarts.min() - timedelta(days=3)
                     self.refs['mean_bias_xr_amp{}_{}'.format(amp, kind)].end = expstarts.max() + timedelta(days=3)
-                    self.refs['mean_bias_yr_amp{}_{}'.format(amp, kind)].start = min(x for x in bias_vals if x is not None) - 20
-                    self.refs['mean_bias_yr_amp{}_{}'.format(amp, kind)].end = max(x for x in bias_vals if x is not None) + 20
+                    min_val, max_val = min(x for x in bias_vals if x is not None), max(x for x in bias_vals if x is not None)
```
mfixstsci, line 184: Thanks for tackling the same issue with the bias monitor as well.

mfixstsci:
@bsunnquist thank you for the submission, I went ahead and had a few questions/comments.

bsunnquist:
thanks @mfixstsci , did your review post? I don't see any questions, just that one comment so far

mfixstsci (approved):

mfixstsci:
@bsunnquist I guess I didn't hit the big green button 😓 let me know if you see it now.

jwql/instrument_monitors/common_monitors/readnoise_monitor.py
bsunnquist: thanks yeah it used a lot less when I was testing locally

jwql/instrument_monitors/common_monitors/readnoise_monitor.py
bsunnquist: yes, and it missed the cases where there were lots of ints but low groups, which should still be able to give a good cds readnoise measure. Though this comment reminded me of something, so I made another tweak to this logic in a new push.

bsunnquist:
thanks @mfixstsci I just made one more minor tweak to the minimum group/int logic, but let me know if you see any issues. This latest strategy will make sure images with lots of ints/low groups will still get processed, but also ensure the total number of these cds frames is still large enough to give meaningful measurements

bsunnquist:
and one more tweak, to push back the lookback time a little more to make sure it picks up the earliest dark data

mfixstsci (approved):
Looks good to me!

PR #923, opened 2022-04-08
Updating JWST Pipeline Version Number

### LINKED ISSUE #903: jwst pipeline > v1.3.3 only supported under python >=3.8
Our test, dev, and production servers are running environments with python 3.7. But the jwst pipeline version 1.3.3 was the last version that is supported under python 3.7. We need to update the server environments to use python something more recent than 3.7. 

Maybe jump to 3.9? Although it looks like we don't have an environment file for 3.9 at the moment.

@mfixstsci what do you think? Move up to 3.8 as a short term fix while we get an environment file for 3.9 ready, and then move to 3.9?

mfixstsci:
A number of issues have turned up because the pipeline version wasn't updated.

- Removed dependencies on `crds` and `stdatamodel`. These are resolved when `jwst` is installed.
- Increased `jwst==1.3.3 ---> 1.4.3` this was done in the `requirement.txt` but neglected in the environment files.

### DIFF
--- a/environment_python_3_8.yml
+++ b/environment_python_3_8.yml
@@ -56,11 +56,9 @@ dependencies:
     - asdf==2.11.0
     - astroquery==0.4.5
     - bandit==1.7.4
-    - crds==11.10.0
-    - jwst==1.3.3
+    - jwst==1.4.3
     - pysiaf==0.11.0
     - pysqlite3==0.4.3
-    - stdatamodels==0.2.4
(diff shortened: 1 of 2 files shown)


bhilbert4 (approved):
Hooray for reducing dependencies! This looks good to me, and all the tests have passed.

mfixstsci:
@bhilbert4 have to include @bsunnquist 's new changes and let the tests run then we can merge.

PR #927, opened 2022-04-14
Hover over proposal thumbnails

### LINKED ISSUE #562: Change display of thumbnail when hovering over/selecting it
It would be a nice feature to have if a thumbnail would change appearance somehow when hovering over it to select it, when viewing an instrument archive page (e.g. http://127.0.0.1:8000/NIRCam/archive/).  For example, the thumbnail could appear in true color, or become slightly larger in size.

shanosborne:
The image opacity changes when you hover over a thumbnail on the `<instr>/archive/` pages. The image gets slightly lighter, but not so light that the white text with the proposal ID and file information becomes difficult to read.

Closes #562

### DIFF
--- a/jwql/website/apps/jwql/static/css/jwql.css
+++ b/jwql/website/apps/jwql/static/css/jwql.css
@@ -373,6 +373,15 @@ li:hover .nav-link, .navbar-brand:hover {
     z-index: 1;
 }
 
+.proposal:hover {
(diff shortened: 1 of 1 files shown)


mfixstsci (approved):
LGTM. Just tested this on the test server, looked at the color online it matched to what I was seeing on test.

PR #932, opened 2022-04-20
removed unused auth keys, fixed lint errors in corresponding files

### LINKED ISSUE #899: client_id and client_secret no longer needed in config.json
The `client_id` and `client_secret` keys were once used when JWQL implemented an `auth.mast` authentication layer, but this is no longer supported.  These keys in `config.json`, along with the supporting documentation and reference in `utils.utils` and `tests.test_utils` may be removed.

BradleySappington:
Removed unused Client keys and client secret in config.json and utils/test_utils
Updated config.json on dev and ops server.  Test server down for maintenance so still need to update there.

### DIFF
--- a/jwql/example_config.json
+++ b/jwql/example_config.json
@@ -1,8 +1,6 @@
 {
     "admin_account" : "",
     "auth_mast" : "",
-    "client_id" : "",
-    "client_secret" : "",
     "connection_string" : "",
     "database" : {
         "engine" : "",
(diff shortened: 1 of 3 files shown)


BradleySappington:
Ready for review

bhilbert4:
Just a quick note as I went to test something unrelated on the test server today. Removing `client_id` and `client_secret` from the config file on the server actually causes JWQL to fail, because they are still listed as being required in the config file. They'll be required until this PR is merged. So I guess we should keep those entries on the servers' versions of the files until after this PR is merged (and until it is released, before changing the ops server file).

mfixstsci (approved):
These changes look good. @bhilbert4 before approving I remember you saying something will need to be done on the server side?

bhilbert4:
Yeah. I just want to make sure that its ok from an operational standpoint that these files are deleted each time. I'm just being paranoid...

PR #933, opened 2022-04-20
Delete unnecessary files created by dark and bad pixel monitors

### LINKED ISSUE #934: Remove unnecessary fits file outputs from dark/bad pix monitors
Similar to #930, delete any unnecessary fits files created by the dark and bad pixel monitors. This will save disk space on the servers.

bhilbert4:
This PR adds code to the dark and bad pixel monitors to delete unnecessary files (the original input files, pipeline products) after the monitor has completed. This will save disk space on the servers.

A further way to save disk space would be to have the monitor create a jpg or png of the main output file (e.g. the mean dark rate image or the file used as the background image in the plot of bad pixels) at runtime. Then that fits file could also be deleted and the jpg/png could be used by the bokeh commands used to create the plots/figures. Currently the bokeh code relies on the fits files being present so that it can open them and display the 2D array as an image.

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -1,3 +1,12 @@
+Unreleased
+==========
+
+Web Application
+~~~~~~~~~~~~~~~
+
+- Dark and Bad Pixel Monitors delete unnecessary outputs to save disk space (#933)
+
+
 1.1.1 (2022-04-05)
 ==================
 
--- a/jwql/instrument_monitors/common_monitors/bad_pixel_monitor.py
+++ b/jwql/instrument_monitors/common_monitors/bad_pixel_monitor.py
@@ -83,6 +83,7 @@ Templates to use: ``FGS_INTFLAT``, ``NIS_LAMP``, ``NRS_LAMP``,
 
 from copy import deepcopy
 import datetime
+from glob import glob
 import logging
 import os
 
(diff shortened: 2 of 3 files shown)


bhilbert4:
Hmm. This won't work for the bad pixel monitor. When the bad pixel monitor webpage is loaded, it retrieves one of the slope images from the output directory and uses that image as a background when showing the location of new bad pixels. Maybe we should add a generic slope image file for each detector and have bokeh use that. Then it would be ok to delete the slope images.

bhilbert4:
Nevermind. The bad pixel monitor actually uses the file in the filesystem to grab the data array and use that as the background for the bad pixel plot. So it is safe to delete the output files from the bad pixel monitor.

bhilbert4:
This is ready for testing on the dev server.

bhilbert4:
Trial run on the test server was a success. I think it's safe to merge this one.

mfixstsci (approved):
Hey @bhilbert4 it looks good to me. I see in the code you `glob.glob('*.fits')` do we want to delete every fits file in the data directory once we are done?

bhilbert4:
Yes. The directory being searched by that `glob` statement contains the pipeline products for only a single exposure. So there should be only 3 or 4 files, and none of them are needed.

PR #936, opened 2022-04-21
Remove unnecessary output files from bias/readnoise monitors

### LINKED ISSUE #930: Remove unnecessary fits files from readnoise/bias monitors
To save disk space, remove any fits files after the bias/readnoise monitor completes. At that point, the only thing the web app needs is the database entries and pngs, so the fits files are not necessary and just take up space.

Also, if the pipeline fails, the uncal files are left around in our directories, so delete those too in that scenario.

bsunnquist:
This PR addresses https://github.com/spacetelescope/jwql/issues/930, to remove fits files from the bias/readnoise monitors that are no longer needed once processing completes, to save memory space on our servers. It also removes the uncal files if pipeline processing fails so they aren't just hanging around taking up space.

The web app only needs the output pngs and database entries to construct the webpages, which is why it's fine to remove these.

### DIFF
--- a/jwql/instrument_monitors/common_monitors/bias_monitor.py
+++ b/jwql/instrument_monitors/common_monitors/bias_monitor.py
@@ -382,6 +382,7 @@ class Bias():
                 set_permissions(processed_file)
             except:
                 logging.info('\tPipeline processing failed for {}'.format(filename))
+                os.remove(filename)
                 continue
 
             # Find amplifier boundaries so per-amp statistics can be calculated
--- a/jwql/instrument_monitors/common_monitors/bias_monitor.py
+++ b/jwql/instrument_monitors/common_monitors/bias_monitor.py
@@ -430,6 +430,10 @@ class Bias():
             self.stats_table.__table__.insert().execute(bias_db_entry)
             logging.info('\tNew entry added to bias database table: {}'.format(bias_db_entry))
 
+            # Remove the raw and calibrated files to save memory space
+            os.remove(filename)
+            os.remove(processed_file)
+
     @log_fail
     @log_info
     def run(self):
--- a/jwql/instrument_monitors/common_monitors/readnoise_monitor.py
+++ b/jwql/instrument_monitors/common_monitors/readnoise_monitor.py
@@ -423,6 +423,7 @@ class Readnoise():
                 set_permissions(processed_file)
             except:
                 logging.info('\tPipeline processing failed for {}'.format(filename))
+                os.remove(filename)
                 continue
 
             # Find amplifier boundaries so per-amp statistics can be calculated
--- a/jwql/instrument_monitors/common_monitors/readnoise_monitor.py
+++ b/jwql/instrument_monitors/common_monitors/readnoise_monitor.py
@@ -437,8 +437,8 @@ class Readnoise():
             # Make the readnoise image
             readnoise_outfile = os.path.join(self.data_dir, os.path.basename(processed_file.replace('.fits', '_readnoise.fits')))
             readnoise = self.make_readnoise_image(cal_data)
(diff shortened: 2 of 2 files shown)


mfixstsci:
@bsunnquist thanks for submitting this! I was curious about the removing the uncal files if the pipeline fails. Does that mean future monitor runs will pick up these files and attempt to process them again?

bsunnquist:
@mfixstsci Yeah that's correct, as long as the filename isn't in the monitor database (i.e. completed processing) it will keep trying each day. I figured it was best to delete the uncal's between runs though so they aren't hanging around, because like we saw recently, if the pipeline fails this can result in a lot of large data files taking up space on our servers

jwql/instrument_monitors/common_monitors/readnoise_monitor.py
```diff
@@ -437,8 +438,8 @@ def process(self, file_list):
             # Make the readnoise image
             readnoise_outfile = os.path.join(self.data_dir, os.path.basename(processed_file.replace('.fits', '_readnoise.fits')))
             readnoise = self.make_readnoise_image(cal_data)
-            fits.writeto(readnoise_outfile, readnoise, overwrite=True)
-            logging.info('\tReadnoise image saved to {}'.format(readnoise_outfile))
+            #fits.writeto(readnoise_outfile, readnoise, overwrite=True)
```
mfixstsci, line 441: This was because these files were taking up too much disk space? Is there a way we could make a `utils.py` function? A function that could get the readnoise from a specific time period and make this file locally with the API or something?

jwql/instrument_monitors/common_monitors/bias_monitor.py
```diff
@@ -430,6 +431,10 @@ def process(self, file_list):
             self.stats_table.__table__.insert().execute(bias_db_entry)
             logging.info('\tNew entry added to bias database table: {}'.format(bias_db_entry))
 
+            # Remove the raw and calibrated files to save memory space
+            os.remove(filename)
```
mfixstsci, line 435: I think this will help with our memory issue, is the `processed_file` the file produced at this step of the pipeline? Are their multiple files produced?

bhilbert4:
Are there any other pipeline products that may need to be removed? I'm thinking about e.g. trapsfilled, if the pipeline is run far enough.

bsunnquist:
I don't think so - these just run dq/superbias/refpix

bhilbert4:
This PR looks good to me. @mfixstsci do you agree?

mfixstsci (approved):

jwql/instrument_monitors/common_monitors/bias_monitor.py
mfixstsci: Thanks for the answer @bhilbert4 and @bsunnquist I will resolve this since you answered it above.

mfixstsci:
@bhilbert4 it does look good, I forgot to fully submit my review 😓 I wanted to ask @bsunnquist a question about the `readnoise` file we are no longer generating at the end. Do we want to make some sort of `utils.py` function to allow users to make the file using our API?

jwql/instrument_monitors/common_monitors/readnoise_monitor.py
bsunnquist: Yeah that would be possible, the readnoise calculation is ultimately just a few lines of code - we also have this same function as part of our jwst reference files tool which let's anyone make a readnoise image with a given file set if they want to further inspect things.

mfixstsci: Just wanted to make sure anyone who needs these images will still be able to produce them!

PR #937, opened 2022-04-21
Remove savefig overwrite call to avoid future deprecation error

### LINKED ISSUE #904: matplotlib savefig overwrite keyword is deprecated
While testing the readnoise monitor this morning, I saw a warning that will become an error down the road:

03/23/2022 12:35:43 PM INFO:    Creating png of readnoise difference image
03/23/2022 12:35:43 PM WARNING: savefig() got unexpected keyword argument "overwrite" which is no longer supported as of 3.3 and will become an error two minor releases later

I've searched matplotlib for details, but haven't found anything yet hinting at whether there is another keyword that replaces overwrite.

Once we find a solution for this, we should make sure that `overwrite` is removed in all JWQL calls to plt.savefig()

bsunnquist:
This PR addresses https://github.com/spacetelescope/jwql/issues/904 . It simply removes the overwrite call when saving figures in the readnoise monitor to avoid a future deprecation error. This is the only place I see this argument in our code base.

### DIFF
--- a/jwql/instrument_monitors/common_monitors/readnoise_monitor.py
+++ b/jwql/instrument_monitors/common_monitors/readnoise_monitor.py
@@ -265,7 +265,7 @@ class Readnoise():
         plt.title('{}'.format(outname))
 
         # Save the figure
-        plt.savefig(output_filename, bbox_inches='tight', dpi=200, overwrite=True)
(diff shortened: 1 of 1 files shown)


mfixstsci (approved):
Thanks for submitting this @bsunnquist LGTM

Issue #802, opened 2021-09-24
Put Program IDs in order on the Archived Images page


bhilbert4:
It would be helpful for users if the program IDs were put in order on the Archived Images page for each instrument. At the moment the order of the IDs appear to be random, which makes it more difficult to find a particular program. 

Bonus points:
It might be nice to have different options for how to arrange the proposals. Once we are in flight and people are coming to the website to (e.g.) check the data that have been taken in the last few days, it might be easier to order the proposals by the date of execution.

bhilbert4:
Somehow I missed the big red button on the right side of the page that will allow the programs to be placed in ascending or descending order. The default ordering, when the page is first loaded, may be neither of these. (Is it random?) I just tried to double check, but the development server doesn't want to load any of the archived images pages for me at the moment.

bhilbert4:
On a particular instrument's Archived Images page, the button to sort the proposals into ascending or descending order works. But when the page is first loaded, the order is neither of these. Would be nice to have it load with the proposals in either ascending or descending order.

Similarly, on the page for a given instrument/program, when the page is loaded the order of the individual exposures is unclear. It's not in order of increasing/decreasing observation number, visit number, or detector name. It would be nice if upon loading the images were ordered by name.

bhilbert4:
For the Archived Images page, a potential solution would be within get_rpoposal_info() in data_containers.py, to sort the `proposals`, and apply the same sorting to the `thumbnail_paths` and `num_files` variables before populating `proposal_info`.

bhilbert4:
For the images page for a given proposal, it looks like the default sorting is most recent first/oldest last, although within this the detector names (for nircam) seem to be random. So perhaps an improvement for this would be: 
1) rename "default" to "most recent" in the drop down menu, so it's clear
2) sort each file root by detector?

See towards the bottom of thumbnails_ajax() in data_containers.py for the location to sort things.

One place that is convenient for sorting by detector is just after the 
`available_files = [item for item in filenames if rootname in item]` line.

bhilbert4:
Programs now default to being shown in descending order, with Ascending and Recent as other options in the drop down.
With the ability now to filter by detector on a program's page, the order of the detectors within each exposure is less important. Closing this, and if we decide later that we want to sort by detector, we can create a new issue.

PR #942, opened 2022-04-28
Speed up MAST queries

### LINKED ISSUE #941: Slow/incomplete results for NIRCam from MAST API search
Recent timing tests on the test server showed that when using the mast_query_filenames_by_instrument() function, it took upwards of 3 minutes to get a response for NIRCam, while it took under 1 second for the other instruments. I also saw that the NIRCam query was returning a list of exactly 50,000 filenames, which seemed suspicious.

I submitted a ticket to the Archive help desk about this. The here is the response:

```
"The behavior you're seeing is expected. There is a default "pagesize" set in "Mast.service_request_async"
that limits the returned results to 50,000 files. Because there are more than 50,000 NIRCAM files in the JWST
Instrument Keyword database (51,796 files at the time I'm writing this message), your query is being slowed
down significantly and you're only getting part of the expected result.

Your best option will likely be to break your single query up into multiple queries using the "filters" parameter
to search by e.g., program ID or date."
```

So, it seems like we will have to rethink how to query the archive, at least within mast_query_filenames_by_instrument(). Do we really need all of the filenames? Perhaps we can filter by suffix, and just get the e.g. uncal files? Or maybe we can query to get a list of program IDs, and then query for each?

A good place to start may be to check where we call this function and see what we really end up using from the results.

bhilbert4:
Resolves #941 

This PR updates JWQL's MAST queries to avoid issues caused by query results that are too large.  According to the Archive Branch, query results are limited to 50,000 rows. Any query result that is larger is cropped down to 50,000 rows and results in a much slower response. When querying for all files for a particular instrument, as we are set up to do in multiple functions in JWQL, we are already beyond this 50,000 row limit for NIRCam. So a NIRCam search was taking upwards of 3 minutes, while a search with one of the other instruments was taking under 1 second.

The easiest way around this I think is to search per proposal. So this PR re-works our MAST queries to run one proposal at a time.

NOTE: MAST queries can be filtered using essentially any of the JWST keywords. Using proposal ID seemed the easiest solution to me, but there may be others. Here is the list of possible filters: https://mast.stsci.edu/api/v0/_jwst_inst_keywd.html


jwql/website/apps/jwql/data_containers.py
```diff
@@ -208,13 +208,10 @@ def get_all_proposals():
...
         filesystem
     """
-
-    proposals = glob.glob(os.path.join(FILESYSTEM_DIR, 'public', '*'))
-    proposals.extend(glob.glob(os.path.join(FILESYSTEM_DIR, 'proprietary', '*')))
-    proposals = sorted(list(set(proposals)))
-    proposals = [proposal.split('jw')[-1] for proposal in proposals]
-    proposals = [proposal for proposal in proposals if len(proposal) == 5]
-
+    proprietary_proposals = os.listdir(os.path.join(FILESYSTEM_DIR, 'proprietary'))
```
bhilbert4, line 212: In my limited speed tests, os.listdir was faster than glob

bhilbert4:
- [x] get_expstart is quite expensive. ~0.1 second for each call, and it's called on all files in a proposal. Better to include expstart column in the mast query for the filenames.

bhilbert4:
data_containers.get_thumbnails_all_instruments() needs to be revised in a similar way. That is a more complex query, so might take some more thought. Maybe looping over proposal numbers would work. I've split that off into a separate issue, #943

bhilbert4:
At this point, it looks like the web app is faster in some cases. When loading a page for a particular proposal, it's quite fast for proposals that have under roughly 150 files. But for proposals with more, the loading time is still very long, if it loads at all.

I've run thumbnails_ajax() manually on the server, and even for proposals with 1800 files, the run time is under 1 second. So the slowdown must be coming in somewhere else. The javascript functions? The slowdown is strange. It's going from a reasonably quick load to never loading the page as the number of files increases. There doesn't seem to be a middle ground where it takes e.g. 60 seconds. @mfixstsci any thoughts? 

I also manually ran views.archived_proposals_ajax() on the server, and that seemed reasonably quick as well. Loading time for each instrument's Archive page now takes ~20-25 seconds. The bottleneck here is querying MAST once for each proposal. I'm not sure how else to break up the MAST queries in order to assure we don't hit the 50,000 image limit and still cover all the data. Perhaps query over some time period, but then we'd have to organize the results by proposal.

bhilbert4:
While looking at lists of thumbnail files, I was able to find and resolve #801. For a given instrument, the entries in the archive page now report the same number of files as on the thumbnail pages (from my spot checks). This means that #821 can be closed without merging.

bhilbert4:
The only remaining issue at the moment is that the thumbnail pages for proposals containing more than ~300 files do not load due to what looks to be a server timeout error. (net::ERR_CONTENT_LENGTH_MISMATCH 200). From my reading about it, it sounds like this may be because a json file is too large, or there is some delay during the reading in of that file. See [here](https://nodejs.org/dist/latest-v10.x/docs/api/http.html#http_server_keepalivetimeout) and [here](https://www.geeksforgeeks.org/node-js-http-server-keepalivetimeout-property/). @mfixstsci do you know anything about this? 

It sounds like a possible fix would be an adjustment of the keepAliveTimeout value on the server. This may also be an argument in favor of tweaking the way we display proposals that have lots of files. In some (rare) cases for NIRCam there are proposals with thousands of files. It might be easier for users to navigate in cases like these if we organize the thumbnails and display only those for one Observation at a time.

bhilbert4:
The previous commit addresses #802. Program IDs are now in descending order when the Archive page is loaded for a particular instrument. The thumbnails for each program's page are sorted into reverse chronological order. Would be nice to sort the exposures into reverse chronological order, and within each exposure, sort by detector, if possible. That might take some work in data_containers.thumbnails_ajax()

environment_python_3_7_with_bokeh2.1.1.yml
```diff
@@ -0,0 +1,40 @@
+channels:
```
mfixstsci: @bhilbert4 did you mean to add this file?

jwql/jwql_monitors/monitor_mast.py
```diff
@@ -143,8 +143,13 @@ def instrument_keywords(instrument, caom=False):
         A DataFrame of the keywords
     """
     # Retrieve one dataset to get header keywords
+    if not caom:
```
mfixstsci, line 146: What does this logic do? I am just curious, what is special about this program?

jwql/website/apps/jwql/static/js/jwql.js
```diff
@@ -520,10 +520,19 @@ function update_thumbnails_page(inst, proposal, base_url) {
         url: base_url + '/ajax/' + inst + '/archive/' + proposal + '/',
         success: function(data){
             // Perform various updates to divs
+            let t1 = performance.now();
```
bhilbert4: Oops. These timing statements can be deleted. I'll do that.

mfixstsci: Are we logging this information now or was this for testing the loading speeds?

jwql/website/apps/jwql/views.py
```diff
@@ -271,36 +271,61 @@ def archived_proposals_ajax(request, inst):
...
-        try:
-            relative_filepath = filesystem_path(filename, check_existence=False)
-            full_filepath = os.path.join(filesystem, 'proprietary', relative_filepath)
-            filenames.append(full_filepath)
-        except ValueError:
-            print('Unable to determine filepath for {}'.format(filename))
-
-    # Gather information about the proposals for the given instrument
-    proposal_info = get_proposal_info(filenames)
+    for proposal in all_proposals:
```
mfixstsci, line 282: This is just where we are breaking the views into proposals based on instrument and propid? This looks more efficient 👍

mfixstsci (approved):
Looks good, thanks for submitting @bhilbert4 !

jwql/jwql_monitors/monitor_mast.py
bhilbert4: This block queries two different MAST services. Either Caom.Filtered or Jwst.Instrument.Filtered.<instrument>. The two services seem to use two different keywords for the program ID, so this just makes sure the correct keyword is used.

jwql/website/apps/jwql/views.py
bhilbert4: Yes, I like this much better. This pushes more of the filtering work onto MAST, which is faster than when we do it ourselves.

environment_python_3_7_with_bokeh2.1.1.yml
bhilbert4: Removed. Not sure how it got in there.

jwql/website/apps/jwql/static/js/jwql.js
bhilbert4: Leftover from testing. Removed. I am excited that I now know how to print information to the console from javascript functions though.

PR #946, opened 2022-05-02
fix ignore of original and stream suffixes for generate_preview_images

### LINKED ISSUE #944: Preview images/thumbnails for "stream" files
It looks like we are still creating preview images/thumbnails for "stream" files (e.g. jw01163148001_gs-fg_2022090163856_stream.fits). Right now the code is creating a preview image for every 100th integration in an exposure. But the stream files can have tens of thousands of integrations, so we're still ending up with a huge number of preview images for each. This is filling up directories and disk space, and greatly slows the preview image generator.

We should either completely turn off the preview image generation for these files, or perhaps just create them for a handful of extensions, if the FGS team decides they'd want to see them. @shanosborne any thoughts on that?

BradleySappington:
properly ignore "original" and "Stream" when getting filenames in generate_preview_images

### DIFF
--- a/jwql/jwql_monitors/generate_preview_images.py
+++ b/jwql/jwql_monitors/generate_preview_images.py
@@ -669,8 +669,8 @@ def process_program(program):
     filenames.extend(glob.glob(os.path.join(SETTINGS['filesystem'], 'proprietary', program, '*/*.fits')))
     filenames = list(set(filenames))
 
-    # Ignore "original" files
-    filenames = [filename for filename in filenames if os.path.splitext(filename.split('_')[-1]) not in IGNORED_SUFFIXES]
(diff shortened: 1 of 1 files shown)


bhilbert4 (approved):
Good catch! This looks good to me.

mfixstsci (approved):
Hey @BradleySappington just went ahead and tested this real quick and it works just fine. Thank submitting this.

mfixstsci:
Just merged #942 which includes some fixes for the tests. Assuming the tests pass, this should be good to go!

mfixstsci:
@BradleySappington and @bhilbert4 looks like the changes solved our time out issue 🎉 @BradleySappington is this ready to go?

PR #948, opened 2022-05-06
Split program page by obsnum

bhilbert4:
Testing has shown that programs with more than ~200-250 files will often fail to load due to a server timeout error.  In order to get around this, as well as make navigation easier for programs with large numbers of exposures, this PR makes changes to a program's page. With these changes, the page for a single program will be replaced with one page per Observation. 

I'm not sure if we should do this for all programs, or maybe just for NIRCam (which has many more files per program/observation). Other instruments typically have many fewer files per program.


mfixstsci:
@bhilbert4 I think this will eventually be useful for all instruments/programs depending on how large the programs get. I think we should think about a couple of things here:

* How does the UI flow from a users perspective, is clicking through each individual observation too much to ask? Particularly for smaller programs.
* How much work will it take to do this for certain programs compared to changing it for the whole application? Does it add too much complexity/variability to the application for just a couple of programs?

Maybe we add it to test and see what some of the instrument team users think?

bhilbert4:
Yeah, I think that makes sense. My initial thought is that only NIRCam will have file numbers large enough to trigger the server time out, but I could be wrong. But I also wonder whether teams will like having things organized a bit more granularly. e.g. if you're looking for the preview images for Observation X of your program, it might be nicer to have a page for Observation X rather than having to hover over images until you find Observation X. Might be worth having some people try it and see.

I think it would potentially be difficult to do this for some programs based on the number of files. I think doing it for one instrument and not another is feasible without too much trouble. 

Another thing I just realized is that even with this change in place, there are some NIRCam programs that can potentially have enough files in a given Observation that they'd trigger the server timeout. I think this would be true of only the very large programs (e.g. 10 detectors x 3 subpixel dithers x 7 mosaic tiles = 210 files would be enough. Or fewer dithers but with more filters would also be a problem.) I don't think there are very many programs which will hit this data volume, but some will. I wonder if in concert with breaking things up by Observation, we ask ITSD if they can increase the threshold for the server timeout?

bhilbert4:
Slowly getting closer. At the moment, the observation-specific pages load and contain the correct data.
Left to do:

- [ ] Add a drop down menu towards the top of the observation-specific page that contains links to other observations for the same program
- [ ] (For nircam only? For everyone?) set the URLs such that jwql.stsci.edu/<inst>/archive/  redirects to jwql.stsci.edu/<inst>/archive/obsXXX where XXX is the value of the first observation.

bhilbert4:
See #956 It looks like the function for generating the full list of obsnums is missing some entries somehow.

bhilbert4:
The page for NIRCam program 1473 doesn't appear quite right at the moment. There is no text in the dropdown for the list of observations.

bhilbert4:
@mfixstsci  I think this is now ready for review. I have it running on the dev server at the moment if you want to play with it. (confirmation-split-large-program-by-obsnum branch).

https://jwql-dev.stsci.edu/NIRSpec/archive/1122/obs023/

One small detail I have not been able to figure out: 
If the new dropdown at the upper right that lists all of the observation numbers, I haven't been able to get it to display the name of the current observation after it has been used, in the same way that the sort and filter dropdowns work. It's a small detail that's not terribly important, but if it's obvious to you how to add that in, it'd be nice to have it.

bhilbert4:
Hmm. Something weird is going on. It won't load an individual image page:
 https://jwql-dev.stsci.edu/NIRCam/jw01073003001_02101_00001_nrca4/

bhilbert4:
I see. This is because I removed the archive_thumb view. But it turns out that view_image.html points to that view in a link to the full proposal at the top of the page. So this will have to change to a link to the first obsnum of the proposal I guess?

bhilbert4:
Or we could create a new page for the full proposal that maybe has links to each obsnum? Would that be useful or too much effort for what it's worth?

jwql/website/apps/jwql/static/css/jwql.css
```diff
@@ -87,6 +87,8 @@
 .dropdown-menu {
     background-color: #2d353c;
     border-radius: 0px;
+    max-height:  400px;
```
mfixstsci, line 90: Thanks for implementing this.

jwql/website/apps/jwql/static/js/jwql.js
```diff
@@ -185,10 +185,42 @@ function determine_page_title(instrument, proposal) {
     var url = document.URL;
     var url_split = url.split('/');
     var url_title = url_split[url_split.length - 2];
+    //var url_end = url_split[url_split.length - 1];
```
mfixstsci: Should we remove this line?

jwql/website/apps/jwql/static/js/jwql.js
```diff
@@ -185,10 +185,42 @@ function determine_page_title(instrument, proposal) {
...
     if (url_title == 'archive') {
         final_title = 'Archived ' + instrument + ' Images: Proposal ' + proposal
     } else if (url_title == 'unlooked') {
         final_title = 'Unlooked ' + instrument + ' Images';
+    //} else if (isNaN(url_title) == false) {
```
mfixstsci: Remove more commented lines?

mfixstsci (approved):
Looks good to me @bhilbert4 just a few comments and maybe remove old lines out.

jwql/website/apps/jwql/static/js/jwql.js
bhilbert4: Done

jwql/website/apps/jwql/static/js/jwql.js
bhilbert4: Done

PR #949, opened 2022-05-06
Not found images replaced with default image

### LINKED ISSUE #925: Placeholder preview image
For programs that have been executed but have not yet had preview images created, the web app shows broken links on the program page. We should create a placeholder preview image that can be used until the real preview image is created.

BradleySappington:
utilize the existing imagenotfound.png

### DIFF
--- a/jwql/website/apps/jwql/static/css/jwql.css
+++ b/jwql/website/apps/jwql/static/css/jwql.css
@@ -273,7 +273,7 @@ display : inline;
   display: inline-block;
 }
 
-.instrument_select {m;
+.instrument_select {
   padding-top: 1rem;
   padding-bottom: 2rem;
   margin-right: 5rem;
--- a/jwql/website/apps/jwql/static/js/jwql.js
+++ b/jwql/website/apps/jwql/static/js/jwql.js
@@ -201,6 +201,18 @@ function determine_page_title(instrument, proposal) {
 };
 
 
+/** 
+ * If an image is not found, replace with temporary image sized to thumbnail
+ */
+function imageError(image) {
+    image.onerror = "";
+    image.src = "/static/img/imagenotfound.png";
+    /* Use thumbnail settings to keep it tidy */
+    image.className = "thumbnail";
+    return true;
+}
+
+
 /**
  * Perform a search of images and display the resulting thumbnails
  */
@@ -491,7 +503,7 @@ function update_thumbnail_array(data) {
             content = '<div class="thumbnail" instrument = ' +filename_dict.instrument + ' detector="' + filename_dict.detector + '" proposal="' + filename_dict.program_id + '" file_root="' + rootname + '", exp_start="' + file.expstart + '">';
             content += '<a href="/' + filename_dict.instrument + '/' + rootname + '/">';
         }
-        content += '<span class="helper"></span><img id="thumbnail' + i + '" onerror="this.src=/static/img/imagenotfound.png">';
+        content += '<span class="helper"></span><img id="thumbnail' + i + '" onerror="imageError(this);">';
         content += '<div class="thumbnail-color-fill" ></div>';
         content += '<div class="thumbnail-info">';
         content += 'Proposal: ' + filename_dict.program_id + '<br>';
@@ -561,3 +573,5 @@ function version_url(version_string) {
     a_line += '">JWQL v' + version_string + '</a>';
     return a_line;
 };
+
+
--- a/jwql/website/apps/jwql/static/js/jwql.js
+++ b/jwql/website/apps/jwql/static/js/jwql.js
@@ -204,11 +204,13 @@ function determine_page_title(instrument, proposal) {
 /** 
  * If an image is not found, replace with temporary image sized to thumbnail
  */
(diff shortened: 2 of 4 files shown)


BradleySappington:
ready for review

bhilbert4:
@BradleySappington there seems to be a misalignment between the default preview image and the information that pops up when you hover the mouse over one of them: https://jwql-test.stsci.edu/NIRCam/archive/01446/

BradleySappington:
@bhilbert4 Looks firefox/mozilla handle a margin setting for a class differently than safari.  I wont change the class in the thumbnail array anymore.  

@mfixstsci ready for review

bhilbert4:
I just checked this on the dev server using chrome, firefox, and safari, and the results look good on all of them!

jwql/website/apps/jwql/static/css/jwql.css
```diff
@@ -273,7 +273,7 @@ display : inline;
   display: inline-block;
 }
 
-.instrument_select {m;
+.instrument_select {
```
mfixstsci, line 276: What does `m` do in css? I've never seen that before

jwql/website/apps/jwql/static/js/jwql.js
```diff
@@ -201,6 +201,20 @@ function determine_page_title(instrument, proposal) {
 };
 
 
+/** 
+ * If an image is not found, replace with temporary image sized to thumbnail
+ */
+function imageError(image, makeThumbnail=false) {
```
mfixstsci, line 207: This looks good, always good to learn a little javascript along the way.

jwql/website/apps/jwql/static/js/jwql.js
```diff
@@ -491,7 +505,7 @@ function update_thumbnail_array(data) {
             content = '<div class="thumbnail" instrument = ' +filename_dict.instrument + ' detector="' + filename_dict.detector + '" proposal="' + filename_dict.program_id + '" file_root="' + rootname + '", exp_start="' + file.expstart + '">';
             content += '<a href="/' + filename_dict.instrument + '/' + rootname + '/">';
         }
-        content += '<span class="helper"></span><img id="thumbnail' + i + '" onerror="this.src=/static/img/imagenotfound.png">';
+        content += '<span class="helper"></span><img id="thumbnail' + i + '" onerror="imageError(this);">';
```
mfixstsci, line 508: What does wrapping the default image in a function accomplish here? Does it force every image to go through the same logic? Are we avoiding an incident where the image might be present but the default image is being displayed?

jwql/website/apps/jwql/templates/view_image.html
```diff
@@ -47,6 +47,7 @@ <h3>{{ file_root }}</h3>
     		    <span class="image_preview">
     		    	<img id="image_viewer"
                          src='{{ static("") }}preview_images/{{ file_root[:7] }}/{{ file_root }}_cal_integ0.jpg'
+                         onerror="imageError(this, true);"
```
mfixstsci, line 50: Is this a situation where you would display the image preview and the `onerror` hook will display the default if an image isn't found?

mfixstsci (approved):
Hey @BradleySappington this all looks great! I am not an expert with JS or CSS but the code is very readable and looks to solve the image preview issues we've been having. Thank you!

jwql/website/apps/jwql/static/css/jwql.css
BradleySappington: My linter was picking it up as an unknown, I assumed it was a bug and removed it.

jwql/website/apps/jwql/static/js/jwql.js
BradleySappington: I was going to initially always set the images to class "thumbnail", however for some reason Firefox/Chrome didn't like setting the actual thumbnails to the thumbnail class a second time, it affected their margins.

jwql/website/apps/jwql/static/js/jwql.js
BradleySappington: We currently have 2 locations in code that are going to fall back to this image on error, and they need to have this image sized accordingly.  Now if something changes with this image, or if we have another use for this image, we only need to change one location instead of hunting down code spread throughout the apps.

jwql/website/apps/jwql/templates/view_image.html
BradleySappington: Yup, pretty much. If there is an error with this image (like "Image Not Found")  `onerror` will initiate what it's told to do, so we are telling it to call `imageError` with the `true` flag to change to thumbnail class.  We set to thumbnail because we dont need a massive pagesize error image :)

jwql/website/apps/jwql/static/css/jwql.css
mfixstsci: That checks out because I wouldn't find anything online about it either!

jwql/website/apps/jwql/static/js/jwql.js
mfixstsci: Thank you!

PR #951, opened 2022-05-10
Fix call to pyvo for all instruments

bhilbert4:
This PR makes a small adjustment to the pyvo call that is used to generate a list of program IDs per instrument. Prior to this fix, the calls were failing for MIRI and FGS due to the % in the call.


bhilbert4:
Tested with all 5 instruments, and all 5 were successful.

bhilbert4:

mfixstsci:
Thanks for the submission @bhilbert4! The folks over in MAST were saying that using the SQL Wildcard operator (`%`) makes the DB freak out over fear of SQL injection. Also turns out that the direct string query, without the `%`, is a lot faster as well!

mfixstsci (approved):
LGTM

Issue #777, opened 2021-08-03
Display preview images of other FITS extentions such as the ERR array


bourque:
During the v1.0 demo with NIRSpec, there was a feature request to be able to view the uncertainty image (i.e. the ERR array) alongside the SCI array when viewing preview images.  Currently the preview image webpages support the viewing of different filetypes (e.g. uncal, rate, etc.), but not the individual FITS extensions.

PR #954, opened 2022-05-13
[WIP]: Produce interactive preview images

### LINKED ISSUE #952: Improve image checking
Feedback from the NIRSpec team at the WMO meeting last week:

The image checking interface (i.e. preview images used to flag anomalies) is not very useful. You can't zoom in or change the scaling in order to more closely examine the image to look for anomalies. This makes it hard to verify if there are features that should be flagged as anomalies.

How can we improve this? Perhaps allow the user to click on a preview image, which will then open a file and show the image using Bokeh, along with options for changing color palettes, scaling, zooming, etc? We could simply replace the preview images with that, but I wonder if that would be too slow.

bhilbert4:
This PR adds code to produce interactive preview images using Bokeh. This is in response to user feedback that our static preview images are not always useful for inspecting images and looking for anomalies. The Bokeh images produced here allow for:

- [x] Displaying data from any of the extensions of the given file
- [x] Displaying the frame corresponding to any requested integration/group number
- [x] Displaying the difference image from any 2 specified integration/group numbers
- [x] Zooming in and out 
- [x] Displaying in log or linear scale
- [x] Uses zscale for initial scaling of image
- [x] User can specify min and max scaling values

To do:
- [x] Create module to create the image given the input parameters. Return components, or save html, or display to screen
- [ ] Create a new form, and add buttons to allow the user to specify min and max scaling values, integration, group numbers, log/lin scaling, and file extension
- [ ] Speed tests: do we want to do away with static preview images completely, or is loading slow enough that we want this new image to be shown only after the user e.g. clicks on the static preview image?
- [ ] Create new html template for display
- [ ] Create new URLs, views
- [ ] Add new tests
- [ ] Documentation updates?

We should also figure out if this is all we need to address #777 (@shanosborne)
Resolves #952

### DIFF
--- a/jwql/utils/interactive_preview_image.py
+++ b/jwql/utils/interactive_preview_image.py
@@ -1,0 +1,284 @@
+#! /usr/bin/env python
+
+"""Module that can be used to read in a JWST observation and display one frame
+(from any extension) as an interactive Bokeh image.
+
+Authors
+-------
+
+    - Bryan Hilbert
+
+Use
+---
+
+    This module can be imported and called as such:
+
+    ::
+
+        from jwql.xxx.xxx import InteractivePreviewImg
+        file = 'jw01602001001_02102_00001_nrcb2_cal.fits'
+        im = InteractivePreviewImg(file, low_lim=None, high_lim=None, scaling='lin', contrast=0.4, extname='DQ')
+
+    Required Arguments:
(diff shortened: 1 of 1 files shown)


bhilbert4:
Superseded by #965. Closing.

PR #955, opened 2022-05-16
Lock python

### LINKED ISSUE #928: Some preview images contain plot axes
In some cases, preview images/thumbnail images appear to show plot axes, rather than a simple image of the data. @cracraft do you have any specific examples?

### LINKED ISSUE #950: Create lock files for preview image generation and monitors
Testing related to #928 has shown that preview images and thumbnails may not contain the correct image when we have multiple instances of generate_preview_images running at the same time. In this case, we can end up with them trying to open and save the same image at the same time. In order to solve this, we should institute a lock, such that only one instance can be running at a time. I think it also makes sense to have this set up for monitors as well. 

This can be done by having generate_preview_images() check for the existence of a lock file before beginning. If it exists, do not run. If it does not exist, create one, then run. Delete the lock file when done.

BradleySappington:
Create Wrapper to institute code protection when a module should only have one instance running at any given time.
This was created for and added into the generate_preview_images, generate_thumbnails, and all jwql monitors.

### DIFF
--- a/jwql/jwql_monitors/create_initial_preview_and_thumbnail_listfiles.py
+++ b/jwql/jwql_monitors/create_initial_preview_and_thumbnail_listfiles.py
@@ -9,11 +9,13 @@ Author:  B. Hilbert
 """
 from glob import glob
 import os
-import re
 
+from jwql.utils.protect_module import lock_module
 from jwql.utils.utils import get_config
 
 
+# Lock module makes create_files() protected code, ensures only one instance of module will run
+@lock_module
 def create_files():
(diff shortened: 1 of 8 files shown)


BradleySappington:
@mfixstsci - ready for review

mfixstsci (approved):
Hey @BradleySappington I don't have many comments since we addressed my questions in the code review we did on Friday. I will say, I think this will solve the issues we are having with the thumbnail/preview image issue and any conflict with the JWQL monitors.

Issue #901, opened 2022-03-11
Possibility to select ONE Instrument when multiple Instruments are associated to a single Program (PID)


anoriegacrespo:
**Is your feature request related to a problem? Please describe.**
Right now multiple instruments e.g. OTE, NIRCAM & FGS are linked to a single identification program

**Describe the solution you'd like**
Given the PID, it would be good to be able to select ONE of the instrument, e.g NIRCAM

**Describe alternatives you've considered**
The above feature could be selected as a multiple choice, or specified as a Keyword, e.g. 01059 NIRCAM
or 01059 FGS

bhilbert4:
In terms of the back end, it looks like the search will need to be modified here:
https://github.com/spacetelescope/jwql/blob/develop/jwql/website/apps/jwql/forms.py#L275

bhilbert4:
Coming back around to this update. I was just testing an update to the filename parser and was playing with the form submission box. It looks to me like all programs are going to be considered multi-instrument simply because the guider data that goes along with each program is included. So for example, program 1442 is all NIRCam, but with the addition of the guide star data, the program becomes multi-instrument and the search returns a "I can't load a proposal with more than one instrument" warning. It might be nice to have the instrument selector at least default to the prime instrument for a given proposal so that users won't have to continually select their instrument in order to avoid guide star data.

mengesser:
I guess the question for this issue now is what exactly we want the query to look like. Does it make sense to type a PID, and then create a new field for selecting from the available instruments? Or does it make more sense to allow the user to give the PID and instrument together, and then throw an error if that combination isn't possible? What is possible with django forms?

bhilbert4:
I think the former makes more sense. @mfixstsci what do you think?

mengesser:
We have a solution for this that Mees developed. I will put it in a PR this week (Right now I am on a loaner waiting for an OS upgrade)

mfixstsci:
Addressed in #1010 changes should be in. @anoriegacrespo  if you have any questions, feel free to ask us. If the program has multiple instrument, links will appear to select the route you are interested in. If your program only has one instrument, you will be routed directly.

PR #962, opened 2022-06-02
Add entry in filename_parser for segmented guider files

### LINKED ISSUE #961: Update filename parser to handle segment files of Guider data
It sounds like in addition to TSO data, FGS track and fine guide data can (optionally) be placed into segment files. At the moment the JWQL filename parser is unable to handle those filenames though. 

From Mark Kyprianou:
https://jwql.stsci.edu entered 1118 as the proposal :
 
ValueError at /
Provided file jw01118005001_gs-fg_2022150070312-seg002_uncal.fits does not follow JWST naming conventions.  See https://jwst-docs.stsci.edu/understanding-jwst-data-files/jwst-data-file-naming-conventions for further information.

Request Method:
POST
Request URL:
http://localhost/
Django Version:
3.1.7
Exception Type:
ValueError
Exception Value:
Provided file jw01118005001_gs-fg_2022150070312-seg002_uncal.fits does not follow JWST naming conventions.  See https://jwst-docs.stsci.edu/understanding-jwst-data-files/jwst-data-file-naming-conventions for further information.
Exception Location:
/home/svc_jwqladm_ops/jwql/jwql/jwql/utils/utils.py, line 432, in filename_parser
Python Executable:
/home/svc_jwqladm_ops/jwql/miniconda3/envs/jwql-3.7/bin/python
Python Version:
3.7.10

bhilbert4:
Resolves #961 
This PR adds code that allows the filename_parser to understand Guider segment files. These files are used in cases where the amount of Guider data is above some threshold, in which case it is split into multiple files.

### DIFF
--- a/jwql/utils/utils.py
+++ b/jwql/utils/utils.py
@@ -368,6 +368,18 @@ def filename_parser(filename):
         r"_gs-(?P<guider_mode>(id|acq1|acq2|track|fg))" \
         r"_((?P<date_time>\d{13})|(?P<guide_star_attempt_id>\d{1}))"
 
+    # Segment guider filenames
+    # e.g. "jw00729011001_gs-id_1_image_cal.fits" or
+    # "jw00799003001_gs-acq1_2019154181705_stream.fits"
+    guider_segment = \
+        r"jw" \
+        r"(?P<program_id>\d{5})" \
+        r"(?P<observation>\d{3})" \
+        r"(?P<visit>\d{3})" \
+        r"_gs-(?P<guider_mode>(id|acq1|acq2|track|fg))" \
+        r"_((?P<date_time>\d{13})|(?P<guide_star_attempt_id>\d{1}))" \
+        r"-seg(?P<segment>\d{3})"
+
     # Build list of filename types
     filename_types = [
         stage_1_and_2,
@@ -377,7 +389,8 @@ def filename_parser(filename):
         stage_3_target_id_epoch,
         stage_3_source_id_epoch,
         time_series,
-        guider]
+        guider,
+        guider_segment]
 
     filename_type_names = [
         'stage_1_and_2',
@@ -387,7 +400,8 @@ def filename_parser(filename):
         'stage_3_target_id_epoch',
         'stage_3_source_id_epoch',
         'time_series',
-        'guider'
+        'guider',
+        'guider_segment'
     ]
 
     # Try to parse the filename
--- a/jwql/utils/utils.py
+++ b/jwql/utils/utils.py
@@ -41,7 +41,8 @@ from jwql.utils.constants import FILE_AC_CAR_ID_LEN, FILE_AC_O_ID_LEN, FILE_ACT_
                                  FILE_DATETIME_LEN, FILE_EPOCH_LEN, FILE_GUIDESTAR_ATTMPT_LEN, \
                                  FILE_OBS_LEN, FILE_PARALLEL_SEQ_ID_LEN, FILE_PROG_ID_LEN, \
                                  FILE_SEG_LEN, FILE_SOURCE_ID_LEN, FILE_SUFFIX_TYPES, FILE_TARG_ID_LEN, \
-                                 FILE_VISIT_GRP_LEN, FILE_VISIT_LEN, JWST_INSTRUMENT_NAMES_SHORTHAND
+                                 FILE_VISIT_GRP_LEN, FILE_VISIT_LEN, FILETYPE_WO_STANDARD_SUFFIX, \
+                                 JWST_INSTRUMENT_NAMES_SHORTHAND
 
 __location__ = os.path.dirname(os.path.dirname(os.path.dirname(__file__)))
 
@@ -422,7 +423,7 @@ def filename_parser(filename):
     for filename_type, filename_type_name in zip(filename_types, filename_type_names):
 
         # If full filename, try using suffix, except for *msa.fits files
-        if not file_root_name and 'msa.fits' not in filename:
+        if not file_root_name and FILETYPE_WO_STANDARD_SUFFIX not in filename:
             filename_type += r"_(?P<suffix>{}).*".format('|'.join(FILE_SUFFIX_TYPES))
         # If not, make sure the provided regex matches the entire filename root
         else:
--- a/jwql/utils/utils.py
+++ b/jwql/utils/utils.py
@@ -37,7 +37,11 @@ import http
 import jsonschema
 
 from jwql.utils import permissions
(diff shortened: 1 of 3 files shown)


bhilbert4:
Local test shows the parser working correctly. Just need to test on a server using the search field on the home page to see if any other changes are needed in support of the filenames.

bhilbert4:
Looks like 1118 has lots of challenging files:
`Provided file jw01118008001_01_msa.fits does not follow JWST naming conventions`

bhilbert4:
I also added code to be able to parse filenames such as jw01118008001_01_msa.fits, which originally failed above. This is an MSA metadata file.

bhilbert4:
@mfixstsci and/or @BradleySappington I think this is ready for review. I tested it on the dev server by attempting to retrieve program 1118 using the "Find a JWST Proposal or File" field on the main page. Previously this program was causing the website to crash because of the filename parser error. Now it loads successfully....with a message saying that it cannot return data from proposals that use more than 1 instrument.

mfixstsci:
> @mfixstsci and/or @BradleySappington I think this is ready for review. I tested it on the dev server by attempting to retrieve program 1118 using the "Find a JWST Proposal or File" field on the main page. Previously this program was causing the website to crash because of the filename parser error. Now it loads successfully....with a message saying that it cannot return data from proposals that use more than 1 instrument.

Just tried it myself and it works as expected. Do we know of any multi-instrument programs we can test?

bhilbert4:
See #901. It looks at the moment like all programs are considered multi-instrument because they all use FGS for guide stars. That's a separate issue to this, and one we should try to complete before the next release. In the meantime, it might take some searching to find a single-instrument program that also has a segmented FGS file in it. And it won't be possible to find a program that has only nirspec and an MSA metadata file, since FGS will also be used in all of those.

jwql/utils/utils.py
```diff
@@ -301,6 +301,15 @@ def filename_parser(filename):
         r"_(?P<detector>((?!_)[\w])+)"\
         r"_(?P<ac_id>(o\d{3}|(c|a|r)\d{4}))"
 
+    # Stage 2 MSA metadata file. Created by APT and loaded in
+    # assign_wcs. e.g. "jw01118008001_01_msa.fits"
+    stage_2_msa = \
+        r"jw" \
+        r"(?P<program_id>\d{5})"\
```
BradleySappington: I know they occur throughout the entire code base, but it makes sense to bring it up now if its an easy enough fix.  We should avoid using "magic numbers" in our code if we don't have to.  "Constants" in python do not exist, so to let the programmer know that the "variable" should not be changed you can write it in ALL_CAPS.  So a solution these magic numbers would be to append our magic numbers to our currently existing constants.py module, and store them there.  Something like:

```
# Filepath Name Lengths
PROGRAM_ID_LEN = 5
OBSERVATION_LEN = 3
VISIT_LEN = 3
```

import this new constants.py module where its needed, and use the defines appropriately
```
r"(?P<program_id>\d{PROGRAM_ID_LEN})"\
r"(?P<observation>\d{OBSERVATION_LEN})"\
r"(?P<visit>\d{VISITS_LEN})"\
```


Once established we can use constants.py to clean up other magic numbers as we encounter them.

jwql/utils/utils.py
```diff
@@ -368,33 +377,49 @@ def filename_parser(filename):
...
+        'guider_segment'
     ]
 
     # Try to parse the filename
     for filename_type, filename_type_name in zip(filename_types, filename_type_names):
 
-        # If full filename, try using suffix
-        if not file_root_name:
+        # If full filename, try using suffix, except for *msa.fits files
+        if not file_root_name and 'msa.fits' not in filename:
```
BradleySappington: Is the 'msa.fits' the only suffix we will ever need isolated here?  Is this the only location 'msa.fits' is isolated in logic?  If there is potential for more in the future, this would be another good spot to put a variable list of excluded suffixes for this situation

jwql/utils/utils.py
bhilbert4: Done. Perhaps at the expense of a bit of readability. But then regular expressions are not the most readable things to begin with.

jwql/utils/utils.py
bhilbert4: I also put this in constants.py. Not sure if there will be any others, but there is now a place to add them later.

bhilbert4:
I also added tests for these two new possibilities in the filename_parser

BradleySappington (approved):
Looks good @bhilbert4 
Could potentially store all constants as a dictionary to make import easier, but that would also create more lines as you'd have to extract the values before putting into the regex to keep it from getting too muddled.  A-OK to keep as is :)

bhilbert4:
> Looks good @bhilbert4 Could potentially store all constants as a dictionary to make import easier, but that would also create more lines as you'd have to extract the values before putting into the regex to keep it from getting too muddled. A-OK to keep as is :)

I agree. I started to implement this as a dictionary, but that made readability in filename_parser() even worse.

BradleySappington:
Yeah, you’d have to assign the “variables” from the dictionary to make it legible, it would essentially just move the massive import line to a massive dictionary dump line.  Keep it as is I say.

From: Bryan Hilbert ***@***.***>
Reply-To: spacetelescope/jwql ***@***.***>
Date: Friday, June 24, 2022 at 2:58 PM
To: spacetelescope/jwql ***@***.***>
Cc: Bradley Sappington ***@***.***>, Mention ***@***.***>
Subject: Re: [spacetelescope/jwql] Add entry in filename_parser for segmented guider files (PR #962)


Looks good @bhilbert4<https://urldefense.com/v3/__https:/github.com/bhilbert4__;!!CrWY41Z8OgsX0i-WU-0LuAcUu2o!x8aBl0UMYrgWE5MNLwipxwU7HDdGYdklROIv0iSqTqMG7WoQC07iUUbhVOWfWoZ-kuDT_bdgvhcUR3n8KOpkykJPONRT$> Could potentially store all constants as a dictionary to make import easier, but that would also create more lines as you'd have to extract the values before putting into the regex to keep it from getting too muddled. A-OK to keep as is :)

I agree. I started to implement this as a dictionary, but that made readability in filename_parser() even worse.

mfixstsci:
> Looks good @bhilbert4 Could potentially store all constants as a dictionary to make import easier, but that would also create more lines as you'd have to extract the values before putting into the regex to keep it from getting too muddled. A-OK to keep as is :)

Agreed, I think it could be done this way and not cost much, but I think that is something we could improve on if we need to import those constants anywhere else :)

bhilbert4:
Woo hoo! Merging! Thanks for the review!

PR #963, opened 2022-06-03
Remove second set of colorbar tick labels from preview images

### LINKED ISSUE #960: Two sets of ticklabels on colorbars
Some preview images contain a colorbar where there are two sets of tick labels. This happens in cases where the max-min values of the labels are less than a factor of 10 - according to matplotlib developers. In this case, we need to turn off the tick labels that are automatically generated by matplotlib. 

I tried filing an issue, but the answer didn't seem to work: https://github.com/matplotlib/matplotlib/issues/22425

After playing with this some more today, I think I have a hacky version that works. Note that running this solution gives:
1) a warning that "AutoMinorLocator does not work with logarithmic scale" (but the plot is still made and saved correctly in my toy example)
2) You can use any number you want in the call to AutoMinorLocator and the result is the same.

Here is the updated code that seems to work:

```
import numpy as np
import matplotlib.colors as colors
mport matplotlib.pyplot as plt
from matplotlib.ticker import AutoMinorLocator

# Create image
image = np.random.random((256, 2048))

# Set the figure size
fig, ax = plt.subplots(figsize=(8, 6))

shiftmin = 1
shiftmax = 2

cax = ax.imshow(image, norm=colors.LogNorm(vmin=shiftmin,vmax=shiftmax), cmap='viridis')
plt.gca().invert_yaxis()

tickvals = np.logspace(np.log10(shiftmin), np.log10(shiftmax), 5)

#test adding a point that's out of range - not helpful
#tickvals = np.append(tickvals, 25)

format_string = "%.3f"
tlabelstr = [format_string % number for number in tickvals]

cbar = fig.colorbar(cax, ticks=tickvals, format=format_string, location='right', fraction=0.2, extend='both', label='something')

# This seems to correctly removes the points we want to remove, even though it gives a warning that
# it doesn't work on log scales.
cbar.ax.yaxis.set_minor_locator(AutoMinorLocator(n=0))

# This is where I'd like to run the fix given in the matplotlib issue above, but this gives an error because the number
# of entries in set_ticks (0) is not the same as the number of ticks (5).
#cbar.set_ticks([], which='minor')

cbar.set_ticklabels(tlabelstr)

# This is the fix suggested in the matplotlib issue above. But all this does is remove the points we
# want to keep while keeping those we want to remove. Using "major" rather than "minor" gives the
# same result
#cbar.set_ticks([], which='minor')

ax.set_xlabel('Pixels')
ax.set_ylabel('Pixels')
plt.savefig('tmp.png')
```

So it seems like all we need to do here is ignore the cbar.set_ticks([], which='minor') lines, and add the line containing the call to AutoMinorLocator().

bhilbert4:
Resolves #960 

This PR adds code that removes the extra set of colorbar ticks and ticklabels from the static preview images.

### DIFF
--- a/jwql/utils/preview_image.py
+++ b/jwql/utils/preview_image.py
@@ -37,6 +37,7 @@ Use:
 import logging
 import os
 import socket
+import warnings
 
 from astropy.io import fits
 import numpy as np
@@ -48,6 +49,7 @@ import matplotlib
 matplotlib.use('Agg')
 import matplotlib.pyplot as plt
 import matplotlib.colors as colors
+from matplotlib.ticker import AutoMinorLocator
 
 # Only import jwst if not running from readthedocs
 if 'build' and 'project' not in socket.gethostname():
@@ -205,6 +207,10 @@ class PreviewImage():
                         data = hdulist[ext].data[:, [0, -1], :, :].astype(np.float)
                     else:
                         data = hdulist[ext].data.astype(np.float)
+                    try:
+                        self.units = hdulist[ext].header['BUNIT']
+                    except KeyError:
+                        self.units = ''
                 else:
                     raise ValueError('WARNING: no {} extension in {}!'.format(ext, filename))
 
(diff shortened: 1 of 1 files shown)


bhilbert4:
I tested this locally, and all looks good. @mfixstsci I think this is ready for a review.

jwql/utils/preview_image.py
```diff
@@ -330,8 +336,15 @@ def make_figure(self, image, integration_number, min_value, max_value,
                 format_string = "%.{}f".format(dig)
                 tlabelstr = [format_string % number for number in tlabelflt]
                 cbar = fig.colorbar(cax, ticks=tickvals)
+
+                # This seems to correctly remove the ticks and labels we want to remove. It gives a warning that
+                # it doesn't work on log scales, which we don't care about. So let's ignore that warning.
+                warnings.filterwarnings("ignore", message="AutoMinorLocator does not work with logarithmic scale")
```
mfixstsci, line 342: @bhilbert4 I am assuming we aren't too worried about the log scaling warning here because we are just trying to remove the minor ticks all together?

mfixstsci (approved):
This all looks straight forward. Thank you for adding this change!

jwql/utils/preview_image.py
bhilbert4: This removes the default set of minor ticks and tick labels, so that all that remains is the custom set of ticks and tick labels.

bhilbert4: This is the only way I found to remove the defaults. Not ideal given the warning, but it works. My Issue on the matplotlib repo didn't really result in a useful answer.

mfixstsci: @bhilbert4 Yeah I remember seeing the back and forth resulting in nothing. This will have to do for now, we can always iterate in the future. Merging.

PR #971, opened 2022-07-06
[WIP]: Fix EDB download button

bhilbert4:
This PR fixes the Download Data button on the EDB page. I've also updated the name of the file to include the mnemonic name and start/end dates, rather than being generic.

### DIFF
--- a/jwql/website/apps/jwql/data_containers.py
+++ b/jwql/website/apps/jwql/data_containers.py
@@ -361,9 +361,9 @@ def get_edb_components(request):
                         # save file locally to be available for download
                         static_dir = os.path.join(settings.BASE_DIR, 'static')
                         ensure_dir_exists(static_dir)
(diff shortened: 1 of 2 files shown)


bhilbert4:
This now allows me to download EDB search results. Looking at the changes though, I'm not sure why it wasn't working before. All I really did was change the variable name and updated the name of the file to save into.

bhilbert4:
I should say that this worked for me locally. Probably still a good idea to test on one of the servers.

bhilbert4:
Switching back to the develop branch and re-testing (locally), I'm now able to download the file. Very confusing. But I've confirmed that the develop branch on the servers fails to download the file.

bhilbert4:
Testing on the test server: The csv file is being saved in the expected location. So the issue must be that when the download button is pressed, it's looking in the wrong place for the file?

PR #976, opened 2022-07-07
fix view proposal button href

BradleySappington:
Hotfix for obsnum/explore_image merge


bhilbert4 (approved):
Looks good. I tested on the test server and everything appears to be working now. Thanks for the quick fix!!

PR #977, opened 2022-07-08
Ignore 1d spectrum files

### LINKED ISSUE #974: Add x1d to the list of ignored suffixes
x1d files are ascii tables, and so should be ignored. Currently x1d shows up as an option when looking at preview images: 
https://jwql.stsci.edu/NIRISS/jw01324001001_11101_00004_nis/

Switching to this shows a broken link, but also seems to resize things such that when switching back to a different suffix, the preview image is very small.

bhilbert4:
Resolves #974 

This PR adds three suffixes (x1d, x1dints, c1d) to the list of ignored suffixes. These files all contain 1D spectra, and therefore will not have preview images. By ignoring the suffixes, we will be sure that radio buttons for those extensions do not appear on pages for individual observations.

I tested this on the test server with: https://jwql.stsci.edu/NIRCam/jw01442003001_02103_00001_nrcalong/
and the x1dints option disappeared from the list of radio buttons.

@mfixstsci this is pretty simple. I think it's ready for review.


mfixstsci (approved):
LGTM

PR #978, opened 2022-07-10
Adding Next and Previous Buttons

### LINKED ISSUE #973: Add Next and Previous buttons to exposure pages
From @slamassa:
Maybe this functionality exists and I missed it, but is it possible to navigate to the “next” image when you have one image selected? I didn’t find a way to do so, so I use the “back” button on my browser to get back to the list of images, and then because I have a goldfish memory, I don’t remember which image in an observation I was inspecting and which the next one is. With ~50 images in 1 observation, it’s really easy for me to lose track of what I inspected and what I need to inspect. (Not sure if this rises to the level of “critical” but I figured I’d bring it up.)

mfixstsci:
Using `django.core.paginator`


jwql/website/apps/jwql/views.py
```diff
@@ -37,12 +37,14 @@
...
 """
 
+from collections import defaultdict
 import csv
 import os
 
 from bokeh.layouts import layout
 from bokeh.embed import components
 from django.http import HttpResponse, JsonResponse
+from django.core.paginator import Paginator
```
BradleySappington: Paginator not used, can be deleted

jwql/website/apps/jwql/views.py
```diff
@@ -960,9 +965,24 @@ def view_image(request, inst, file_root, rewrite=False):
...
+    rootnames = get_rootnames_for_instrument_proposal(inst, prop_id)
+    file_root_list = defaultdict(list)
+
+    for root in rootnames:
+        try:
+            file_root_list[(filename_parser(root)['observation'])].append(root)
+        except KeyError:
+            pass
+
+    file_root_list = {key : sorted(file_root_list[key]) for key in sorted(file_root_list)}
```
BradleySappington: currently an initial sort.  Sorting to follow page sorting will be implemented with a future PR?

BradleySappington (approved):
Looks good @mfixstsci

jwql/website/apps/jwql/views.py
mfixstsci: So this was my first attempt at sorting and just figured I would do it alphabetically by default. Should we go ahead and remove it and implement it in another pr?

jwql/website/apps/jwql/views.py
mfixstsci: Thanks -- I'll make sure to do this.

jwql/website/apps/jwql/views.py
BradleySappington: Leaving it this way is a-ok.  Some sort of known sort is better than leaving it out.  Verifying that we weren't pursuing advanced sorting in this PR.

mfixstsci:
Looks like I merged the TA monitor, accidentally force pushed, and then re-merged it again.

PR #994, opened 2022-07-19
Alert user to non-existant proposal number

### LINKED ISSUE #992: NIRCam 2739 program page crashes
https://jwql.stsci.edu/NIRCam/archive/2739/obs004/ gives a "list index out of range" error and crashes. Looks like an issue with scraping the proposal information

bhilbert4:
Resolves #992 

This PR adds some code to prevent a page from crashing in the case where a user types a proposal number for a non-existent or unavailable program into a URL. In this case, a note is added to the proposal information stating that the proposal does not exist or is unavailable.


bhilbert4:
@mfixstsci  I think this is ready for review.

bhilbert4:
I checked on the test server and this looks good. The page for NIRCam program 2739 now loads without crashing, even though there is no public phase 2 information.

mfixstsci (approved):
I wrote the little web scraper here and didn't anticipate some of the meta data could be hidden. Thanks for the work around.

PR #996, opened 2022-07-20
Fix radio buttons with e.g. o001

### LINKED ISSUE #983: Make the integration slider bar smarter
Currently the integration slider bar that is under the static preview image shows "1/N" for cases where the data contain a single integration. 

It would be more clear if the bar showed "1/1" in these cases. Also, the slider bar really should only appear for a handful of calibration filetypes. Only uncal, rateints, calints, and dark filetypes will ever have more than one integration. (Check to be sure this is the complete list). All other file suffixes will be guaranteed to have a single integration, in which case maybe the bar should not appear at all? Or at least just show "1/1".

### LINKED ISSUE #990: o001 radio button
Some image view images have a radio button for o001 as an option. I think these are actually from the crfints files, which have the filenaming convention of e.g. jw01442001001_04103_00001-seg004_nrca3_o001_crfints.fits. The filename parser must be grabbing o001 as the part of the filename after the detector name, and assuming that is the calibration suffix.

We should:
1) see if clicking on the radio button shows the correct image or not -- it doesn't. It assumes a filename ending in _o001.fits
2) change the listed suffix to be crfints

bhilbert4:
Resolves #990 

This PR fixes the bug that was causing entries such as o001 to show up in the list of filename suffixes in the radio buttons for an individual file. 

Previously, the suffix of the file was being found by assuming that it was the 4th section of the filename (separated by '_'). This doesn't work for crf files because they have an extra section that lists the association value, immediately before the 'crf' suffix. By using the filename parser, this should now work correctly for all file types.


bhilbert4:
- [x] Still needs to be tested on the test server.

bhilbert4:
Tested on the test server and it seems to be working. The radio button for crf files now includes the name of the association  (e.g. 'o001_crf'). This is because crf filenames have an extra string when you break up a filename between underscores. 

I also found and fixed the bug that was causing the slider bar under the image to always list the image as '1/NaN' integrations. There was a typo in the directory being searched for jpgs. I also tweaked the code a bit so that the value now used in place of the NaN is the the total number of integrations in the file, rather than the number of integrations for which we have created preview images. These numbers are only different for exposures that have more than 100 integrations, which means only TSO observations I think. But using the true number of integrations is much less confusing than the number of jpg files, since for observations with more than 100 integrations we start making preview image for every 100th integration, in order to save disk space. I also tested this on the test server and it looks good.

- [x] Check a crfints example
Actually, I'm not sure where to find crfints data. That will only be made if someone runs the stage 3 pipeline and feeds it calints files. I'm not sure under what situation that would happen. Maybe only for moving target (i.e. solar system) data?

bhilbert4:
I just checked the Jupiter observations from commissioning and there are no crfints files in the filesystem. So maybe we assume this fix is ok for those too and fix it if we find a problem in the future? 

If you agree @mfixstsci then this is ready for review.

mfixstsci (approved):
LGTM.

mfixstsci:
@bhilbert4 I went reviewed and fixed conflicts. Anything else before merging?

bhilbert4:
@mfixstsci I think it's ready to go!

PR #968, opened 2022-07-01
Turn off thumbnail creation for all but rate files

bhilbert4:
Resolves #957 

This PR makes changes to the preview image generation code such that thumbnail images are created only for rate files. Currently we are creating thumbnail files for all file types (uncal, rate, cal, etc). However, we only need one thumbnail for each file, since the instrument archive pages do not show separate images for each file type.

PR #997, opened 2022-07-21
 Fix bad thumbnails

### LINKED ISSUE #958: Bad thumbnail images
Looks like we are still seeing some bad thumbnail images even after #949 . We need to investigate whether these are left over bad pixels from before the fix, or if there are still bad thumbnails being produced.

bhilbert4:
This PR makes some tweaks to the preview image generator in an effort to finally fix the bad thumbnail images that have been plaguing us forever. 

Previously, the thumbnail images were created using matplotlib's imshow() function. This function creates an AxesImage object. However, preview images are generated by first creating a matplotlib Figure object, and then populating the axes of that object with an AxesImage object via imshow(). It turns out that `plt.close` only closes Figure objects. This means that thumbnail images were not being closed at the end of the preview image generation process, and this must have been leading to multiple images open at the same time, resulting in bad thumbnail images.

This PR changes the thumbnail creation to be like that of the preview images. We first create a Figure object and then populate that using imshow. Local testing on several programs showed no bad thumbnails while using this. We should try a larger scale test on the server, but it looks promising that this PR will put an end to our bad thumbnail images.

I also fixed a small bug that was preventing some thumbnails from being created, where NaN values in the data were causing bad min/max values for the display. 

Resolves #958

### DIFF
--- a/jwql/jwql_monitors/generate_preview_images.py
+++ b/jwql/jwql_monitors/generate_preview_images.py
@@ -672,6 +672,10 @@ def process_program(program):
 
     # remove specific "ignored" suffix files (currently "original" and "stream")
     filenames = [filename for filename in filenames if os.path.splitext(filename.split('_')[-1])[0] not in IGNORED_SUFFIXES]
+
+    # Remove guiding files, as these are not currently visible in JWQL anyway
+    filenames = [filename for filename in filenames if 'guider_mode' not in filename_parser(filename)]
+
     logging.info('Found {} filenames'.format(len(filenames)))
     logging.info('')
 
--- a/jwql/utils/preview_image.py
+++ b/jwql/utils/preview_image.py
@@ -302,9 +302,9 @@ class PreviewImage():
             if thumbnail:
                 self.fig, ax = plt.subplots(figsize=(3, 3))
                 cax = ax.imshow(shiftdata,
-                                      norm=colors.LogNorm(vmin=shiftmin,
-                                                          vmax=shiftmax),
-                                      cmap=self.cmap)
+                                norm=colors.LogNorm(vmin=shiftmin,
+                                                    vmax=shiftmax),
+                                cmap=self.cmap)
                 # Invert y axis
                 plt.gca().invert_yaxis()
 
--- a/jwql/utils/preview_image.py
+++ b/jwql/utils/preview_image.py
@@ -167,6 +167,11 @@ class PreviewImage():
         """
         nelem = np.sum(pixmap)
         numclip = np.int(clipperc * nelem)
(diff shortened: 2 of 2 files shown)


bhilbert4:
We also need to explicitly set the figure size for thumbnail images now, with these changes. At the moment that is hardwired to be 3x3. Is this appropriate? We don't want the figures to be much larger than necessary, since that would slow down image loading and therefore the JWQL webpages.

bhilbert4:
Tests on the test server were successful. I didn't see any bad thumbnails when using this code. Also, running generate_preview_images is a lot faster when it ignores all of the guide star tracking files. 

I feel like this has been a long battle that @BradleySappington, @mfixstsci, and I have been fighting. I'm excited for it to be over. 🤞 

@mfixstsci I think this is ready for review.

BradleySappington:
I'm absurdly excited

mfixstsci:
@bhilbert4 @bhilbert4 lets go! I am going to start my review

jwql/jwql_monitors/generate_preview_images.py
```diff
@@ -672,6 +672,10 @@ def process_program(program):
 
     # remove specific "ignored" suffix files (currently "original" and "stream")
     filenames = [filename for filename in filenames if os.path.splitext(filename.split('_')[-1])[0] not in IGNORED_SUFFIXES]
+
+    # Remove guiding files, as these are not currently visible in JWQL anyway
```
mfixstsci, line 683: Good catch, will save us time.

jwql/utils/preview_image.py
```diff
@@ -295,20 +300,21 @@ def make_figure(self, image, integration_number, min_value, max_value,
 
             # If making a thumbnail, make a figure with no axes
             if thumbnail:
-                fig = plt.imshow(shiftdata,
-                                 norm=colors.LogNorm(vmin=shiftmin,
-                                                     vmax=shiftmax),
-                                 cmap=self.cmap)
+                self.fig, ax = plt.subplots(figsize=(3, 3))
```
mfixstsci, line 303: Ha, this must be the `plt`/`fig` object mix up.

mfixstsci (approved):
Looks good @bhilbert4 I know you and @BradleySappington have been working hard on this and hope this solves it once and for all.

jwql/utils/preview_image.py
bhilbert4: Yeah, this was the crux of the problem with the bad thumbnails.

bhilbert4:
Fixed a small conflict that came about after the merge of #968

PR #1001, opened 2022-07-25
Allow proposal thumbnails to come from dark thumbnails

bhilbert4:
This PR adds code to allow proposal-level thumbnails to be created from dark file thumbnails. Currently only rate or uncal thumbnails will be copied over to become proposal level thumbnails. But with #997 we are now making thumbnails only for rate.fits and dark.fits files. So for proposals that contain only dark.fits data, we need to be able to copy one of the resulting thumbnails over to be the proposal level thumbnail.

### DIFF
--- a/jwql/utils/preview_image.py
+++ b/jwql/utils/preview_image.py
@@ -217,6 +217,15 @@ class PreviewImage():
                 if 'PIXELDQ' in extnames:
                     dq = hdulist['PIXELDQ'].data
                     dq = (dq & dqflags.pixel['NON_SCIENCE'] == 0)
+
+                    # Case where all pixels are flagged as NON_SCIENCE
+                    # This shouldn't happen, but preview_image did crash
+                    # at one point saying that there were no good pixels to
+                    # work with.
+                    if not np.any(dq):
+                        logging.warning('All pixels flagged as NON_SCIENCE. Using all pixels to determine scaling.')
+                        yd, xd = data.shape[-2:]
+                        dq = np.ones((yd, xd), dtype="bool")
                 else:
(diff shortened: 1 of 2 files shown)


mfixstsci:
@bhilbert4 just wanted to check, is this ready for review?

bhilbert4:
@mfixstsci I think this is ready for review. It hasn't been tested on one of the servers yet. Do you think it's worth a test?

mfixstsci:
@bhilbert4 I would say we review it just in case. Don't want to add to anymore thumbnail troubles we've been experiencing 😓

bhilbert4:
Sounds good. When the run of new MIRI thumbnails is finished on the test server I'll switch to this branch and try it out.

jwql/utils/preview_image.py
```diff
@@ -161,6 +161,14 @@ def find_limits(self, data):
         # Ignore any pixels that are NaN
         finite = np.isfinite(data)
 
+        # If all non-science pixels are NaN then we're sunk. Scale
+        # from 0 to 1.
+        if not np.any(finite):
```
mfixstsci, line 166: Are non-science pixels always `NaN`?

mfixstsci (approved):
LGTM! @bhilbert4 I just had a question about the non-science pixels.

jwql/utils/preview_image.py
bhilbert4: No. Avoiding NaN pixels is a separate issue. In this block I'm looking across all pixels (not just non-science pixels) for NaN values. I don't expect this condition where all pixels are NaN to ever happen, but I figured it would be good to cover it just in case my expectations are wrong.

PR #1006, opened 2022-07-27
Default scaling params for preview images where all pix are NaN

### LINKED ISSUE #1007: Preview image unable to be scaled when all pixels are NaN
In a recent run of generate_preview_images, the code crashed when it ran into an image where all of the science pixels were NaN. In this case, the code had no good pixels from which to select the min and max values to use for scaling. 

In this case when combing the map of science pixels with the map of pixels with a finite signal:
pixmap = pixmap & finite
the resulting pixmap was False for all pixels.

And so when indexing the data array with pixmap, the result was an empty array.
sorted = np.sort(data[pixmap], axis=None)

bhilbert4:
A recent run of generate_preview_images failed when it came across an image where all science pixels were NaN. In this case, the find_limits() method of preview_image failed because there were no good pixels from which to select a min and max value for the image scaling.

This PR tweaks find_limits() such that if there are no good science pixels, it falls back to scaling the image from 0 to 1.


bhilbert4:
@BradleySappington @mfixstsci sorry. One more quick review if you have a moment. This issue crashed the preview image generator yesterday and I forgot to fix it until just now. Unfortunately the log didn't provide the name of the offending file, but based on the traceback, it must have encountered an image where all pixels were NaN. (Either that or a case where a subarray was contained completely within an area of non-science pixels, but I don't think there are any subarrays that are actually defined that way).

BradleySappington (approved):

bhilbert4:
Thanks @BradleySappington! Sorry to distract you from interactive preview images this morning.

PR #1008, opened 2022-07-27
Allow archive page to be sorted by time

### LINKED ISSUE #984: Allow instrument archive page to be sorted by time
Currently the Archive page for an instrument can be sorted by increasing or decreasing program ID number. We should also add the ability to sort the programs by execution time, so that the most recently acquired data appear at the top. Since JWST doesn't observe programs in order by number, being able to sort on time will be helpful for folks doing quicklook duties each day.

bhilbert4:
Resolves #984 

This PR adds a sorting function to the archive page that allows the proposals to be sorted according to exposure start time. This will allow the most recent proposals to appear towards the top of the page, which will help teams to notice new observations that need to be viewed.


bhilbert4:
This is now functional and appears to be working correctly. 

This PR also makes a small tweak to the sorting of thumbnails for an individual observation. The definition of the previously-used 'Default' order was a) not obvious to the user and b) most recent exposure start time first. So this PR renames Default to be 'Most Recent', whose name is consistent with the 'Most Recent' for the sorting of the archive page. The previously-defined sorting by 'Exposure Start Time' has also been renamed to 'Oldest', since it puts the oldest files first. I think this sorting is identical to sorting by filename since the filenames are created using counters that don't span multiple observations.

Some discussion points before a code review (@mfixstsci @BradleySappington):
1) What should the default sorting be upon the initial archive page load? Currently the programs on the archive page are ordered by decreasing program ID number. The default ordering for the files in an individual observation is most recent first. Being consistent between the two might be nice, although maybe that's not really necessary? It seems like it might depend on how most users use the site.
2) With sorting by filename and sorting with oldest first being identical, should we remove the oldest first option? Or maybe we should keep both since users might look for oldest first given that we have Most Recent?

BradleySappington:
1) I think most recent first is the way to go.  Eventually we will be filtering out the older ones anyway which makes me think folks will inherently been working with newer data
2) It seems like we just need sorting by oldest/newest and "filename" sorting just falls out.  You can probably remove the filename sort, and have all sorting be time based so its easily understandable.

bhilbert4:
Banging my head against the wall trying to get the default sorting on the archive pages to be most recent first. It seems like every time I reload the page, the dev server gives me a different answer. I'm also getting a lot of internal server errors that go away after a few minutes. 

I \*think\* the code is correct. The list of proposal numbers and exp start times are ordered with the most recent first when archived_proposals_ajax() in views.py finishes. So the order is correct going into archive.html and jwql.js.

bhilbert4:
@mfixstsci this is ready for review. It works successfully on the test server. The dev server gives inconsistent results and changing error statements when I reload multiple times. The more I tested that though, the more I'm convinced that's a problem with the dev server rather than this code.

jwql/website/apps/jwql/data_containers.py
```diff
@@ -1541,7 +1548,8 @@ def thumbnails_ajax(inst, proposal, obs_num=None):
     data_dict['dropdown_menus'] = dropdown_menus
     data_dict['prop'] = proposal
 
-    # Order dictionary by descending expstart time.
+    # Order dictionary by descending expstart time. This will dictate the order of the
+    # observation's thumbnails upon loading the observation's page.
     sorted_file_data = OrderedDict(sorted(data_dict['file_data'].items(),
```
mfixstsci, line 1564: Nice use of `OrderedDict` and `lambda` here.

jwql/website/apps/jwql/views.py
```diff
@@ -317,21 +328,30 @@ def archived_proposals_ajax(request, inst):
...
+            all_proposal_info['num_proposals'] = int(all_proposal_info['num_proposals'] + 1)
             all_proposal_info['proposals'].append(proposal)
             all_proposal_info['min_obsnum'].append(proposal_info['observation_nums'][0])
             all_proposal_info['thumbnail_paths'].append(proposal_info['thumbnail_paths'][0])
             all_proposal_info['num_files'].append(num_files)
+            all_proposal_info['expstart'].append(exp_start)
+
+    # Sort proposals from most recent to oldest, in order to have this as the
+    # default when the page is loaded
+    #print('should we do this? or should we keep the default to be increasing proposal ID order?')
```
mfixstsci: I think this is a good idea since we are trying to look at new data as it's rolling in.

mfixstsci (approved):
Changes look good to me! Thanks for submitting it @bhilbert4

mfixstsci:
Hey @BradleySappington and @bhilbert4 I know this has been sitting here a while. @BradleySappington do you know if these changes will conflict with your sorting PR?

BradleySappington:
I'm afraid that we double dipped this work.  This PR may now be obsolete.  Confirming with @bhilbert4 and @mfixstsci

BradleySappington:
This work was handled in a different PR.  Confirmed with @bhilbert4 to close without merge.

PR #1010, opened 2022-07-27
Add inst select home search

mengesser:
Adding Mees' solution for returning multiple instruments.


mfixstsci:
@mengesser had to clean up some stuff from your original pull request but now all of the tests are passing, just need to test on one of the servers to verify!

mfixstsci:
@BradleySappington @bhilbert4 this is ready for review!

bhilbert4:
I'm getting an error when I test this using a program that does not contain multiple instruments. (e.g. 1068 is nircam-only).

jwql/utils/constants.py
```diff
@@ -238,6 +238,9 @@
                                       'G235H', 'G395H', 'PRISM']
                           }
 
+# Filename extensions for guider data
+GUIDER_FILENAME_TYPE = ['guider', 'gs-fg', 'gs-track', 'gs-id', 'gs-acq1', 'gs-acq2']
```
bhilbert4: 'guider' will be in science files, while the 'gs-' options will be in guide star data. So it seems like you want to take out the 'guider' option given the way you look for matches in forms.py, right?

mfixstsci:
@bhilbert4 should be fixed now.

jwql/utils/constants.py
mfixstsci: Thanks for information @bhilbert4, fixed this in [f0c68d8](https://github.com/spacetelescope/jwql/pull/1010/commits/f0c68d8efe2f6f431b1363af00073bfb89a2bcd0)

mfixstsci:
Okay @bhilbert4 I have all of the updated changes on dev. I tested your example: `01068` (NRC only) and my example `02739` which contains MIR and NRC data but excludes the guider data.

bhilbert4:
Functionally this looks good! I tested with proposals that use various combinations of instruments, and all the returned values looked good.

mfixstsci:
Thanks @bhilbert4!

bhilbert4 (approved):
This looks good to me. Nice work! I think this box is going to get a lot of use.

PR #1024, opened 2022-08-18
Scaling tweaks for preview images/thumbnails

bhilbert4:
This PR makes a small change to the scaling algorithm for the creation of preview images and thumbnails. Previously the code looked for the PIXELDQ extension in order to retrieve a DQ map and find the NON_SCIENCE pixels. However PIXELDQ wil only exist in intermediate files saved in calwebb_detector1. Once the rate files are saved at the end of that pipeline, the extension name is changed to DQ. So this update has the code look for the DQ extension, as JWQL never sees files between uncal and rate.

Also, note that MIRI does not populate NON_SCIENCE flags until the flat field step in calwebb_image2, so this method of retrieving NON_SCIENCE pixels does not work for MIRI rate files. It also does not work for any instruments' uncal files, as there are no DQ-related extensions in those files.


bhilbert4:
@mfixstsci @BradleySappington I think this is finally ready for review. I've run it on the test server on all program numbers > 2000 and the results look good. There may be some small tweaks that can be made to improve some of the higher level products, but I don't see any preview images any more that are badly scaled such that targets aren't visible.

I ended up having to add files containing maps of non-science pixels. Uncal, i2d, and MIRI rate files don't contain DQ extensions, and so previously for those files, JWQL did not know which pixels to omit when scaling the preview image. So now for these files, an external file is read in and used as the DQ array.

Once this is reviewed, I think we may finally be ready to make a big push and regenerate all the preview images for existing data.

jwql/tests/test_preview_image.py
```diff
@@ -42,6 +44,46 @@
     ON_READTHEDOCS = os.environ['READTHEDOCS']
 
 
+def test_crop_to_subarray():
```
mfixstsci, line 47: @bhilbert4 you rock, thank you for writing tests.

jwql/jwql_monitors/generate_preview_images.py
```diff
@@ -537,6 +537,11 @@ def generate_preview_images():
     program_list = [os.path.basename(item) for item in glob.glob(os.path.join(SETTINGS['filesystem'], 'public', 'jw*'))]
     program_list.extend([os.path.basename(item) for item in glob.glob(os.path.join(SETTINGS['filesystem'], 'proprietary', 'jw*'))])
     program_list = list(set(program_list))
+
+
+    program_list = ['jw01068', 'jw02260', 'jw01837', 'jw01532', 'jw01566', 'jw02560', 'jw02420', 'jw02736']
```
mfixstsci: @bhilbert4 is there a reason you rewrite the program list here?

jwql/tests/test_preview_image.py
mfixstsci, line 47: Thank you for writing tests, this is great.

mfixstsci (approved):
Hey @bhilbert4 this looks good to me!

jwql/jwql_monitors/generate_preview_images.py
bhilbert4: Removed. That was leftover from development.

PR #1026, opened 2022-08-19
fix axis values in explorer mode

### LINKED ISSUE #1025: Axis flip in explore_image
It looks like the axes are flipped in explore_image. 
See here:
/NIRSpec/jw02560013001_41101_00002_nrs2_cal/explore_image/

In this case, the static preview image has dimensions (x, y) = (~20, 2048), while in explore image, the dimensions are (2048, ~20).  But the image itself seems to show a spectral trace going from left to right. So it looks like the image itself is correct, and only the axes are flipped.

Here's a NIRCam observation that uses a similarly-shaped subarray, and the situation is the same. The data look correct, but the axes are flipped.
/NIRCam/jw01442001001_04103_00001-seg002_nrca3_rate/explore_image/

@BradleySappington

BradleySappington:
ready for review

### DIFF
--- a/jwql/utils/interactive_preview_image.py
+++ b/jwql/utils/interactive_preview_image.py
@@ -122,7 +122,7 @@ class InteractivePreviewImg():
                 palette="Viridis256", low=limits[0], high=limits[1])
             ticker = BasicTicker()
         yd, xd = self.data.shape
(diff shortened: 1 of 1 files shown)


mfixstsci (approved):
LGTM, thanks for getting this in quickly!

PR #1030, opened 2022-08-26
Build archive page from database

### LINKED ISSUE #1023: Speed up archive page loading
Many of the app pages are really slow to load. We should look into how to speed them up.

bhilbert4:
Resolves #1023 

This PR changes the way an instrument's archive page is built. The task of querying MAST for each proposal and then working through all of the files for that proposal has been moved to a script that can be run any time. The results of the script are used to populate new database tables.

Then when someone navigates to an instrument's archive page, the database table is loaded and the information is used to build the page.


bhilbert4:
This is still very much a work in progress, but I figured I should get the skeleton of it uploaded. I'm always very slow when trying to work with databases.

BradleySappington (commented):
Initial thoughts that we will eventually implement an "Image" class to hold the observation specific thumbnails and filterable data.  This Image class should be where we have the Exposure Type kept.  Exposure Types can just be defined and stored as objects in the Image class (along with name and thumbnail location).

BradleySappington:
can also decide if we want to handle images and thumbnails using the `FileField`, `ImageField`, or just maintaining a path location as is

bhilbert4:
The new page is working, but is still slower than I had anticipated.

Looking at how long it's taking to do various things, the database access appears fairly fast. The slowest task is generating the list of proposal numbers. Timing results are below for the updated archived_proposals_ajax. Values are in units of seconds. Each "proposal loop" iteration shows the results when working on one proposal. Everything within the proposal loop is small (5-20ms), compared to getting the list of proposals (10-12 seconds). And note that this is timing the tasks in archived_proposals_ajax only. This doesn't say anything about archived_proposals nor the JS. But still, this argues for finding a faster way to get a list of proposals.

Initial filter:  0.0008904417045414448
Get list of proposals: 12.152243920136243
In proposal loop:
    filter:  0.00029786117374897003
    thumbnail entry:  0.02204055292531848
    get obsnums:  0.008105554152280092
    get filecount:  5.180947482585907e-06
  
In proposal loop:
    filter:  0.0003039059229195118
    thumbnail entry:  0.022973330225795507
    get obsnums:  0.014654692728072405
    get filecount:  6.702262908220291e-06
  
In proposal loop:
    filter:  0.00036222394555807114
    thumbnail entry:  0.015787335112690926
    get obsnums:  0.007101284805685282
    get filecount:  5.035195499658585e-06

bhilbert4:
@mfixstsci this is ready for review. Works well when tested on the test server.

bhilbert4:
Ah , right. I forgot about PEP8...

bhilbert4:
Hmm, PEP8 seems to be stuck checking an old version of the files.

jwql/website/apps/jwql/views.py
```diff
@@ -272,67 +275,45 @@ def archived_proposals_ajax(request, inst):
...
-
-        if len(filenames) > 0:
-            # Gather information about the proposals for the given instrument
-            proposal_info = get_proposal_info(filenames)
-            all_proposal_info['num_proposals'] = all_proposal_info['num_proposals'] + 1
-            all_proposal_info['proposals'].append(proposal)
-            all_proposal_info['min_obsnum'].append(proposal_info['observation_nums'][0])
-            all_proposal_info['thumbnail_paths'].append(proposal_info['thumbnail_paths'][0])
-            all_proposal_info['num_files'].append(num_files)
+
```
mfixstsci, line 278: This is awesome, you really obscure a lot of the details here and this looks and behaves more efficiently. Great work @bhilbert4

mfixstsci (approved):
Looks great, I really like this models based approach. thanks for submitting this!

jwql/website/apps/jwql/views.py
bhilbert4: Thanks!

mfixstsci:
I also checked out the pep8 issues listed here and pulled your repo locally and it looked fine 🤷‍♂️

PR #1032, opened 2022-08-30
Add overwrite option when generating preview images

bhilbert4:
This PR adds an `overwrite` option to generate_preview_images.py, which will allow existing preview images and thumbnail images to be re-created and overwritten in cases where they already exist. This will make it easier to regenerate preview images, since we won't have to go in and manually delete them prior to running the code. This will also eliminate the time where a particular file has no preview image, which is currently the case between the time when the old image is deleted and the new image is generated.


bhilbert4:
@mfixstsci @BradleySappington this is ready for review. I tested it on the test server and all looks well. I ran it on program 01068 both with and without the new `overwrite` flag and it worked as expected.

I also removed a strange line from preview_images.py that was causing a crash. I'm not sure how that line got in there (or maybe, how it was not removed in a prior PR), but it was referencing `pixmap` before it existed. I believe that line was meant to be replaced by the line below it (that uses self.dq rather than pixmap) in a previous PR. Either way, that whole function is re-written in #1024, which we also need to use when regenerating the preview images on ops.

jwql/jwql_monitors/generate_preview_images.py
```diff
@@ -529,16 +540,23 @@ def get_base_output_name(filename_dict):
...
+    """
 
     # Process programs in parallel
     program_list = [os.path.basename(item) for item in glob.glob(os.path.join(SETTINGS['filesystem'], 'public', 'jw*'))]
     program_list.extend([os.path.basename(item) for item in glob.glob(os.path.join(SETTINGS['filesystem'], 'proprietary', 'jw*'))])
     program_list = list(set(program_list))
+    program_list = [(element, overwrite) for element in program_list]
     pool = multiprocessing.Pool(processes=int(SETTINGS['cores']))
-    results = pool.map(process_program, program_list)
+    results = pool.starmap(process_program, program_list)
```
BradleySappington, line 559: Since each tuple will have the same second parameter in the list, you could use the program_list with map (instead of starmap), and just send overwrite as the 'y' parameter.  This wouldn't change the functionality, just simplifies the code.  This is fine as is, just giving feedback.

jwql/jwql_monitors/generate_preview_images.py
```diff
@@ -701,10 +724,16 @@ def process_program(program):
...
+        if not overwrite:
+            # If overwrite is False, we create preview images only for files that
+            # don't have them yet.
+            file_exists = check_existence([filename], preview_output_directory)
+            if file_exists:
+                logging.info("\tJPG already exists for {}, skipping.".format(filename))
+                continue
+        else:
+            # If overwrite is set to True, then we always create a new image
+            file_exists = False
```
BradleySappington, line 736: Is this else statement needed?

BradleySappington (approved):
minor comments, nothing critical

jwql/jwql_monitors/generate_preview_images.py
bhilbert4: Yes. If overwrite is set to True, then we fall into this else block, which is basically lying and telling the code below that the preview image for the file does not exist. That then allows PreviewImage to be called and the image to be regenerated.

jwql/jwql_monitors/generate_preview_images.py
bhilbert4: I'm not sure I understand. map() only accepts one parameter, right? So in that case I'd have to also make program_list a tuple, and I'd have to change the parameters of process_program() such that it accepts a single parameter that is then broken out into program id and overwrite status, right? Or am I misinterpreting the way that map() works?

jwql/jwql_monitors/generate_preview_images.py
BradleySappington: I dont see any additional spots where file_exists is used.  The initial `if not overwrite:` is essential, but I dont think the `else` is needed

jwql/jwql_monitors/generate_preview_images.py
BradleySappington: I withdraw my comment, Map can take multiple parameters, but they both need to be iterable anyway, this way makes more sense.

PR #1036, opened 2022-09-05
Hostname contents and list lengths

snoyes:
In the database interface, `and` cannot be used to check for two substrings like this code attempts. This code effectively only checks for the word 'project' in the hostname. Correctly check for both words in the socket hostname.

In the preview image, the group and integ are allowed to be integers or lists of length 2, but there was no check if the list was of length 0 or 1.


BradleySappington (approved):
Thanks for your submission!

PR #1043, opened 2022-09-16
Sort Filenames Based on User Selected Dropdown

mfixstsci:
This update sorts the filenames in the list passed to the next and previous buttons. This ensures that the order they appear in the archive view is the same as the order passed to the next and previous button div elements.


mfixstsci:
@BradleySappington took up this work and implemented the sorting in #1052, closing.

PR #1045, opened 2022-09-20
Add missing MAST_QUERY_LIMIT import

bhilbert4:
This PR adds a missing import to the archive database population script. The import of MAST_QUERY_LIMIT from constants.py is used to check that the MAST query is not bumping up against the maximum number of returned records.

@mfixstsci this is ready for review


mfixstsci (approved):
LGTM

PR #1048, opened 2022-09-26
Raise the upper limit on number of returned Mast entries

bhilbert4:
By default, querying Mast through astroquery will return only 50,000 entries. Any query that hits this limit will take several minutes to complete, and will return only 50,000 entries regardless of how many entries actually match the search criteria. This PR raises that limit to be 500,000. At the moment, the only time we were running into the 50,000 limit was when querying for all filenames for PID 1187. This is an early WFSS program, where the pipeline was extracting a spectrum for ~all sources in each exposure. This was leading to separate cal.fits and i2d.fits files for each source, which caused the number of files for a given observation to be ~15,000-20,000. When we then queried the filenames for all observations, the result was over the 50,000 limit. This update should take care of that situation. At last check, there were no other programs where this was a problem. The pipeline has also changed its default behavior to only extract the brightest 100 sources from each exposure, so this shouldn't be a problem for future programs either.

This method of increasing the Mast limit was suggested by @tomdonaldson

### DIFF
--- a/jwql/jwql_monitors/monitor_mast.py
+++ b/jwql/jwql_monitors/monitor_mast.py
@@ -26,7 +26,7 @@ from bokeh.embed import components
 from bokeh.io import save, output_file
 import pandas as pd
 
-from jwql.utils.constants import JWST_INSTRUMENT_NAMES, JWST_DATAPRODUCTS
+from jwql.utils.constants import JWST_INSTRUMENT_NAMES, JWST_DATAPRODUCTS, MAST_QUERY_LIMIT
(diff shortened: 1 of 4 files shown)


bhilbert4:
Just need to give this a test on the test server using a run of archive_database_update.py

bhilbert4:
Ran archive_database_update.py on the test server, which completed successfully. Also manually queried MAST for all files for program 1187. It returned 117,811 filenames in a fraction of a second. So I think this PR is all set, and ready for review. @mfixstsci

mfixstsci (approved):
@bhilbert4 this looks good to me!

PR #1052, opened 2022-10-05
Initial session update with "rough" sorting

### LINKED ISSUE #1038: Sort Thumbnails and Proposals for Next & Previous Buttons
When using the previous and next buttons on the proposal and thumbnail views, they are not sorted and stored in the same way they are displayed. Currently we only sort alphabetically based on filename, but if a use selects to sort by `exptime` and then selects a thumbnail, if they click next or previous, the thumbnail displayed in the app would not be the image the user expects and are still sorted only by filename.

To fix this, we need to add a solution where the sort parameter from the web app is passed to the view and image list is sorted according to that.

BradleySappington:
making the list used in previous/next buttons reflect the order of the thumbnails shown on the proposal page

### DIFF
--- a/jwql/website/apps/jwql/static/js/jwql.js
+++ b/jwql/website/apps/jwql/static/js/jwql.js
@@ -819,8 +819,9 @@ function update_thumbnail_array(data) {
  * @param {String} observation - The observation number within the proposal (e.g. "001")
  * @param {List} observation_list - List of all observations in this proposal
  * @param {String} base_url - The base URL for gathering data from the AJAX view.
+ * @param {String} sort - Sort method string saved in session data image_sort
  */
-function update_thumbnails_per_observation_page(inst, proposal, observation, observation_list, base_url) {
+function update_thumbnails_per_observation_page(inst, proposal, observation, observation_list, base_url, sort) {
     $.ajax({
         url: base_url + '/ajax/' + inst + '/archive/' + proposal + '/obs' + observation + '/',
         success: function(data){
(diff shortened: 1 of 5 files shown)


jwql/website/apps/jwql/static/js/jwql.js
```diff
@@ -469,23 +469,27 @@ function sort_by_proposals(sort_type) {
...
-    } else if (sort_type == 'Default') {
-        tinysort(thumbs, {selector: 'img', attr:'id'});
-    } else if (sort_type == 'Exposure Start Time') {
-        tinysort(thumbs, {attr:'exp_start'});
+    if (sort_type == 'Ascending') {
+        tinysort(thumbs, {attr:'file_root', order:'asc'});
+    } else if (sort_type == 'Descending') {
+        tinysort(thumbs, {attr:'file_root', order:'desc'});
     }
+    $.ajax({
```
mfixstsci, line 487: @BradleySappington this is clever. Have you seen what happens where the is a failure? Is anything reported to the page for the user to see?

jwql/website/apps/jwql/views.py
```diff
@@ -342,7 +340,7 @@ def archive_thumbnails_ajax(request, inst, proposal, observation=None):
     inst = JWST_INSTRUMENT_NAMES_MIXEDCASE[inst.lower()]
 
     data = thumbnails_ajax(inst, proposal, obs_num=observation)
-
+    data['thumbnail_sort'] = request.session.get("image_sort", "Ascending")
```
mfixstsci, line 343: @BradleySappington is this an example of storing session data like we discussed in the past? Also, does this set the default for the thumbails to be in Ascending order?

mfixstsci (approved):
Hey @BradleySappington the code looks good, have a couple of questions! I need to test the changes on the server as well.

jwql/website/apps/jwql/static/js/jwql.js
BradleySappington: It will report it only in the console, so if you have the console up on your terminal you will see it.  I didn't think a session not-updating is something that the user should be aware of.

jwql/website/apps/jwql/views.py
BradleySappington: Yes and Yes

PR #1054, opened 2022-10-17
Proposal thumbnail search includes only proposal subdirectories

bhilbert4:
When creating proposal-level thumbnails, generate_proposal_thumbnails.py globs over the thumbnail system directory. However, it was searching for everything in this directory using a simple '*'. This PR updates the search to look only for 'jw*' entries, so that only the subdirectories containing the thumbnails are found. Logs from previous runs show that it was finding the text files in the same directory. This wasn't causing any real problem, since the next step is to search for files within each of the glob results, but it was leading to confusing log entries that seemed to indicate it was having trouble with the thumbnail list files.

### DIFF
--- a/jwql/jwql_monitors/generate_proposal_thumbnails.py
+++ b/jwql/jwql_monitors/generate_proposal_thumbnails.py
@@ -44,7 +44,7 @@ def generate_proposal_thumbnails():
     """The main function of the ``generate_proposal_thumbnails`` module.
     See module docstring for further details."""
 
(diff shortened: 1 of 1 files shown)


bhilbert4:
@mfixstsci simple tweak. Ready for review.

mfixstsci (approved):
Hey @bhilbert4 this looks good, I am always weary of globs with `.` or `*` as the search criteria. Looks good to me.

PR #1056, opened 2022-10-21
Fixing having PID as an integer rather than a string

york-stsci:
Single minor change.


mfixstsci (approved):
LGTM

mfixstsci:
Merging besides test failure, MAST server is backed up and it has nothing to do with this change.

PR #1057, opened 2022-10-22
added back type casting because otherwise we get int64

york-stsci:
Because the CR monitor has been failing to add to the database with exactly that error. Sadly.


bhilbert4 (approved):
LGTM.

PR #1058, opened 2022-10-24
Cr database bigint

york-stsci:
Currently the cosmic ray monitor database stores cosmic ray magnitudes as follows:

- A histogram covering -65535 to +65535, showing the number of cosmic rays found at each magnitude
- An integer array of outliers (cosmic rays with magnitudes larger than the histogram boundaries)

However, apparently some cosmic rays have magnitudes that are too large to fit into a 32-big signed integer (i.e. magnitude > 2 billion), so trying to store them generates "integer out of range" errors. This fixes that issue.

*NOTE* that the cosmic ray databases on production will need to be reset after this PR is approved, because the database definition has changed from array-of-integer to array-of-bigint

*NOTE* that, prior to this, the database_interface.py file didn't import the BigInteger definition from SQLAlchemy, so it had to be added to the import statements, and to the column mappings.

### DIFF
--- a/jwql/database/database_interface.py
+++ b/jwql/database/database_interface.py
@@ -63,6 +63,7 @@ import os
 import socket
 import pandas as pd
 
+from sqlalchemy import BigInteger
 from sqlalchemy import Boolean
 from sqlalchemy import Column
 from sqlalchemy import create_engine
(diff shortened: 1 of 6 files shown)


mfixstsci (approved):
LGTM

PR #1059, opened 2022-10-25
Created a clone_tables script

york-stsci:
This script allows you to dump a database table to pandas, or to dump a pandas dataframe (stored on disk as a file whose name matches the name of a database table) into the corresponding database table. This can be used to populate test or dev database tables from production, and could also (at least potentially) dump out a table's contents before modifying the table if, for some reason (e.g. anomaly tables) you don't want to lose the existing data.


mfixstsci (approved):
This looks like it will come in handy! Thank you for doing this @york-stsci

PR #1061, opened 2022-10-26
Add function to remove all but rate thumbnails

bhilbert4:
Add a function that will delete thumbnail files for all but the rate files. This will help clean up the thumbnail filesystem, and ensure that proposal-level thumbnails are correct.


mfixstsci (approved):
Looks good, I know we only make thumbnails from rate files but if for some reason we need to change the extension, this function has more flexibility. Thank you!

PR #1064, opened 2022-10-26
Allow thumbnail files with multiple substrings to be kept

bhilbert4:
This makes a modification to the function introduced by #1061. With this update, the function will delete all thumbnails whose names do not contain any of a list of substrings. Previously only a single substring was allowed.

Now we can keep thumbnails for "rate" and "dark" files. The thumbnails from the dark files is our one exception to the rule of making thumbnails only for rate files. That's because dark files don't get processed through the pipeline far enough to create rate files.

### DIFF
--- a/jwql/utils/utils.py
+++ b/jwql/utils/utils.py
@@ -599,7 +599,7 @@ def delete_non_rate_thumbnails(extensions=['_rate_', '_dark']):
     Parameters
     ----------
     extension : list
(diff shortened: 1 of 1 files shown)


bhilbert4:
@mfixstsci @BradleySappington a small update to the function we merged earlier today. This is ready for review.

mfixstsci (approved):
Hey @bhilbert4 these changes look good to me. Good catch including the dark file thumbnails, totally forgot about those.

PR #1066, opened 2022-10-27
Added chunked iteration to CR monitor

york-stsci:
- Added a "grouper" function to jwql.utils.utils.py that breaks an iterable into chunks
- Using that function to break cosmic ray monitor file lists into chunks of 100 files.

### DIFF
--- a/jwql/utils/utils.py
+++ b/jwql/utils/utils.py
@@ -29,6 +29,7 @@ References
 
 import getpass
 import glob
+import itertools
 import json
 import pyvo as vo
 import os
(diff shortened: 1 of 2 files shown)


jwql/utils/utils.py
```diff
@@ -625,3 +626,31 @@ def query_unformat(string):
     unsplit_string = string.replace(" ", "_")
 
     return unsplit_string
+
+
+def grouper(iterable, chunksize):
```
mfixstsci, line 630: This looks great!

mfixstsci (approved):
@york-stsci this looks really good, thank you so much for your additions. The split screen view is a little crazy (probably because of indents) but I think I get the idea, you nested loops to generate the chunks, loop through the files in the chunk, add to the database, move onto the next chunk?

jwql/utils/utils.py
york-stsci: Yes, that's exactly what I'm doing.

PR #1067, opened 2022-10-27
Removed the initial copy of all input data to the internal directory

york-stsci:
So, looking at the iteration of the CR monitor that I was running, I missed a copy iteration. Before even running `process()`, the monitor copies all of the raw data off of central storage, so it was *still* trying to create a local copy of ever single file before it started iterating at all. Since even the raw files are many terabytes now, that will cause the same problems.

This fixes that issue.


mfixstsci (approved):
LGTM

PR #1068, opened 2022-10-27
Fixed the input file name

york-stsci:
Maybe the last one this time?

Fixed that I'm no longer copying the files locally, and thus should no longer give the monitor the (nonexistent) list of local files to look at.


mfixstsci (approved):
👍

PR #1069, opened 2022-10-27
Now explicitly take the first bunch of characters from the file name …

york-stsci:
…as a subdirectory, instead of assuming the file's path is fixed

Another fix.


mfixstsci (approved):
For real this time! Looks good

PR #1070, opened 2022-10-28
Multi entry model fix

BradleySappington:
Fix bug causing archive database update to save incorrectly and crash on failure.

### DIFF
--- a/jwql/website/apps/jwql/archive_database_update.py
+++ b/jwql/website/apps/jwql/archive_database_update.py
@@ -290,13 +290,17 @@ def update_database_table(instrument, prop, obs, thumbnail, obsfiles, types, sta
     # Get all unsaved root names in the Observation to store in the database
     nr_files_created = 0
     for file in obsfiles:
-        root_file_info_instance, rfi_created = RootFileInfo.objects.get_or_create(root_name=file,
-                                                                                  instrument=instrument,
-                                                                                  obsnum=obs_instance,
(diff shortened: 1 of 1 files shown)


BradleySappington:
@mfixstsci @bhilbert4 Ready for review!

jwql/website/apps/jwql/archive_database_update.py
```diff
@@ -119,7 +119,7 @@ def get_updates(inst):
                 latest_date = np.max(all_end_dates)
 
                 # Get the number of files in the observation
-                propobs = f'{proposal}{obsnum}'
+                propobs = f'jw{int(proposal):05}{obsnum}'
```
bhilbert4, line 122: Ah. It was a missing leading 0 that was causing the trouble?

bhilbert4 (approved):
This looks good to me. Maybe in another PR we should create a tool that looks through the log files for this exception, so that we don't miss it if it occurs?

jwql/website/apps/jwql/archive_database_update.py
BradleySappington: it checked for the string, but didn't verify the positioning,  so the string that it checked for existed in 2 files.  One where it was correctly the proposal and obs number, and another one where it was the end of the proposal, the obsnum, and the beginning of the visit number.
I'm now making sure we are checking from the very front of the name.

PR #1074, opened 2022-10-31
Created pipeline-specific environment file

york-stsci:
This creates an environment file specifically for pipeline calibrations (and monitors using them), as well as a shell script to create the environment (needed because the environment is at least partly remote, in that it uses JWST pipeline environment reference files that are only available as URLs), and (finally) creates a text file whose purpose is to hold the version of the JWST pipeline that should be used here.


bhilbert4:
@york I'm assuming we can close this now that we have the jwql-current environment tied to releases, yes?

PR #1076, opened 2022-11-02
Make log from generate_preview_images less verbose

bhilbert4:
This makes some adjustments to generate_preview_images.py to cut down on the information that is placed in the log. Rather than making a note for each file that a new preview image was created (or that an existing preview image was skipped over), the log simply tallies the number of new and the number of skipped preview images for a given program, and reports those numbers. 

This should make the log file more readable, and have it take up less space on disk.

### DIFF
--- a/jwql/jwql_monitors/generate_preview_images.py
+++ b/jwql/jwql_monitors/generate_preview_images.py
@@ -709,11 +709,13 @@ def process_program(program, overwrite):
     logging.info('Found {} filenames'.format(len(filenames)))
     logging.info('')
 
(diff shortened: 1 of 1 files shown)


mfixstsci (approved):
Thanks @bhilbert4 so now we will only see information for failures when debugging now? The only information displayed in the logs for success is where the pngs are written to?

bhilbert4:
We will still see log entries for failures. See line 776. We just won't see a separate status log entry for each file, unless the script is run manually and the logger is adjusted to add entries at the debug level.

PR #1077, opened 2022-11-02
Fixed cosmic ray monitor chunk iteration

york-stsci:
This fixes the cosmic ray iteration by iterating through the list of files that were just calibrated, rather than the original list, after calibration has happened.


mfixstsci:
@york-stsci Good catch here. Waiting for the CI to pass

mfixstsci (approved):
LGTM

PR #1079, opened 2022-11-03
Update Common Monitor Scripts

york-stsci:
This updates the common instrument monitor scripts (bad pixel, bias, cosmic ray, dark, read noise) to be more tolerant of some issues that can occur when processing files to generate references:

- Some files may not be processed because they have >1000 groups
- Some files may fail to process for unrelated reasons
- Some files may already have calibrated versions available locally, and so may not need to be calibrated again

In general, these monitor versions retain the current scripts, with the following changes:

- files that fail to calibrate are removed from the list that will be added to the database (or used to create a reference file)
- If calibrated data already exists locally, then the file will not be sent as a pipeline task
- If the monitor has a threshold number of files to produce output, then the threshold value is checked both before pipeline tasks start and after failed calibrations have been removed.

### DIFF
--- a/jwql/instrument_monitors/common_monitors/bad_pixel_monitor.py
+++ b/jwql/instrument_monitors/common_monitors/bad_pixel_monitor.py
@@ -743,7 +743,7 @@ class BadPixels():
                 parameters['CHANNEL'] = 'SHORT'
         return parameters
 
-    def process(self, illuminated_raw_files, illuminated_slope_files, dark_raw_files, dark_slope_files):
+    def process(self, illuminated_raw_files, illuminated_slope_files, flat_file_count_threshold, dark_raw_files, dark_slope_files, dark_file_count_threshold):
         """The main method for processing darks.  See module docstrings
         for further details.
 
@@ -773,9 +773,9 @@ class BadPixels():
             For cases where a raw file exists but no slope file, the
             slope file should be ``None``
         """
+
         # Illuminated files - run entirety of calwebb_detector1 for uncal
         # files where corresponding rate file is 'None'
-        all_files = []
         badpix_types = []
         badpix_types_from_flats = ['DEAD', 'LOW_QE', 'OPEN', 'ADJ_OPEN']
         badpix_types_from_darks = ['HOT', 'RC', 'OTHER_BAD_PIXEL', 'TELEGRAPH']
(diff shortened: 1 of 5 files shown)


jwql/instrument_monitors/common_monitors/bad_pixel_monitor.py
```diff
@@ -794,19 +794,42 @@ def process(self, illuminated_raw_files, illuminated_slope_files, dark_raw_files
                     copy_files([uncal_file], self.data_dir)
                     if hasattr(self, 'nints') and self.nints > 1:
                         out_exts[short_name] = ['jump', '1_ramp_fit']
-                    in_files.append(local_uncal_file)
+                    needs_calibration = False
+                    for file_type in out_exts[short_name]:
```
mfixstsci, line 798: This is a good check, thanks for adding this.

jwql/instrument_monitors/common_monitors/bad_pixel_monitor.py
```diff
@@ -894,16 +940,16 @@ def process(self, illuminated_raw_files, illuminated_slope_files, dark_raw_files
...
-        logging.info("\tflat_slope_files are: {}".format(illuminated_slope_files))
-        logging.info("\tdead__search_type={}".format(dead_search_type))
-        logging.info("\tflat_mean_normalization_method={}".format(flat_mean_normalization_method))
-        logging.info("\tdead_flux_check_files are: {}".format(dead_flux_files))
-        logging.info("\tdark_slope_files are: {}".format(dark_slope_files))
-        logging.info("\tdark_uncal_files are: {}".format(dark_raw_files))
-        logging.info("\tdark_jump_files are: {}".format(dark_jump_files))
-        logging.info("\tdark_fitopt_files are: {}".format(dark_fitopt_files))
-        logging.info("\toutput_file={}".format(output_file))
+#         logging.info("Calling bad_pixel_mask.bad_pixels")
```
mfixstsci, line 943: Did you intend to leave this commented out? Can it be removed if yes?

mfixstsci (approved):
Thanks @york-stsci these changes look good. I like the extra checks and seeing the extensions are a little bit more flexible.

jwql/instrument_monitors/common_monitors/bad_pixel_monitor.py
york-stsci: It can probably be removed. The reason I had it in place initially is that, whenever the bad pixel monitor crashed on making its file, I always wanted to do a re-run with the logging enabled, so I could see what files it thought it was calibrating, whether they all actually existed, and whether there were issues with any of them. It just seemed like a lot of extra text to add on occasions where the bad pixel file creation wasn't passing.

Thinking of it, I guess I'm happier if the log always prints out that diagnostic than if it never does, because the reference file creation functions can be a bit of a black box, and sometimes figuring out what's going wrong in the depth of them is difficult to trace.

PR #1080, opened 2022-11-03
Add Logging View

mfixstsci:
Adding page that will allow users to view logs interactively in the application.


mfixstsci:
@BradleySappington or @bhilbert4 this is ready for a review when ready!

jwql/website/apps/jwql/views.py
```diff
@@ -746,6 +747,50 @@ def jwqldb_table_viewer(request, tablename_param=None):
...
+    if 'dljwql' in full_uri:
+        server = 'dev'
+    elif 'tljwql' in full_uri:
+        server = 'test'
+    else:
+        server = 'ops'
+
+    full_log_paths = sorted(glob.glob(os.path.join(log_path, server, '*', '*')), reverse=True)
+    full_log_paths = [log for log in full_log_paths if not os.path.basename(log).startswith('.')]
+    log_dictionary = {os.path.basename(path): path for path in full_log_paths}
```
bhilbert4, line 780: Dictionary comprehension! I had forgotten that was a thing. :)

jwql/website/apps/jwql/views.py
```diff
@@ -746,6 +747,50 @@ def jwqldb_table_viewer(request, tablename_param=None):
...
+        Outgoing response sent to the webpage
+    """
+
+    template = 'log_view.html'
+    log_path = get_config()['log_dir']
+    log_name = request.POST.get('log_submit', None)
+
+    full_uri = request.build_absolute_uri()
+
+    if 'dljwql' in full_uri:
```
bhilbert4: Do you think this is something that would be useful to have as its own general function somewhere, or is there not enough use for it to warrant a separate function?

jwql/website/apps/jwql/views.py
```diff
@@ -746,6 +747,50 @@ def jwqldb_table_viewer(request, tablename_param=None):
...
+        server = 'dev'
+    elif 'tljwql' in full_uri:
+        server = 'test'
+    else:
+        server = 'ops'
+
+    full_log_paths = sorted(glob.glob(os.path.join(log_path, server, '*', '*')), reverse=True)
```
BradleySappington, line 778: you may need more tests to verify you are on the correct server.  "pljwql" is showing up on the test server.

bhilbert4 (approved):
Do you think we should be able to see logs from all servers regardless of what server the user is on?

BradleySappington (commented):
May be cleaner to keep javascript code in the javascript module when appropriate.

jwql/website/apps/jwql/views.py
mfixstsci: This was super useful here.

jwql/website/apps/jwql/views.py
mfixstsci: So this was causing the issue with the production logs being show on the test server. I went ahead and changed it to use the `socket` library to get the hostname of the machine rather than the URL.

jwql/website/apps/jwql/views.py
mfixstsci: I went ahead and used a different library for this portion of the app. I have it working successfully now!

PR #1081, opened 2022-11-04
Updated pipeline task

york-stsci:
This update properly checks for which output files were requested, and whether they are all present, after every pipeline step has been run. If they are, then the pipeline is halted.


mfixstsci (approved):
LGTM

PR #1078, opened 2022-11-02
[WIP]: dependency updates

bhilbert4:
Updates to our dependency files.

PR #1082, opened 2022-11-04
New try at updating 3.8 env

bhilbert4:
Trying to prevent the conflicts seen in #1078


bhilbert4:
Tests are all passing (!!) while using the values below. I'll try stepping through some of the other packages that were updated in #1079 and update those to see what happens.

  - astropy=5.1
  - bokeh=2.4.3
  - beautifulsoup4=4.11.1
  - celery=5.0.5
  - codecov=2.1.11
	@@ -33,20 +33,20 @@ dependencies:
  - ipython=8.2.0
  - jinja2=3.0.2
  - jsonschema=3.2.0
  - matplotlib=3.5.3
  - nodejs=10.13.0
  - numpy=1.21.5
  - numpydoc=1.2
  - pandas=1.4.4
  - pip=21.0.1
  - postgresql=12.2
  - psycopg2=2.8.6
  - pytest=7.1.1
  - pytest-cov=3.0.0
  - python=3.8.13
  - pyyaml=6.0
  - redis
  - scipy=1.9.1
  - setuptools=52.0.0
  - sphinx=4.2.0
  - sphinx_rtd_theme=0.4.3
	@@ -57,13 +57,13 @@ dependencies:
  - pip:
    - astroquery==0.4.6
    - bandit==1.7.4
    - jwst==1.8.2
    - pysiaf==0.18.0
    - pysqlite3==0.4.7
    - pyvo==1.4
    - redis
    - stsci_rtd_theme==1.0.0
    - stcal==1.2.1
    - git+https://github.com/spacetelescope/jwst_reffiles

bhilbert4:
Tests pass again with the versions below. So it seems like setuptools, pip, jsonschema, or nodejs is the problem.

  - astropy=5.1
  - bokeh=2.4.3
  - beautifulsoup4=4.11.1
  - celery=5.0.5
  - codecov=2.1.11
  - cryptography=38.0.1
  - django=3.2.5
  - flake8=4.0.1
  - inflection=0.5.1
  - ipython=8.4.0
  - jinja2=3.1.2
  - jsonschema=3.2.0
  - matplotlib=3.5.3
  - nodejs=10.13.0
  - numpy=1.21.5
  - numpydoc=1.4.0
  - pandas=1.4.4
  - pip=21.0.1
  - postgresql=12.2
  - psycopg2=2.8.6
  - pytest=7.1.2
  - pytest-cov=3.0.0
  - python=3.8.13
  - pyyaml=6.0
  - redis
  - scipy=1.9.1
  - setuptools=52.0.0
  - sphinx=5.0.2
  - sphinx_rtd_theme=0.4.3
  - sqlalchemy=1.4.39
  - twine=3.7.1
  - wtforms=2.3.3

  - pip:
    - astroquery==0.4.6
    - bandit==1.7.4
    - jwst==1.8.2
    - pysiaf==0.18.0
    - pysqlite3==0.4.7
    - pyvo==1.4
    - redis
    - stsci_rtd_theme==1.0.0
    - stcal==1.2.1
    - git+https://github.com/spacetelescope/jwst_reffiles

bhilbert4:
Brought django up to 4.1 and the environment is still successfully built.

bhilbert4:
Lots of conflicts after updating nodejs and jsonschema.

bhilbert4:
Env is successfully built with nodejs brought back down to 10.13.0.

bhilbert4:
So it looks like the nodejs update is the problem. All other packages other than this are now up to where they were in #1078. (in the python 3.8 env file).

bhilbert4:
@mfixstsci  this is ready for review. 

redis: has very inconsistent versions on conda vs pypi. @york-stsci reports that we need both in order for everything to work. So I didn't pin any versions there.

Should we add gunicorn to this?

environment_python_3_9.yml
```diff
@@ -21,50 +21,50 @@ channels:
   - defaults
 
 dependencies:
-  - astropy=5.0.4
-  - bokeh=2.4.2
+  - astropy=5.1
+  - bokeh=2.4.3
   - beautifulsoup4=4.11.1
-  - celery
+  - celery=4.4.0
```
mfixstsci, line 27: celery version doesn't match 3.8 environment of `5.0.5`

requirements.txt
```diff
@@ -1,33 +1,33 @@
...
-sphinx==4.4.0
+pyvo==1.4
+scipy==1.9.1
+sphinx==5.0.2
 sphinx_rtd_theme==0.4.3
-sqlalchemy==1.4.32
-stdatamodels==0.4.0
-stsci_rtd_theme==0.0.2
+sqlalchemy==1.4.39
+stdatamodels==0.4.3
```
mfixstsci, line 29: Is there a reason we need datamodels in the pip requirements but not in the conda env files?

mfixstsci (approved):
@bhilbert4 looks good, I just have a couple of questions. Thank you for doing this, I know it's tedious!

environment_python_3_9.yml
york-stsci: That's because it doesn't look like any celery version >= 5 is compatible with the way we set up the 3.9 environment. I don't know why, honestly, the dependencies here are a tangle. I don't *think* I'm using any celery features that require version 5, but it's entirely possible that we are. As such, moving to python 3.9 while keeping the full dependency list, may cause problems.

requirements.txt
bhilbert4: I'm not sure. I just kept the entry where it was in the file. Looks like 0.4.3 is present both in conda and pypi

environment_python_3_9.yml
bhilbert4: Yeah, if you look at the commit messages above, you'll see that I started with 5.0.5 and backed down one version at a time. 4.4.0 was the latest version that worked.

mfixstsci: Okay, that is strange. I am a little worried about using 2 different versions in two different environments. We normally would just default to the minimum working version for all of our dependency files. @bhilbert4 if @york-stsci doesn't think that version>=5 is necessary, should we just keep version `4.4.0` as the working version for all of our files?

bhilbert4: I guess it depends on how sure we are that we're not using any features requiring version >=5. Is there an easy way to check that? My thought is that our 3.8 environment is the most important one here, since that's what we'll have on the servers. I just want to make sure we don't break that one just for the sake of making it agree with the 3.9 environment.

york-stsci: I can run a test (and, connected to that, I can put together my own PR that updates the celery cheatsheet to cover the quick process of making a celery environment for testing.

mfixstsci: I hear you, this shouldn't be an issue we should have to deal with (pain) but I just want to make sure we aren't "looking too far ahead". For instance, if someone is developing with version>=5 locally and then we are running a version<=5 on the servers, there could be conflicts and cause crashes. As for `redis`, should we just resort to the lowest common version? How different are the versions?

york-stsci: I'm currently running some testing, and looks good so far with celery 4.4 and redis 5.0.3. I *think* that we want redis >=5 if we possibly can, but I don't remember why that's in my brain.

mfixstsci: Okay, I am cool with that since we are only using redis on the servers, as long as python can see the dependency locally to test and build the package, I don't see that being a problem.

mfixstsci:
@bhilbert4 @york-stsci Which technology has the more up-to-date versions of `redis`, `conda` or `pip`? I see we left it out in the `pip` file, so I am assuming `pip`?

bhilbert4:
conda has:
3.5.3
4.0.2
4.0.8
4.0.9
4.0.10
4.0.11
5.0.0
5.0.3

pip has:
4.2.0
4.2.1
4.3.0
4.3.1
4.3.2
4.3.3
4.3.4
4.4.0 rc 1 and 2

york-stsci:
Okay, that's definitely something I hadn't even considered. For whatever reason, the jwql 3.9 environment doesn't have the issue that the 3.8 environment has where I couldn't get redis to work without installing it from *both* conda and pip (which is good, because if you look at those lists of versions, you can see that they have no overlap at all). It's fortunate that redis is a very forgiving protocol, given that in the 3.8 environment I'm running an instance of redis 5.0.3, and talking to it via the python interface in redis 4.4.0. Let's hear it for stable protocols I guess?

bhilbert4:
To sumarize:

python 3.8:
celery 5.0.5 works. It looks like we can also run using celery 4.4.0 (??)
redis: we need to install from both conda and pypi (different versions) in order for everything to work

python 3.9:
Only celery 4.4.0 works. Anything later does not.
redis: only needs to be installed from one source (pip vs conda)

Does that sound right?

mfixstsci:
> To sumarize:
> 
> python 3.8: celery 5.0.5 works. It looks like we can also run using celery 4.4.0 (??) redis: we need to install from both conda and pypi (different versions) in order for everything to work
> 
> python 3.9: Only celery 4.4.0 works. Anything later does not. redis: only needs to be installed from one source (pip vs conda)
> 
> Does that sound right?

That sounds correct from what I remember from last week. As of right now we should probably keep running python 3.8 on the servers but before we wrap this up, @bhilbert4 how do we plan on doing all of this software installation? Do we have a practical approach or will there be a lot of added manual lever pulling? @york-stsci do you think you could send a message or find a point of contact for the redis/celery folks? Could you see if they plan on upgrading their software to be compatible with python>=3.9?

bhilbert4:
The environments on the servers should be upgraded automatically, right? At the moment we have a cron job that activates the environment and uses the environment file to make updates once per hour. Are you worried that with changes to so many packages that something will go wrong? We could always create the environment manually.

Should we add gunicorn here, or keep it a manual addition on the servers?

mfixstsci:
Yes, when we merge this I think adding gunicorn manually is the best way. The `jwql-3.8` environment that we activate, does it live in the repo or somewhere else? Because updating that manually could conflict with the cronjob pulling the repo down.

bhilbert4:
The 3.8 environment on the server does come from the repo. In the crontab we have:
```
source activate jwql-3.8 && conda env update -n jwql-3.8 -f environment_python_3_8.yml
```

gunicorn persisted in the 3.7 environment even with our hourly updates, so I assume the same would be true of the 3.8 environment.

mfixstsci:
> The 3.8 environment on the server does come from the repo. In the crontab we have:
> 
> ```
> source activate jwql-3.8 && conda env update -n jwql-3.8 -f environment_python_3_8.yml
> ```
> 
> gunicorn persisted in the 3.7 environment even with our hourly updates, so I assume the same would be true of the 3.8 environment.

Okay, perfect. Then I say this is ready to go. If you tested everything you wanted to, feel free to pull the lever.

bhilbert4:
@york-stsci @mfixstsci Do we want to keep celery 5.0.5, or back it down to 4.4.0?

mfixstsci:
@bhilbert4 Currently we are using `celery 5.0.5` on the dev server according to `conda list`. I am worried about downgrading because this could cause some issues down the road. The way I am looking at it now we have two potential options:

1. We downgrade now and have @york run the monitors to see how much stuff we break and fix it now.
2. Run with it now since we know version 5.0.5 runs with python 3.8 but will need to do the testing down the road.

My gut tells me 1 is the more work intensive route but will clear up issues once we deprecate python 3.8, but who knows, before that happens celery 5.0.5 support for python 3.9 could happen.

york-stsci:
Okay, this is what I can say:
- On the test server I was running python 3.9 and celery 4.4.0
- I'm reasonably sure that celery 4.4.0 will also work under python 3.8.
- The biggest potential source of instability under 3.8 is probably that we have to install two different instances of redis, with very different versions.
As such, I would suggest going with celery 4.4 for the moment, and trying to figure out what exactly is going on with celery 5, because it at least claims to support python 3.9 (and 3.10), so I don't think the issue is coming from python-celery, but I don't know where it *is* coming from.

mfixstsci:
@york-stsci that sounds good to me. @bhilbert4 if you could downgrade the version of celery to 4.4.0 for jwql-3.8 and we can get this in. We can revisit the versioning issue when needed.

bhilbert4:
@mfixstsci celery downgraded.

PR #1085, opened 2022-11-07
Hot Fix for JWQLDB View

### LINKED ISSUE #1083: Bad pixel monitor stats table
https://jwql.stsci.edu/jwqldb/nircam_bad_pixel_stats
The bad pixel monitor stats table page crashes, and returns:

TypeError at /jwqldb/nircam_bad_pixel_stats
jwqldb_table_viewer() got an unexpected keyword argument 'tablename_param'

mfixstsci:
When accessing the table view from monitors, we pass the table name as a url parameter. This PR will grab the parameter from the URL.


mfixstsci:
@BradleySappington or @bhilbert4 sorry to wait until Friday, but this is ready for review!

BradleySappington (commented):
Are there changes that actually call this method with a second parameter?

jwql/website/apps/jwql/views.py
```diff
@@ -594,7 +594,7 @@ def instrument(request, inst):
     return render(request, template, context)
 
 
-def jwqldb_table_viewer(request):
+def jwqldb_table_viewer(request, tablename_param=None):
```
BradleySappington, line 558: Are there changes that actually call this function with the second parameter?

jwql/website/apps/jwql/views.py
```diff
@@ -614,7 +614,10 @@ def jwqldb_table_viewer(request):
     try:
         tablename = request.POST['db_table_select']
     except KeyError:
-        tablename = None
+        if tablename_param:
+            tablename = tablename_param
```
BradleySappington, line 579: if you are sending a default of `None` then you dont need the if/else.  You can simply say:
```
except KeyError:
     tablename = tablename_param
```

BradleySappington (commented):
A few questions before approval

jwql/website/apps/jwql/views.py
mfixstsci: Yes, when you are coming from the monitor view, the tablename is passed as a url parameter which is passed to this view. But if you are accessing the JWQLDB view from the navbar/other means that aren't directly linked from a monitoring page, the default is set to `None`

BradleySappington: Great thanks!

jwql/website/apps/jwql/views.py
mfixstsci: I like this idea, if `tablename_param` is None, we still achieve the same goal here.

BradleySappington: @mfixstsci , are you making that change or should I merge as is?

BradleySappington (approved):

PR #1089, opened 2022-11-08
Fixes for crashes in the cosmic ray and bad pixel monitors

york-stsci:
Since updating the monitors so that they generally cover existing files (etc.), there have been a number of issues with particular monitors. This PR fixes the known issues.


mfixstsci (approved):
LGTM

PR #1095, opened 2022-11-10
Make scripts executable

### LINKED ISSUE #1094: Cron jobs failing because files are not executable
It looks like some cron jobs on the servers are failing because the scripts being called are not executable.

When adding new scripts, easiest is to make those files executable in the feature branch. Git should see that and keep the permissions when the repo is cloned/pulled.

For existing files, it looks like permissions can be updated like this:
https://stackoverflow.com/questions/10516201/updating-and-committing-only-a-files-permissions-using-git-version-control

bhilbert4:
Update permissions on several of our scripts such that the files are executable. Cron jobs for these scripts are failing because the files at the moment are not executable.


bhilbert4:

mfixstsci (approved):
Thanks @bhilbert4

PR #1111, opened 2022-11-16
Exposure page: Default to show rate img, keep order of radio buttons constant

### LINKED ISSUE #998: Display rate images by default for MIRI MRS
For MIRI MRS data, the auto-scale of the colorbar stretch makes it fairly difficult to see key detector features in the cal images.  In contrast, these are much easier to see with the auto-scale in the rate images.  I therefore recommend that the default displayed JWQL images for MIRI MRS data are RATE files instead of CAL files.

### LINKED ISSUE #1105: Default to rate (and dark) suffix when opening image pages
We should make sure that the app is consistent when the page for a given exposure is opened. It should default to show one extension/suffix in as many cases as possible. 

Just as is done with the thumbnails, it probably makes sense to default to showing the rate file, or the dark file for dark current observations (which do not have rate files).

bhilbert4:
Small adjustments such that the rate image is shown by default when clicking on the page for a given exposure. If there is n rate image, then it looks for a dark suffix. 

There are also changes such that we try to keep the order of the radio buttons that list the available suffixes as consistent as possible, with the only changes coming from which suffixes are present.

### DIFF
--- a/jwql/utils/constants.py
+++ b/jwql/utils/constants.py
@@ -184,6 +184,11 @@ EXPTYPES = {"nircam": {"imaging": "NRC_IMAGE", "ts_imaging": "NRC_TSIMAGE",
                        "pom": "NIS_IMAGE", "wfss": "NIS_WFSS"},
             "fgs": {"imaging": "FGS_IMAGE"}}
 
+EXPOSURE_PAGE_SUFFIX_ORDER = ['uncal', 'dark', 'trapsfilled', 'ramp', 'rate', 'rateints', 'fitopt', 'cal', 'calints',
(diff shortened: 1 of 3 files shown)


bhilbert4:
@mfixstsci This is ready for review. I've tested on the dev server and everything looks good. rate images are shown upon page loads for all images except darks, where the dark suffix is shown. I've set up some logging that creates a log only if an unknown  suffix is encountered. In my testing tonight so far, I haven't run into any instances of that.

bhilbert4:
And the radio buttons listing the suffixes now appear in a consistent order from image to image.

BradleySappington (approved):
Great comments!

PR #1092, opened 2022-11-09
Filter type

BradleySappington:
filter by exp_type

PR #1113, opened 2022-11-21
Sort obs date

### LINKED ISSUE #1110: Sort Archive page by obs date
Add the ability to sort an instrument's Archive page by the date of the most recent observation. In this way, teams can reorder the programs such that programs that contain the most recent data are at the top.

BradleySappington:
Create "Recent" sort for the archive page

NOTE:  This branch has already had "filter type" merged into it in my local workspace.


BradleySappington:
Should go in after https://github.com/spacetelescope/jwql/pull/1092

BradleySappington:
@mfixstsci @bhilbert4 - Ready for review!

bhilbert4 (approved):
This looks good to me. I checked on the dev server, and upon load, it looks like the archive page defaults to Descending order, but the button says 'Ascending'. 

The sorting choice appears to be persistent when moving from one observation page to the next, which I think makes sense. 

What do you think about 'Ascending' and 'Descending'? They mean different things at different levels. At the archive level they order by proposal number, while at the observation level they order by filename. 'Recent' implies time, which is good, but do you think it's worth changing 'Ascending' and 'Descending' to be more specific?

Similar to the exp_type sorting, it would be nice if we could eventually have the sorting choice persist when moving from the archive level to the proposal level. For example, sorting NIRCam by Recent, 1483 is listed second, but when clicking on the link for 1483, we're taken to the 1483 obs 001 page, which contains older exposures. It might be nice to be able to be taken directly to the observation with the most recent observations. Again, this seems like it might be complicated and something we can look at in a future PR.

BradleySappington:
Great feedback, we can definitely have all sorting (and most settings) be persistent.  We've already implemented sessions for some of them.  If you make an issue with your preferences and specifics I don't foresee it being too much work in all honesty.  Changing the landing page depending on ascending/descending will be the most logic work that needs to change.

bhilbert4:
What do you think about the mislabeled Archive page button, and the names 'Ascending' and 'Descending'? Are these changes we should make? And if so, do you want to add that in to this PR, or use new ones?

BradleySappington:
I would have a new PR for the mislabelled button to set the default explicitly (as there is currently not a default it seems).
If ascending descending are different on different pages, then they should be named differently somehow, it would also mean they wouldn't persist between 2 pages. (if you select ascending on one page, it shouldn't default to ascending on another as they mean different sorts)

bhilbert4:
That makes sense to me. I'll create an Issue for the default button label.

And we can think about 'Ascending' and 'Descending' in a new PR as well. My thought is that they don't mean the same thing between the Archive and Observation levels, and so there isn't a need to have them persist between the two. Or maybe we just say that the meaning is close enough between the two levels and allow the value to persist even though it's not technically the same thing.

PR #1115, opened 2022-11-28
Tweaks to mnemonics queried by the EDB monitor

### LINKED ISSUE #1116: Changes to EDB mnemonic plots from user feedback
We have received a list of notes from the MIRI team outlining desired changes to the organization and appearance of some of the MIRI plots in the EDB. The majority of these can be implemented with simple changes to the json files that list the EDB mnemonics to monitor. 

We should also remove the high-frequency mnemonics from the NIRISS page, in order to avoid having very large results files that make loading the page with the plots difficult. Down the road we should try to find a solution for this, assuming the NIRISS team is really interested in these mnemonics.

bhilbert4:
No code changes in this one. This PR changes only the details of the EDB mnemonics queried by the EDB monitor. It removes the high frequency mnemonics from the NIRISS page in order to avoid creating an extra large json file to place in the html.

It also makes adjustments to the expected/nominal values and yellow limits for a number of MIRI mnemonics. It also shifts the plots for several of the mnemonics from one tab to another. This is all in response to user feedback.

There will be some code changes to the EDB monitor coming later, also as a result of user feedback. But those will take longer to implement and are very MIRI-focused. So those will go into a separate PR. In the meantime, the instruments other than MIRI can have updated EDB plots.

### DIFF
--- a/jwql/instrument_monitors/common_monitors/edb_monitor_data/miri_mnemonics_to_monitor.json
+++ b/jwql/instrument_monitors/common_monitors/edb_monitor_data/miri_mnemonics_to_monitor.json
@@ -32,8 +32,8 @@
                 }
             ],
             "plot_data": "*SE_ZBUSVLT",
-            "nominal_value": 7.57,
-            "yellow_limits": [7.0, 8.13],
+            "nominal_value": 7.11,
+            "yellow_limits": [6.54, 7.68],
             "plot_category": "Power"
         },
         {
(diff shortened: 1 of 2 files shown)


bhilbert4:
@mfixstsci this is ready for review. No code changes. Just an update to json files. EDB code changes will come in a future PR.

BradleySappington (approved):

PR #1122, opened 2022-12-05
Edb monitor updates from MIRI feedback

### LINKED ISSUE #1123: EDB imrpovements requested by MIRI
I've been in touch with Brian O'Sullivan of the MIRI team to talk about EDB monitor imrpovements. At the moment, there is a problem with the algorithm for computing the values in the filter position ratio plots. 

He'd also like to overplot the maximum value in each block for the case of the power plots where we are currently plotting only the mean value in each block of time.

bhilbert4:
This PR makes some changes to the EDB monitor based on feedback from the MIRI team. 

1. Add the ability to store and plot the max/min/mean/median for a given menmonic, in addition to the data currently plotted. This was requested for one of the power plots where we are currently plotting the block averages (the average value for each block of the data), and the desire was to also see the max value for each block.
2. For the fliter position plots: refine the algorithm to ignore data that are obviously bad. The timing between the filter position ratio and the commanded position is not quite aligned. Often the position ratio will start to change just before or after the commanded position changes, and the telemetry value will be some intermediate value between expected positions. Also, in the middle of long stretches where the filter wheel does not move, the position ratio will sometimes go to zero. The algorithm has been modified to ignore obviously bad data by comparing all values to a list of the expected value and standard deviation for each filter. Also, the mean positions are now normalized by the expected positions (rather than the computed mean value).
3. Database table definitions have been modified, in order to store the potentially needed max/min/median/mean values. The database tables for the filter position data have also been modified, but only to change the name of the column from 'mean' to 'median', since the latter is actually what is being stored. (Means were being contaminated by bad data values).


bhilbert4:
@mfixstsci this is ready for review. Lots of changed files, but only 2 of them actually have updated code.

bhilbert4:
Oh shoot. I just saw the conflicts. I'll fix those.

bhilbert4:
@mfixstsci I re-ran the failing documentation test and it was successful this time.

jwql/instrument_monitors/common_monitors/edb_monitor_data/miri_mnemonics_to_monitor.json
```diff
@@ -1841,7 +1841,7 @@
                     "threshold": 25
                 }
             ],
-            "plot_data": "*SE_ZBUSVLT",
+            "plot_data": "*SE_ZBUSVLT,max",
```
mfixstsci, line 1844: What does the added `max` do here?

jwql/tests/test_edb_telemetry_monitor.py
```diff
@@ -282,6 +283,10 @@ def test_organize_every_change():
     mnem.every_change_values = ec_vals
     data = etm.organize_every_change(mnem)
 
+
+    print(data)
```
mfixstsci: Maybe we delete this print statement here if you are done testing @bhilbert4

mfixstsci (approved):
Would like to test this on the server first but the changes look good @bhilbert4 thank you!

mfixstsci:
Looks like we have a conflict that needs to be resolved before merging as well.

jwql/tests/test_edb_telemetry_monitor.py
bhilbert4: Done

jwql/instrument_monitors/common_monitors/edb_monitor_data/miri_mnemonics_to_monitor.json
bhilbert4: This tells the EDB monitor to plot the maximum value of the mnemonic, along with the average value

bhilbert4:
@mfixstsci conflicts resolved, comments addressed, and tests are passing!

mfixstsci (approved):
Thanks for submitting this @bhilbert4

PR #1127, opened 2022-12-07
Clarify reset_database.py statement

bhilbert4:
Just a small change to the message printed to the screen when asking the user to confirm the database reset. The instrument name and monitor name were being printed in the wrong order, which made the message confusing. I also enclosed the values in brackets, just to separate them from the rest of the text a bit more.


mfixstsci (approved):
Small change, short review. LGTM

PR #1131, opened 2022-12-12
Date Range Page

### LINKED ISSUE #1107: Show all data that were taken between 2 dates
The NIRSpec team requested a page that would show all the data that have come in in the last day, as a way to quickly see which data need to be examined. They also mentioned having a page that shows all of this data across all programs. So, sort of a 1 stop shop for new data. 

This seems like it would be useful. And in fact, it might be useful to set it up as a page that will show all data across all programs that were taken between 2 user input dates. That would be an easy way to see all data that were taken e.g. after a tilt event but before the mirrors were re-aligned.

BradleySappington:
Instrument specific page to view all thumbnails within a date range


BradleySappington:
@mfixstsci @bhilbert4 - Ready for test/review

mfixstsci:
@BradleySappington do you have this PR up on one of the servers currently? Would just like @bhilbert4 to test it and you showed it to me yesterday and it looked good.

jwql/website/apps/jwql/data_containers.py
```diff
@@ -50,8 +50,7 @@
 from jwql.utils.utils import check_config_for_key, ensure_dir_exists, filesystem_path, filename_parser, get_config
 from jwql.utils.constants import MAST_QUERY_LIMIT, MONITORS, PREVIEW_IMAGE_LISTFILE, THUMBNAIL_LISTFILE, THUMBNAIL_FILTER_LOOK
 from jwql.utils.constants import IGNORED_SUFFIXES, INSTRUMENT_SERVICE_MATCH, JWST_INSTRUMENT_NAMES_MIXEDCASE, JWST_INSTRUMENT_NAMES
-from jwql.utils.constants import JWST_INSTRUMENT_NAMES_SHORTHAND, SUFFIXES_TO_ADD_ASSOCIATION, SUFFIXES_WITH_AVERAGED_INTS
-from jwql.utils.preview_image import PreviewImage
```
mfixstsci, line 54: Were we not using these imports?

jwql/website/apps/jwql/data_containers.py
```diff
@@ -1535,6 +1534,138 @@ def thumbnails_ajax(inst, proposal, obs_num=None):
...
+                if 'detector' not in filename_dict.keys():
+                    filename_dict['detector'] = 'Unknown'
+
+                # Weed out file types that are not supported by generate_preview_images
+                if 'stage_3' in filename_dict['filename_type']:
+                    continue
+
+            except ValueError:
+                # Temporary workaround for noncompliant files in filesystem
+                filename_dict = {'activity': rootname[17:19],
```
mfixstsci, line 1585: What are the issues with these files?

bhilbert4:
@BradleySappington I just tried using nircam and dec 28-29, and it came back with >1000 images. I'm guessing it's easier to raise the limit than it would be to start breaking things down into time periods of less than a day. I don't think I've tried loading a page with >1000 images recently. Hopefully that won't take forever?

BradleySappington:
@bhilbert4, In that situation you should do 1 day instead of two (the error message should tell you to reduce your time-frame).  If we get to a situation where there are over 1000 per day we will need to adjust the limit, or enter a time function, but the site itself will crash when it gets much higher than 1000 (I think I crashed it at 1300 if I remember correctly...)

bhilbert4:
Ah, I see. I had assumed that entering a start date of Dec 28, 2022 and an end time of Dec 29, 2022 was limiting the query to 1 day (e.g. midnight on the 28th to midnight on the 29th). I just tried putting in Dec 28 for both starting and ending date and that did bring the list of exposures down under 1000. It hadn't occurred to me to enter the same date for both starting and  ending dates.

BradleySappington:
You can leave the end date empty as well and it will autofill to the start date.  Maybe I should mark that field as "optional"?  What do you think?

bhilbert4:
That seems reasonable. And/or maybe a note that explains that not filling in the end date, or entering the same starting and ending dates will get you 1 full day's worth of data?

BradleySappington:
went ahead and implemented time functionality since a single day has had over 1k 
Loaded on Dev, @bhilbert4 @mfixstsci ready for review

bhilbert4:
Looks good. My one comment is that it took me a minute to remind myself the difference between "Ascending" and "Recent". I had at first assumed Ascending and Recent would give the same order. But then I remembered that Ascending applies to proposal numbers rather than time, and the proposal numbers are not as in-your-face as they are on the Archive page.

BradleySappington:
@bhilbert4 Agreed, with the sorting using common names/functions I decided to keep as is (assuming you all are fine with that).  If it becomes an issue we can spend the time to update.

mfixstsci (approved):
Hey @BradleySappington thank you for submitting this and walking me through the review.

jwql/website/apps/jwql/data_containers.py
BradleySappington: Correct, I've banished them

jwql/website/apps/jwql/data_containers.py
BradleySappington: This is a failsafe that I stole from thumbnails_ajax.  We can probably discuss with @bhilbert4 if anything else should happen in situations like this.  If so, it should be in a different PR though due to scope.

PR #1135, opened 2022-12-13
Add more details to EDB documentation

### LINKED ISSUE #1101: Add EDB documentation
The EDB monitor is in, but at the moment 3 of the 5 instruments have it running with a set of test mnemonics. We need to get lists of which mnemonics they actually want to monitor. As part of that, teams need to know what options there are for monitoring. We should add documentation that describes the filtering and the various averaging schemes that are available.

bhilbert4:
Flesh out the documentation on the EDB monitor a bit, in order to explain the format and use of the JSON file that goes into specifying which mnemonics to monitor.

Resolves #1101

### DIFF
--- a/jwql/website/apps/jwql/views.py
+++ b/jwql/website/apps/jwql/views.py
@@ -74,7 +74,8 @@ from .data_containers import thumbnails_ajax
 from .data_containers import thumbnails_query_ajax
 from .forms import AnomalyQueryForm
 from .forms import FileSearchForm
-from .models import Observation, Proposal, RootFileInfo
+if not os.environ.get("READTHEDOCS"):
+    from .models import Observation, Proposal, RootFileInfo
 from astropy.io import fits
 
 
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -1,5 +1,6 @@
 # pyproject.toml
 [build-system]
 requires = ["setuptools<=65.5.0",
+		    "numpy",
 			"wheel",
 			"setuptools_scm"]
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -1,0 +1,5 @@
+# pyproject.toml
+[build-system]
+requires = ["setuptools<=65.5.0",
+			"wheel",
+			"setuptools_scm"]
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -1,6 +1,6 @@
 # pyproject.toml
 [build-system]
 requires = ["setuptools<=65.5.0",
-		    "numpy",
+			"numpy",
 			"wheel",
 			"setuptools_scm"]
--- a/rtd_requirements.txt
+++ b/rtd_requirements.txt
@@ -1,3 +1,4 @@
+bokeh==2.4.3
 cython==0.29.28
 django==3.1.7
 docutils==0.18.1
--- a/rtd_requirements.txt
+++ b/rtd_requirements.txt
@@ -1,9 +1,9 @@
 cython==0.29.28
-django==3.2.5
+django==4.1
 docutils==0.18.1
 flake8==4.0.1
 jwst==1.4.3
-jwql==1.0.0
+jwql==1.1.1
 pygments==2.11.2
 pytest==7.0.1
 sphinx>=2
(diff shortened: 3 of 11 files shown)


bhilbert4:
Hurray! RTD docs are now being built successfully! I get the feeling they hadn't been built for quite a while. Not that this PR still keeps fail_on_warnings set to False in .readthedocs.yml. That's because one of the warnings that was popping up was a decprecation warning in one of our dependencies. So it is possible we will run into this case again, where the RTD build is reported as passing, but is in fact failing. We'll have to keep a close eye on any PRs that update documentation to check for this.

bhilbert4:
@mfixstsci this is ready for review. No code changes other than ignoring a couple of import statements when on RTD. The rest of the changes are documentation-related.

rtd_requirements.txt
```diff
@@ -1,11 +1,14 @@
+sphinx_automodapi==0.14.1
+bokeh==2.4.3
+celery==4.4.0
 cython==0.29.28
-django==3.2.5
+django==3.1.7
```
mfixstsci, line 5: Is there a reason we downgraded django here?

setup.py
```diff
@@ -15,7 +15,7 @@
     'astropy>=3.2.1',
     'astroquery>=0.3.9',
     'bandit',
-    'bokeh',
+    'bokeh==2.4.3',
```
mfixstsci, line 18: I know why we are pinning the bokeh version here but we should see if we can update the panels and other calls here.

jwql/website/apps/jwql/views.py
```diff
@@ -74,7 +74,8 @@
 from .data_containers import thumbnails_query_ajax
 from .forms import AnomalyQueryForm
 from .forms import FileSearchForm
-from .models import Observation, Proposal, RootFileInfo
+if not os.environ.get("READTHEDOCS"):
```
mfixstsci, line 77: So when we build the docs, we don't want to try and import these modules. Why?

pyproject.toml
```diff
@@ -0,0 +1,6 @@
+# pyproject.toml
+[build-system]
+requires = ["setuptools<=65.5.0",
```
mfixstsci, line 3: The spacing looks a little funky here, is there a reason or should they fall on the same column?

mfixstsci (approved):
Awesome work @bhilbert4 the edb monitor docs are thorough but still short. Good catches on typos and other updates!

jwql/website/apps/jwql/views.py
bhilbert4: Because these modules try to get information from the config file, which is not present on RTD.

pyproject.toml
bhilbert4: I copied this from Mirage without thinking too much about it. Maybe when I pasted the lines, the indentation was screwed up.

jwql/website/apps/jwql/views.py
bhilbert4: There was some other similar reason as well. Trying to access something that RTD didn't have access to.

setup.py
bhilbert4: Set them equal to what they are in the environment files?

rtd_requirements.txt
bhilbert4: Downgraded to match what is in setup.py

pyproject.toml
bhilbert4: Strange. The alignment in my local copy of the file is fine.

bhilbert4: Deleted some spaces, added some spaces. Now everything looks well-aligned.

setup.py
mfixstsci: So bokeh is currently at v3.0+ but of course some of the modules/methods have moved around. Particularly the `Panel` which caused some issues for `Exoctk`. I figured you were pinning it to avoid new versions from installing.

jwql/website/apps/jwql/views.py
mfixstsci: Awesome, thanks. I am a novice when it comes to read the docs.

pyproject.toml
mfixstsci: That is weird haha, thanks for fixing it!

PR #1137, opened 2022-12-19
Update actions version to move from node12 to node16

bhilbert4:
A test from a previous PR showed the warning:
Node.js 12 actions are deprecated. For more information see: https://github.blog/changelog/2022-09-22-github-actions-all-actions-will-begin-running-on-node16-instead-of-node12/. Please update the following actions to use Node.js 16: actions/checkout@v2

blog entry here: https://github.blog/changelog/2022-09-22-github-actions-all-actions-will-begin-running-on-node16-instead-of-node12/

This PR makes the requeted update to the actions version

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -15,9 +15,9 @@ jobs:
         python-version: [3.8, 3.9]
 
     steps:
-      - uses: actions/checkout@v2
+      - uses: actions/checkout@v3
 
-      - uses: conda-incubator/setup-miniconda@v2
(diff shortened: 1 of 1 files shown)


bhilbert4:
Tests are failing because they can't reach the MAST server. I'll try again in a bit.

bhilbert4:
Tests are passing, and the warning about needing to switch to node16 is gone. https://github.com/spacetelescope/jwql/actions/runs/3732554173

mfixstsci (approved):
Good catch, looks like I should update my builds on my projects too.

PR #1159, opened 2023-02-01
[WIP]: Set django env variable in data_containers

bhilbert4:
When running "pytest test_data_containers.py" locally, the script crashes because the DJANGO_SETTINGS_MODULE environment variable is not defined, and data_continers.py (which test_data_containers.py imports) imports the Django models.

This PR adds the environment variable definition to data_containers.py. This allows the tests to run, and should allow other calls to data_containers.py when running the web app locally.

It made more sense to me to add the definition to data_containers.py rather than test_data_containers.py in order to support other scripts that may import data_containers.py.

- [ ] Test on one of the servers to make sure this does no harm


bhilbert4:
Resolved via #1167

PR #1167, opened 2023-02-06
Unit test cleanup

melanieclarke:
Fixes a number of issues with running unit tests locally against central storage data. Some tests require a running local server instance prior to starting the test suites; they will be skipped if the server is not found.


.gitignore
```diff
@@ -3,6 +3,8 @@
 *.ipynb_checkpoints/
 config.json
 .DS_Store
+.idea
```
mfixstsci, line 6: @melanieclarke Are `.idea` and `.coverage` local directories to your machine?

jwql/database/clone_tables.py
```diff
@@ -78,7 +80,7 @@
         active_tables = INSTRUMENT_TABLES[instrument]
     elif args.monitor is not None:
         monitor = args.monitor.lower()
-        active tables = MONITOR_TABLES[monitor]
+        active_tables = MONITOR_TABLES[monitor]
```
mfixstsci, line 83: Good catch 👍

jwql/tests/test_api_views.py
```diff
@@ -42,23 +42,32 @@
...
-             '01106']  # NIRSpec
+
+# TODO: currently missing in central storage
+# proposals = ['01018',  # FGS
+#              '01022',  # MIRI
+#              '01068',  # NIRCam
+#              '01059',  # NIRISS
+#              '1059',  # NIRISS
+#              '01106']  # NIRSpec
+proposals = ['2640', '02733', '1541', '02589']
```
mfixstsci, line 45: I see we have 4 proposals included here. @bhilbert4 proposals beginning with `0` are pre-flight or commissioning data, right? Do we want to test these proposals or should we stick to in flight data? Also, do we want to make sure we include enough proposals to test test each instrument as well?

jwql/tests/test_api_views.py
```diff
@@ -42,23 +42,32 @@
...
-             'jw01059121001_02107_00001_nis',  # NIRISS
-             'jw01106002001_02101_00002_nrs1']  # NIRSpec
+
+# TODO: currently missing in central storage
+#rootnames = ['jw01018001004_02101_00001_guider1',  # FGS
+#             'Jjw1022017001_03101_00001_mirimage',  # MIRI
+#             'jw01068001001_02102_00001_nrcb1',  # NIRCam
+#             'jw01059121001_02107_00001_nis',  # NIRISS
+#             'jw01106002001_02101_00002_nrs1']
+rootnames = ['jw02733002001_02101_00001_mirimage',  # MIRI
```
mfixstsci, line 53: same here -- do we want to include enough files to make sure we are testing each instrument?

jwql/tests/test_cosmic_ray_monitor.py
```diff
@@ -51,7 +51,9 @@ def define_test_data(nints):
         rate_data = np.ones((2, 10, 10))
 
     filesystem = get_config()['filesystem']
-    filename = os.path.join(filesystem, 'public', 'jw00623', 'jw00623087001', 'jw00623087001_07101_00001_mirimage_rate.fits')
+    #filename = os.path.join(filesystem, 'public', 'jw00623', 'jw00623087001', 'jw00623087001_07101_00001_mirimage_rate.fits')
```
bhilbert4: I wouldn't spend much time thinking about the CR monitor at the moment. The MIRI team decided that what we had was not useful, and they are currently working on deciding what exactly it should be doing. So these tests are not relevant. Maybe easiest to just delete them?

jwql/tests/test_data_containers.py
```diff
@@ -128,7 +128,7 @@ def test_get_preview_images_by_proposal():
 def test_get_preview_images_by_rootname():
     """Tests the ``get_preview_images_by_rootname`` function."""
 
-    preview_images = data_containers.get_preview_images_by_rootname('jw01068001001_02102_00001_nrcb3')
+    preview_images = data_containers.get_preview_images_by_rootname('jw01033006001_gs-fg_2022149123539')
```
bhilbert4: I think ideally we should stay away from guide star preview images. (Here gs-fg is "guide star - fine guide) We stopped creating preview images for these a while back, mainly because each observation has so many guide star products with so many frames that they were taking up a ton of disk space and no one was looking at them. We deleted them from the servers, but obviously missed those that thad been moved to central store.

jwql/tests/test_data_containers.py
```diff
@@ -160,7 +160,7 @@ def test_get_thumbnails_by_proposal():
 def test_get_thumbnails_by_rootname():
     """Tests the ``get_thumbnails_by_rootname`` function."""
 
-    preview_images = data_containers.get_thumbnails_by_rootname('jw01068001001_02102_00001_nrcb3')
+    preview_images = data_containers.get_thumbnails_by_rootname('jw01033006001_gs-fg_2022149123539')
```
bhilbert4: Same comment about guide star products here

jwql/tests/test_msata_monitor.py
```diff
@@ -305,13 +305,14 @@ def test_plt_mags_time():
 
 
 @pytest.mark.skipif(ON_GITHUB_ACTIONS, reason='Requires access to central storage.')
-def test_mk_plt_layout():
+def test_mk_plt_layout(tmp_path):
```
bhilbert4: Was this not working with the output path from the config file?

jwql/tests/test_api_views.py
```diff
@@ -90,10 +99,8 @@ def test_api_views(url):
 
     try:
         url = request.urlopen(url)
-    except error.HTTPError as e:
-        if e.code == 502:
-            pytest.skip("Server problem")
-        raise(e)
+    except (error.HTTPError, error.URLError):
```
mfixstsci, line 87: So the HTTP Error makes sense if the server is down or we can't connect, what does the adding the URL error catch here? Do we expect to build URLs that might not exist on the app in this test?

jwql/website/apps/jwql/tests/test_context_processors.py
```diff
@@ -31,6 +31,7 @@
     from jwql.utils.utils import get_base_url
 
 
+@skipIf(True, "Needs updating: throws errors for missing settings and apps")
```
bhilbert4: I got this test to work locally by removing "users" from the list of keys, and by adding:
```
# These lines are needed in order to use the Django models in a standalone
# script (as opposed to code run as a result of a webpage request). If these
# lines are not run, the script will crash when attempting to import the
# Django models in the line below.
from django import setup
os.environ.setdefault("DJANGO_SETTINGS_MODULE", "jwql.website.jwql_proj.settings")
setup()
```

bhilbert4 (commented):
Very nice! Getting all tests to work locally is something we've neglected for way too long. I added a few specific comments. My only general comments are: 1) if the tests are working with the updated filenames in the tests, then feel free to delete the old names, and 2) only tangentially related, but I created #1168 to ask if we should bring more data over to the central store filesystem to make it more useful.

I like the updates. I'll let @mfixstsci look things over and give the final approval.

jwql/tests/test_msata_monitor.py
melanieclarke: This one was writing to the output path in central storage, but was failing for me because a previous user had written test files there that could not be overwritten.

jwql/website/apps/jwql/tests/test_context_processors.py
melanieclarke: Thanks! I'll make that change.

jwql/tests/test_msata_monitor.py
bhilbert4: I'll see if I can get Maria to delete the file that is currently in the directory. I think a good way around this is to use the [set_permissions](https://github.com/spacetelescope/jwql/blob/develop/jwql/utils/permissions.py#L25) function on the output file. Then it will always be owned by jwqladm and should have permissions set such that it can be overwritten. That way anyone should be able to run the tests locally.

melanieclarke: Checking now, I don't see a file there, but the test directory is not group-writeable so the test still fails.  Should the set_permissions go in the monitor code or the test?

bhilbert4: Maria said she's deleted the file. So hopefully the test will work now.

bhilbert4: I've asked her to update permissions for the directory. I think set_permissions should go in the test for sure, since that is creating this version of the file. The monitor itself is always run by the administrative accounts on the jwql servers, so in that case I think everything is ok. @mfixstsci do you agree?

jwql/tests/test_cosmic_ray_monitor.py
melanieclarke: I deleted the one test that I marked skip due to failure.  Unless you object, we can leave the others in case they're useful for later reference?

jwql/tests/test_data_containers.py
melanieclarke: Replaced with a different preview image currently in central store.

jwql/tests/test_data_containers.py
melanieclarke: Replaced with a different preview image currently in central store.

jwql/tests/test_cosmic_ray_monitor.py
```diff
@@ -169,6 +171,7 @@ def test_magnitude_fake_data():
     assert mag == 10.
 
 
+@pytest.mark.skip(reason='Needs update: API has changed')
```
mfixstsci: This makes sense here. We are currently rewriting the requirements for the CR monitor/how we calculate the stats and store the data. I am assuming we will need to update the tests when the new and improved version is implemented.

jwql/tests/test_cosmic_ray_monitor.py
bhilbert4: Sounds good to me.

jwql/tests/test_msata_monitor.py
melanieclarke: Okay, thanks.

The same issue exists for the wata_monitor - can you please have her update that file/directory too?

bhilbert4: I passed that along. @penaguerrero

melanieclarke: Thanks for the fix, Maria!

I pushed an update that writes the file to the expected directory and sets group write permission on it so it can be overwritten.

jwql/tests/test_data_containers.py
mfixstsci: For this test, I would suggest we use a different file. `gs-fg = guide star-fine guiding` we don't want to use these preview images for our testing.

jwql/tests/test_data_containers.py
mfixstsci: Same here for `gs-fg` data.

jwql/tests/test_database_interface.py
```diff
@@ -48,11 +50,14 @@ def test_all_tables_exist():
...
 
     # Get list of tables that are actually in the database
-    existing_tables = di.engine.table_names()
+    existing_tables = inspect(di.engine).get_table_names()
 
     # Ensure that the ORMs defined in database_interface actually exist
     # as tables in the database
     for table in table_orms:
+        # TODO: table doesn't exist on dev server
+        if table in ['nirspec_ta_stats']:
```
mfixstsci: @bhilbert4 some of these tables may not exist in the dev DB but when these tests get pushed to production all of the tables should be there, right?

mfixstsci (approved):
@melanieclarke thank you for submitting these changes! Just a few questions and comments.

.gitignore
melanieclarke: .idea is for PyCharm configuration, so specific to me if no one else is using that IDE.  I can ignore it locally if you prefer.

.coverage files are dropped by the coverage tool which I frequently run with pytest to check code coverage.  I think it's useful to have in .gitignore, but again I can ignore it locally if no one else uses that tool.

jwql/tests/test_api_views.py
melanieclarke: I included the URL catch in case some of the specific data referenced on central storage goes missing again.  I can let that fail instead if you prefer.

jwql/tests/test_data_containers.py
melanieclarke: Done! Bryan pointed this one out too -- I've replaced it with another randomly selected science file.

jwql/tests/test_data_containers.py
melanieclarke: Done!

jwql/tests/test_database_interface.py
bhilbert4: In general we should probably make sure that all our databases contain all the tables, but in this case, I'm not convinced that nirspec_ta_stats is actually used anywhere. I see the MSATA monitor connecting to it, but I don't see any data actually being put into it. I've asked Maria to confirm. 
So it seems like we could add the table to the dev server so that we don't need any conditional clause here, and then in a separate PR we can remove the table completely if Maria says we don't need it? What do you think?

jwql/tests/test_api_views.py
bhilbert4: I think this is a good set to test with. JWQL is designed to prepend a 0 to proposal numbers that have only 4 digits. So it should translate 2640 to 02640, and 02733 should stay as 02733. It's good to confirm that's working. In terms of real programs (commissioning and cycle 1) that have data, we're not much beyond 2733. We used to have fake data with program numbers >10000, but we probably shouldn't touch that data any more, especially because none of that is in MAST.

jwql/tests/test_api_views.py
bhilbert4: It's probably not terribly important, since the instrument name isn't part of the URLs being tested.

.gitignore
mfixstsci: Awesome, thanks. This will be useful for our other pycharm users.

jwql/tests/test_database_interface.py
mfixstsci: I agree, I think for this case we can push this through and when Maria gets back to us we can submit the PR and update this test as part of it.

bhilbert4: Maria reports that the nirspec_ta_stats table is not used at all. So I think we can add the table to the dev server and delete these two lines. Then we won't need to worry about this later, when we remove the table completely.

mfixstsci: Thanks @bhilbert4 should we just delete the table as well as remove the lines? I am just curious of where the table is being pulled from/why it's created in the first place.

melanieclarke: I've removed the lines from the test and confirmed that the test passes when run against the production database.  It currently fails when run against dev; I can check again if/when we add the table there.

PR #1174, opened 2023-02-08
Fix Bugs Causing Local Crash for Query Page

### LINKED ISSUE #1125: Create a page showing all exposures flagged with a particular anomaly
From the NIRCam team meeting:

It would be helpful to have a page that shows the thumbnails/preview images for all exposures that have been flagged with a particular anomaly. 

We have this mostly done, from the Query The Archive button on the front page of JWQL. 
But:
1) We should probably change the button name to make it more clear what it does
2) At the moment, after selecting an instrument and anomaly and hitting Submit, the page fails to load.

BradleySappington:


BradleySappington:
@bhilbert4 @mfixstsci @melanieclarke 
I have this site working to the capabilities of the original rough design.  However the site is not sustainable and will get errors if trying to query on a very broad topic (like querying for a single anomaly and no other fields).  
To fix this, a re-design is going to be required to be implemented on this page.  We have already discussed importing the external db's into the django DB.
However, that work should be associated with a separate issue/PR.

BradleySappington:
@bhilbert4 @mfixstsci @melanieclarke - Ready for review.  Currently loaded on dev server in branch "update_query_page"

jwql/website/apps/jwql/api_views.py
```diff
@@ -200,7 +200,7 @@ def thumbnails_by_proposal(request, proposal):
     return JsonResponse({'thumbnails': thumbnails}, json_dumps_params={'indent': 2})
 
 
-def thumbnails_by_rootname(request, rootname):
+def thumbnail_by_rootname(request, rootname):
     """Return a list of available thumbnails in the filesystem for the
     given ``rootname``.
```
melanieclarke: I think the docstring needs updating here - it's now returning a single thumbnail, right?

melanieclarke (commented):
Looks good to me!

Local unit tests are passing with these changes.  I tested a couple queries on production and dev and was able to get a result on dev for a search that got stuck on 'Loading' on production. 

I was not able to find a non-empty result for a local install. Do you have a suggested query that should work?

One tiny specific comment, for a docstring update.

jwql/website/apps/jwql/api_views.py
BradleySappington: aha! good catch

BradleySappington:
@melanieclarke 
> I was not able to find a non-empty result for a local install. Do you have a suggested query that should work?

I used: NIRCAM, NRCB1 FULL, F212N, NRCB1, SHALLOW2, NRC IMAGE
Leave anomaly blank, and then you can add an anomaly to your local DB from these results and re-test including the new anomaly

mfixstsci (approved):
With @melanieclarke's approval, I am going to push it through. Thank you for submitting and thank you for reviewing @melanieclarke

melanieclarke:
> @melanieclarke
> 
> > I was not able to find a non-empty result for a local install. Do you have a suggested query that should work?
> 
> I used: NIRCAM, NRCB1 FULL, F212N, NRCB1, SHALLOW2, NRC IMAGE Leave anomaly blank, and then you can add an anomaly to your local DB from these results and re-test including the new anomaly

Thanks, that worked.

PR #1196, opened 2023-02-16
Dependency updates

melanieclarke:
Updates the conda environment files and pip requirement files to the latest possible versions.

To find the latest working versions, I made a new environment file and unpinned every dependency except for a few known versions we don't support yet (we currently require bokeh<3, django<3.2, sqlalchemy<2). From this environment, I found the latest installed version and then specified it precisely in the environment_python_*.yml files.  Python 3.8 and 3.9 had existing environment files; I added one for 3.10 as well. All three have identical requirements, except for the python version specification.  All three are now able to build and install, and unit tests and spot checks a local copy of the app are passing.  There may be other small fixes needed for code not currently covered by unit tests.

There were a few inconsistencies between the pip files and the conda files and the setup.py file.  I resolved them by copying the specifications from the conda environment into pip format, then testing the install with a bare conda environment and `pip install -r requirements.txt`. All inconsistencies should be resolved now, with the exception of the psycopg2 package: the requirements.txt specifies psycopg2-binary as the prerequisite rather than psycopg2 (required by the environment files and setup.py).  This was the case before this update as well, so I have left it as is.

I have also included a couple minor fixes needed for compatibility with newer numpy versions, relating to array creation and floating point equality.  Also, I fixed a couple small documentation build errors that came up in testing the rtd_requirements.txt changes.


jwql/website/apps/jwql/views.py
```diff
@@ -1151,7 +1151,10 @@ def view_image(request, inst, file_root, rewrite=False):
             suff_arr = np.array(image_info['suffixes'])
             files_arr = np.array(image_info['all_files'])
             splits = np.array([ele.split('_')[-1] for ele in image_info['suffixes']])
-            idxs = np.where(splits == poss_suffix)[0]
+            if splits.size > 0:
```
mfixstsci, line 1154: Was this intended to be included in this PR?

jwql/website/apps/jwql/monitor_pages/monitor_dark_bokeh.py
```diff
@@ -624,7 +624,7 @@ def stats_data_to_lists(self):
         self._stdev = np.array([e.stdev for e in self.db.stats_data])
         self._obs_mid_time = np.array([e.obs_mid_time for e in self.db.stats_data])
         self._stats_mean_dark_image_files = np.array([e.mean_dark_image_file for e in self.db.stats_data])
-        self._stats_numfiles = np.array([e.source_files for e in self.db.stats_data])
+        self._stats_numfiles = np.array([len(e.source_files) for e in self.db.stats_data])
```
mfixstsci, line 627: Was this intended to be a part of this PR?

mfixstsci (approved):
Hey @melanieclarke thanks for submitting these updates!

jwql/website/apps/jwql/views.py
melanieclarke: Yes, this fixes an issue that throws an error in the unit tests with later numpy versions.

jwql/website/apps/jwql/monitor_pages/monitor_dark_bokeh.py
melanieclarke: This one too -- this is allowed under earlier numpy versions but not later ones.  I think the original form was a mistake anyway.  The _stats_numfiles is not used anywhere, but I believe it was intended to be an array of numbers, not an array of lists of filenames.

mfixstsci: Awesome, thank you for clarifying. I am always cautious when seeing code changes in non-code change related PR 😄

melanieclarke: Same! And thanks for your careful and thorough review.

PR #1199, opened 2023-02-21
Catch root file / file list mismatch

melanieclarke:
Minor fix to the view_image template to catch mismatch between file_root_list and currently displayed file.


melanieclarke:
@mfixstsci - This fix just hides the previous/next buttons in the error case.

BradleySappington (approved):
Looks good.

PR #1202, opened 2023-02-23
Improve bad pixel monitor plots

### LINKED ISSUE #1141: Strip bokeh templating out of dark and bad pixel montiors
bokeh templating is no longer maintained. We need to replace the monitors' plotting to use straight bokeh rather than bokeh templating, so that we can more easily maintain the monitors for new versions of bokeh.

### LINKED ISSUE #1200: bokeh.io.export_png may fail on server
While testing some updates to the bad pixel monitor today, I tried using bokeh.io.export_png while running on one of the witserv machines. I got the following error:

RuntimeError: Neither firefox and geckodriver nor a variant of chromium browser and chromedriver are available on system PATH. You can install the former with 'conda install -c conda-forge firefox geckodriver'.

I never saw this while testing the dark current monitor, which also uses export_png. 
We should check the environments on the jwql servers and see if we need to install this package.

bhilbert4:
Remove bokeh templating from the bad pixel monitor plots, and make some tweaks to make them more useful and faster to load.

I'm trying to make these a little more interactive than the plots from the dark monitor. In this case, the figure/plot creation will be called as the final step of the bad pixel monitor itself (or it can be called separately at any time). The output will be a separate html file for each instrument. Then when a user requests the bad pixel monitor page, it simply loads the html file.

Bokeh creates and saves the html file itself. There is then a bit of editing to place the relevant lines from that file into our jinja template. The bokeh lines can end up being quite long since they contain all of the plotted points and hover tools. So we sacrifice some speed there compared to the dark monitor plots, which are simply png files that are loaded and displayed. But in return for that we get more interesting/useful plots.


bhilbert4:
In order for this to work on the servers, we'll need to install the packages described in #1200. This is done on the test server and I didn't see any problems.

bhilbert4:

jwql/utils/constants.py
```diff
@@ -143,6 +143,12 @@
 DARKS_BAD_PIXEL_TYPES = ['HOT', 'RC', 'OTHER_BAD_PIXEL', 'TELEGRAPH']
 FLATS_BAD_PIXEL_TYPES = ['DEAD', 'OPEN', 'ADJ_OPEN', 'LOW_QE']
 
+# The maximum number of bad pixels allowed on a bad pixel monitor plot. If there
+# are more than this number of bad pixels identified for a particular type of
+# bad pixel, then the figure is saved as a png rather than an interactive plot,
+# in order to reduce the amount of data sent to the browser.
+BAD_PIXEL_MONITOR_MAX_POINTS_TO_PLOT = 15000
```
mfixstsci, line 150: This is a good idea for performance. Do we expect there to be this many bad pixels 😨

jwql/utils/utils.py
```diff
@@ -117,6 +125,88 @@ def _validate_config(config_file_dict):
...
+        output_filename = os.path.join(outdir, os.path.basename(filename).replace('fits','png'))
+        export_png(plot, filename=output_filename)
+        permissions.set_permissions(output_filename)
+        return output_filename
+    else:
+        return None
+
+
+def screenshot_as_png():
+    from bokeh.io.export import get_screenshot_as_png
```
mfixstsci: Could you move the import to the top of the file please :)

jwql/website/apps/jwql/monitor_pages/monitor_dark_bokeh.py
```diff
@@ -235,18 +235,7 @@ def create_plot(self):
...
-
-                # Create an array representation for the image `img`, and an 8-bit "4
-                # layer/RGBA" version of it `view`.
-                img = np.empty((ydim, xdim), dtype=np.uint32)
-                view = img.view(dtype=np.uint8).reshape((ydim, xdim, 4))
-
-                # Copy the RGBA image into view, flipping it so it comes right-side up
-                # with a lower-left origin
-                view[:,:,:] = np.flipud(np.asarray(rgba_img))
+                view = read_png(self.dark_image_picture)
```
mfixstsci, line 238: nice use of making this a function here.

jwql/website/apps/jwql/monitor_views.py
```diff
@@ -62,20 +62,13 @@ def bad_pixel_monitor(request, inst):
...
     """
-
-    # Ensure the instrument is correctly capitalized
-    inst = JWST_INSTRUMENT_NAMES_MIXEDCASE[inst.lower()]
-
-    tabs_components = bokeh_containers.bad_pixel_monitor_tabs(inst)
-
-    template = "bad_pixel_monitor.html"
+    # Locate the html file for the instrument
+    template = f"{inst.lower()}_bad_pix_plots.html"
```
mfixstsci, line 66: Much cleaner here without all of the templating.

mfixstsci (approved):
Thanks for submitting this @bhilbert4 just a couple of comments here.

jwql/utils/constants.py
bhilbert4: Not expected really, but we do see it. I think it means the algorithms for identifying bad pixels probably need to be tweaked. But this at least functions as a safety net for cases where it happens.

jwql/utils/utils.py
bhilbert4: It turns out that this entire function I wrote, create_png_from_fits(), is not used anywhere. Should I get rid of it completely? Or is it potentially useful enough that it's worth keeping?

mfixstsci: It's so funny you said that because I thought this was `read_png`. I would say we should remove it if we don't use. Is there a chance you might want use it in the future?

bhilbert4: I removed it. If we ever need to generate png files from screenshots, I'm sure we can re-create it.

mfixstsci:
@bhilbert4 thanks for the responses, I am going to wait for the branch to rebase and then will merge the changes.

PR #1203, opened 2023-02-28
Create archive page asynchronously

### LINKED ISSUE #959: Speed up getting the number of files for a proposal
In order to get the number of files for a given proposal as listed on the instrument archive page, JWQL queries MAST for each program. These queries at the moment seem to be taking about 0.1 seconds per program. So as the number of programs goes up, so will the page loading time. 

But, eventually each program completes, at which point the number of files will no longer change. I wonder if there's a way to check the status of a program and once it is complete, we can store the number of files somewhere, so that we no longer need to query MAST for it?

### LINKED ISSUE #1157: Archive page for NIRCam is very slow to load
The NIRCam page is a lot slower to load than the pages for the other instruments, probably because there are more NIRCam proposals, and each proposal has more images, than those of the other instruments. It would be great if we could find a way to speed up the loading of this page.

One thought: Could this be related to the change we made a while back where we replaced a glob call over each proposal's list of preview images/thumbnails and replaced that with a text file that is read in? That sped things up at the time, but maybe now that text file is getting so big that reading it is taking a long time?

bhilbert4:
Rather than querying the database and constructing the context for an instrument's archive page at the time of the request, do that work asynchronously, and save the results in a json file. Then when there is a request for the page, read in the json file and create the page.

### DIFF
--- a/jwql/website/apps/jwql/views.py
+++ b/jwql/website/apps/jwql/views.py
@@ -333,7 +333,7 @@ def archived_proposals_ajax(request, inst):
         Outgoing response sent to the webpage
     """
     # Read in the json file created by data_containers.create_archived_proposals_context
-    # as use as the context
+    # and use as the context
     output_dir = get_config()['outputs']
     context_file = os.path.join(output_dir, 'archive_page', f'{inst}_archive_context.json')
 
--- a/jwql/website/apps/jwql/views.py
+++ b/jwql/website/apps/jwql/views.py
@@ -44,6 +44,7 @@ Dependencies
 from collections import defaultdict
 from copy import deepcopy
 import csv
+import json
 import glob
 import logging
 import os
@@ -181,12 +182,12 @@ def api_landing(request):
 def save_page_navigation_data_ajax(request):
     """
     Takes a bracketless string of rootnames and expstarts, and saves it as a session dictionary
-    
+
     Parameters
     ----------
     request: HttpRequest object
         Incoming request from the webpage
-    
+
 
     Returns
     -------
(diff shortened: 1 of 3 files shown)


bhilbert4:
Github won't let me link it, but this should also resolve #959

bhilbert4:
For this to work on the servers, we need to create the os.path.join(get_config('outputs'), "archive_page")  directory.

bhilbert4:
On the test server: 
Time needed to load the nircam archive page using the develop branch: 1 minute and 45 seconds
Time needed when using this branch: 13 seconds.
🥳

bhilbert4:
@mfixstsci I think this is ready for review.

bhilbert4:
With these changes, the json files containing the content of the archive pages are updated after each run of archive_database_update.py, which is done via cron job.

bhilbert4:
In addition to creating the new archive_page subdirectory under the outputs directory, we'll also need to run the new script on all of the instruments when we merge this.

```
from datetime import datetime
from jwql.website.apps.jwql.data_containers import create_archived_proposals_context
insts = ['nircam', 'miri', 'nirspec', 'niriss', 'fgs']
for inst in insts:
   print(f'Working on {inst}, {datetime.now()}')
   create_archived_proposals_context(inst)
```

jwql/website/apps/jwql/views.py
```diff
@@ -331,81 +332,13 @@ def archived_proposals_ajax(request, inst):
...
-    thumbnails_dict['proposals'] = proposal_nums
-    thumbnails_dict['thumbnail_paths'] = thumbnail_paths
-    thumbnails_dict['num_files'] = total_files
-    thumbnails_dict['viewed'] = proposal_viewed
-    thumbnails_dict['exp_types'] = thumb_exp_types
-    thumbnails_dict['obs_time'] = thumb_obs_time
+    # Read in the json file created by data_containers.create_archived_proposals_context
+    # and use as the context
+    output_dir = get_config()['outputs']
+    context_file = os.path.join(output_dir, 'archive_page', f'{inst}_archive_context.json')
```
mfixstsci, line 338: This is slick 👍

mfixstsci (approved):
LGTM @bhilbert4 I am going to rebase the branch and then merge it once the CI is done.

PR #1206, opened 2023-03-01
Catch missing obs in navigation data

melanieclarke:
Minor fix to the thumbnails_per_obs template to catch mismatch between the obs_list generated from navigation data and the currently displayed observation.


BradleySappington (approved):
Looks good!

melanieclarke:
@mfixstsci - can we get this one merged in to patch the bug on production?

PR #1211, opened 2023-03-09
Make pyvo searches robust against updated instrument names

### LINKED ISSUE #1208: Recent data not showing up via instrument archive page
On the NIRSpec archive page (https://jwql.stsci.edu/nirspec/archive/#) sorting the proposals by 'Recent' results in 1491, 2418, and 2302, 1272, and 1981 as the first five entries.
Clicking down into these programs though, the exposure dates for the data are:
1491: Aug 1, 2022
2418: Feb 14, 2023
2302: Feb 3, 2023
1272: Dec 2, 2022
1981: Dec 25, 2022

But when I search using the date range page, I see that these programs do in fact have new data taken in the last day or two.  (Note that there seem to be a lot of missing thumbnails for these new files, and the preview images are almost completely full of what look like CRs https://jwql.stsci.edu/NIRSpec/jw01491027001_01202_00001_nrs2/. #1207 is open for the bad preview images)

Back on the archive page when sorted by most recent, the thumbnail link does not go to the observation level page for the new data. Neither is the new observation number listed in the drop down menu from the program/observation-level page. 

Checking the archive_database_update logs:
grep "1491 - obs:027" archive_database_update_pljwql1.stsci.edu_2023-03-07*log
grep "1491 - obs:027" archive_database_update_pljwql1.stsci.edu_2023-03-06*log
It looks like 4 root_file_info entries were created in archive_database_update_pljwql1.stsci.edu_2023-03-06-03-41.log

Not quite sure what is going on here. I'll continue looking into it.

bhilbert4:
Resolves #1208 

MAST has updated the values that are present in the "instrument" column of the CAOM database such that they are no longer e.g. "MIRI", but instead e.g. "MIRI/IMAGE" or "MIRI/LRS". This doesn't affect the Mast.Jwst.Filtered services that JWQL uses for most queries, but it does affect all of the pyvo queries that we make. This PR updates those pyvo queries to add a wildcard after the instrument name, so that we will get back results for all modes of a given instrument. Without this, our queries currently do not return recent data, since those have been switched to the new instrument/mode value, and as data are slowly reprocessed in MAST with the new values, the size of our query results would get smaller and smaller. The pyvo queries are used primarily to get a list of program numbers per instrument, as well as a list of rootnames for a given instrument and program. The missing results here were causing new data not to be displayed on observation level pages.

### DIFF
--- a/jwql/utils/utils.py
+++ b/jwql/utils/utils.py
@@ -561,8 +561,8 @@ def get_rootnames_for_instrument_proposal(instrument, proposal):
     rootnames : list
         List of rootnames for the given instrument and proposal number
     """
-    tap_service = vo.dal.TAPService("http://vao.stsci.edu/caomtap/tapservice.aspx")
-    tap_results = tap_service.search(f"select observationID from dbo.CaomObservation where collection='JWST' and maxLevel=2 and insName like '{instrument.lower()}' and prpID='{int(proposal)}'")
+    tap_service = vo.dal.TAPService("https://vao.stsci.edu/caomtap/tapservice.aspx")
+    tap_results = tap_service.search(f"select observationID from dbo.CaomObservation where collection='JWST' and maxLevel=2 and insName like '{instrument.lower()}%' and prpID='{int(proposal)}'")
     prop_table = tap_results.to_table()
     rootnames = prop_table['observationID'].data
     return rootnames.compressed()
--- a/jwql/website/apps/jwql/data_containers.py
+++ b/jwql/website/apps/jwql/data_containers.py
@@ -958,8 +958,8 @@ def get_instrument_proposals(instrument):
     inst_proposals : list
         List of proposals for the given instrument
     """
-    tap_service = vo.dal.TAPService("http://vao.stsci.edu/caomtap/tapservice.aspx")
-    tap_results = tap_service.search(f"select distinct proposal_id from dbo.ObsPointing where obs_collection='JWST' and calib_level>0 and instrument_name like '{instrument.lower()}'")
+    tap_service = vo.dal.TAPService("https://vao.stsci.edu/caomtap/tapservice.aspx")
+    tap_results = tap_service.search(f"select distinct proposal_id from dbo.ObsPointing where obs_collection='JWST' and calib_level>0 and instrument_name like '{instrument.lower()}%'")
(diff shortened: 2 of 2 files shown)


bhilbert4:
Tom Donaldson also recommended switching the pyvo TAP url from http to https, when the former were being blocked by the STScI firewall during testing.

bhilbert4:
Testing on the dev server: results look good so far. Miri program 2311 obs 001 page shows in the dropdown menu, options for observations 1 through 7 (as expected). On the ops server, without this fix, there are only options for obs 1,2,3,5, since the others were observed after Feb 28 and use the updated instrument names in MAST.

bhilbert4:

jwql/utils/utils.py
```diff
@@ -561,8 +561,8 @@ def get_rootnames_for_instrument_proposal(instrument, proposal):
     rootnames : list
         List of rootnames for the given instrument and proposal number
     """
-    tap_service = vo.dal.TAPService("http://vao.stsci.edu/caomtap/tapservice.aspx")
-    tap_results = tap_service.search(f"select observationID from dbo.CaomObservation where collection='JWST' and maxLevel=2 and insName like '{instrument.lower()}' and prpID='{int(proposal)}'")
+    tap_service = vo.dal.TAPService("https://vao.stsci.edu/caomtap/tapservice.aspx")
```
mfixstsci, line 564: Security is a good thing.

mfixstsci (approved):
Hey @bhilbert4 these changes look fine to me. You mentioned wild card assignment for the instrument names to handle the name changes, where are these wild cards placed? I see updates to the pyvo calls but they still accept `instrument.lower()` as an argument.

bhilbert4:
@mfixstsci the '%' after the instrument name is the wildcard.

mfixstsci:
> @mfixstsci the '%' after the instrument name is the wildcard.

Ha! You're right. I am thinking in python formatted string brain. I will go ahead and merge them.

jwql/utils/utils.py
bhilbert4: When I was running the code manually, the ST firewall was not allowing the use of http. It kept crashing the script. I'm not sure why it was crashing for me yesterday, but hasn't been crashing on the JWQL servers. Either way, https seems to be the answer.

PR #1197, opened 2023-02-20
List viewed/new status by instrument

melanieclarke:
Adds a button to the instrument archive pages to download a CSV report containing the new/viewed status by root filename.  Additional useful observation descriptors, configured by instrument, are also added to the report.  These may be any field defined in the Django models (RootFileInfo, Observation, or Proposal).

If accessed via the archive page button, the report will respect the current filter and sort settings on the page. The report can also be downloaded directly at {instrument}/report, with optional filtering and sorting passed as query parameters, e.g. {instrument}/report?sort_as=descending&look=viewed.

For a JSON report of the same data, there is an API endpoint at api/{instrument}/looks.

PR #1213, opened 2023-03-13
Display all detectors

### LINKED ISSUE #1106: Optionally show multiple detectors on an exposure page
The NIRSpec team requested the ability to view the two detectors side by side for an exposure. This will help when examining traces that span multiple detectors.

Similarly for NIRCam, this could be helpful for viewing extended sources that span multiple detectors.

There are two options for this that come to mind at the moment:
1. Implement a button that the user can toggle. When pressed, the preview images for neighboring detectors are shown next to the preview image for the current detector.
2. We re-implement support for level 3 pipeline outputs (at least for imaging mode). In this case, we make preview images for i2d files, and add them to the list of suffixes. This method has the advantage that the JWQL team doesn't need to think about relative detector locations, and doesn't require significant changes to the exposure-level webpages.

For modes other than imaging, we may not want to support i2d files yet, since in many cases those files contain (possibly lots of) 1-d tables of spectra.

### LINKED ISSUE #1138: Apply anomalies to all detectors of an exposure
Adding a button near the list of anomalies that would, when pressed, apply the selected anomalies to all detectors used in that exposure could potentially save a good amount of time for NIRCam, and maybe NIRSpec. Certain anomalies, like scattered light, persistence, etc will often show up on multiple detectors for a given exposure. Only having to open the page for one detector and populating the database for all of them at once could be handy.

Or is this another argument to implement #1106? In that case, we could then have a single set of anomaly buttons and anything selected would apply to all detectors, or we'd need a separate set of anomaly buttons to go with each detector.

### LINKED ISSUE #1209: Revisit tests that are skipped on github actions
All of the data_container unit tests are skipped in github actions due to “central store access” but it appears many of them don't need to be skipped.  We should revisit our tests that are being skipped to verify that they shouldn’t run with the CI code. It may be that many of the tests were set to be skipped before launch because we assumed they would need central store access, when in fact they don't.

melanieclarke:
On the observation pages, there is now an option to group by exposure or file. For instruments with multiple detectors, grouping by exposure pulls up a page where previews from all simultaneous detector data are displayed together. From this page, you can mark an anomaly to apply to the whole group of files, or mark the whole group as Viewed. The preview images in the exposure group page are clickable: they will take you to the individual image page for that file, which works the way it always has.

For instruments with no simultaneous detectors, the exposure grouping is allowed, but has no real effect. The exposure pages have equivalent functionality to the existing image pages.

The archive date image page and JWQL query page also allow grouping by exposure.


melanieclarke:
Note: PR https://github.com/spacetelescope/jwql/pull/1197 should be merged before this one.

jwql/website/apps/jwql/data_containers.py
```diff
@@ -316,22 +447,20 @@ def get_current_flagged_anomalies(rootname, instrument):
```
bhilbert4, line 474: Could you add docstrings for instrument and n_match? Is n_match always 1 since a True value from the database will translate into a 1?

bhilbert4: Ah I see, from the call in get_anomaly_forms(), n_match will be the number of rootnames. I thought the call to first() in line 457 would return only the first matching row?

melanieclarke: Good catch -- I'll update the docstring.

N_match is for distinguishing which anomalies belong to a whole exposure group and which belong only to a single file in a group.  If you're only checking for a single file, n_match should be 1.  If you're checking for an exposure group, n_match should be the total number of files in the group.  This is needed as an argument because not every rootfile has an entry in the anomaly table: they only appear there if someone has marked an anomaly at some point.

The first() in line 457 takes the first entry for each rootfile because it follows the groupby function on the dataframe.  This keeps the last marked anomaly state for each rootfile, in case it has multiple entries in the database.

For example, if there are 2 files in an exposure group and one is marked with 'persistence' and 'crosstalk' and the other is marked with 'crosstalk' only, then n_match=2 and the current anomalies for the group will include only 'crosstalk'.  If there are 3 files in the group, and the third file has no entry in the database, then n_match=3 and no current anomalies are returned for the group.

jwql/website/apps/jwql/data_containers.py
```diff
@@ -781,6 +927,79 @@ def mast_query_filenames_by_instrument(instrument, proposal_id, observation_id=N
     return result
 
 
+def get_filesystem_filenames(proposal=None, rootname=None,
```
bhilbert4: Clever. I like this.

jwql/website/apps/jwql/data_containers.py
```diff
@@ -958,25 +1163,42 @@ def get_image_info(file_root, rewrite):
...
-        # be predicted for a given program.
-        if ((suffix not in SUFFIXES_WITH_AVERAGED_INTS) and (suffix[-3:] != 'crf')):
-            image_info['total_ints'][suffix] = fits.getheader(filename)['NINTS']
+        # Record how many integrations exist per filetype.
+        if suffix not in SUFFIXES_WITH_AVERAGED_INTS:
+            header = fits.getheader(filename)
+            nint = header['NINTS']
+            if 'time_series' in parsed_fn['filename_type']:
+                # time series segments need special handling
+                intstart = header.get('INTSTART', 1)
```
bhilbert4, line 1205: Good catch

jwql/website/apps/jwql/data_containers.py
bhilbert4: So if this function is called with a rootname of jw86600008001_02101_00001_guider2, the query will only return entries for guider2 though, right? Entries for the same exposure but using guider1 won't be returned?  I think that's part of what was confusing me before.

jwql/website/apps/jwql/templates/archive.html
```diff
@@ -35,14 +35,19 @@ <h3 id='title'>Archived {{ inst }} Images</h3>
             <!-- Sort Options -->
             <div class="mx-4 mr-5 ml-auto">
                 Sort by:<br>
-                <span class="dropdown">
-                  <button class="btn btn-primary dropdown-toggle" type="button" id="sort_dropdownMenuButton" data-toggle="dropdown" aria-haspopup="true" aria-expanded="false">Ascending</button>
+                <div class="dropdown">
+                  <button class="btn btn-primary dropdown-toggle" type="button" id="sort_dropdownMenuButton" data-toggle="dropdown" aria-haspopup="true" aria-expanded="false">Descending</button>
```
bhilbert4, line 39: Thanks for catching this. I've been meaning to get the default button value updated for quite a while.

jwql/website/apps/jwql/views.py
```diff
@@ -978,11 +1038,30 @@ def explore_image_ajax(request, inst, file_root, filetype, scaling="log", low_li
...
+def save_image_group_ajax(request):
+    """Save the latest selected group type in the session."""
+    image_group = request.GET['group_type']
+    request.session['image_group'] = image_group
+    context = {'item': request.session['image_group']}
+    return JsonResponse(context, json_dumps_params={'indent': 2})
+
+
+def save_image_sort_ajax(request):
+
```
bhilbert4, line 1062: Can you fix this docstring?

jwql/website/apps/jwql/views.py
```diff
@@ -978,11 +1038,30 @@ def explore_image_ajax(request, inst, file_root, filetype, scaling="log", low_li
     return JsonResponse(context, json_dumps_params={'indent': 2})
 
 
+def save_image_group_ajax(request):
+    """Save the latest selected group type in the session."""
```
bhilbert4: This docstring needs params and returns as well.

jwql/tests/test_data_containers.py
```diff
@@ -49,18 +134,197 @@ def test_get_all_proposals():
     assert len(proposals) > 0
 
 
-@pytest.mark.skipif(ON_GITHUB_ACTIONS, reason='Requires access to central storage.')
-def test_get_expstart():
+@pytest.mark.parametrize('untracked,input_suffixes,expected',
```
bhilbert4, line 137: Good catch on these tests that we don't need to skip on github actions. Your updated tests match the list I created the other day. I'll attach the relevant issue to this PR so that it will be closed.

bhilbert4 (commented):
This looks great, and was a ton of work! I know the NIRCam team will be excited to be able to see all the detectors on a single page. 
@mfixstsci I gave everything a look except for the js updates.

jwql/website/apps/jwql/data_containers.py
melanieclarke: Yes, that's right.  The rootname is used as a starts-with filter, so if jw86600008001_02101_00001_guider2 is passed, only that file will be matched.  If jw86600008001_02101_00001 is passed, then either jw86600008001_02101_00001_guider1 or jw86600008001_02101_00001_guider2 will be matched.  I'll clarify the example root name in the docstring.  

The group root name is passed only in the context of the exposure pages: the image pages still pass the full file root name.

jwql/website/apps/jwql/views.py
melanieclarke: Done!

jwql/website/apps/jwql/views.py
melanieclarke: Done!

jwql/tests/test_data_containers.py
melanieclarke: Oh good, thanks.  I didn't realize there was an issue for this.

jwql/website/apps/jwql/static/css/jwql.css
```diff
@@ -537,7 +542,21 @@
...
+   .thumbnail-group {
+      display: inline;
+      font-size: 0.75rem;
+  }
+  .thumbnail-group-active {
+      display: block;
+      width: 90%;
+      height: 90%;
+      border: 1px solid #2d353c;
+      box-shadow: 5px 5px #c85108, 10px 10px #bec4d4;
```
BradleySappington, line 556: Nice touch

jwql/website/apps/jwql/static/js/jwql.js
```diff
@@ -6,8 +6,93 @@
...
+
+    // Change the radio button to check the right filetype
+    document.getElementById(type).checked = true;
+
+    // Store the currently displayed suffix
+    document.getElementById("view_file_type").setAttribute('data-current-suffix', type);
+
+    // Clean the input parameters
+    num_ints = num_ints.replace(/&#39;/g, '"');
+    num_ints = num_ints.replace(/'/g, '"');
```
BradleySappington: These two cleaning lines are used a lot of places.  Consider making a function in case we ever need to expand on it or use it elsewhere? 

```suggestion
    num_ints = clean_int_param(num_ints);

    function clean_int_param(int_to_clean){
        int_to_clean = int_to_clean.replace(/&#39;/g, '"');
        int_to_clean = int_to_clean.replace(/'/g, '"');
        return int_to_clean;
    } 
```

jwql/website/apps/jwql/static/js/jwql.js
```diff
@@ -6,8 +6,93 @@
...
+    available_ints = available_ints.replace(/'/g, '"');
+    available_ints = JSON.parse(available_ints)[type];
+
+    // Get the total number of integrations
+    total_ints = total_ints.replace(/&#39;/g, '"');
+    total_ints = total_ints.replace(/'/g, '"');
+    total_ints = JSON.parse(total_ints);
+
+    // Update the APT parameters
+    document.getElementById("proposal").innerHTML = group_root.slice(2,7);
```
BradleySappington: duplicated slicing code with magic numbers.
Its only used one other place so I'm okay with keeping it like this, but maybe something we should keep an eye on with future javascript changes.

jwql/website/apps/jwql/static/js/jwql.js
```diff
@@ -6,8 +6,93 @@
...
+                           '/' + group_root + '_' + detector + '_' + type + '_integ0.jpg';
+        img.src = jpg_filepath;
+        img.alt = jpg_filepath;
+
+        // Show/hide the viewer as appropriate
+        show_viewer(detector, jpg_filepath);
+    }
+
+    // Reset the slider values
+    document.getElementById("slider_range").value = 1;
```
BradleySappington: This should probably be a standard reset slider function as well (used a lot of places with same naming convention)

jwql/website/apps/jwql/static/js/jwql.js
```diff
@@ -73,16 +161,14 @@
         document.getElementById("int_after").disabled = true;
     }
 
-    // Update the image download and header links
-    // document.getElementById("download_fits").href = '/static/filesystem/' + file_root.slice(0,7) + '/' + fits_filename + '.fits';
-    // document.getElementById("download_jpg").href = jpg_filepath;
+    // Update the image exploration and header links
```
BradleySappington, line 139: Thank you for removing dead code!

jwql/website/apps/jwql/static/js/jwql.js
```diff
@@ -97,37 +183,41 @@
  * @param {String} direction - The direction to switch to, either "left" (decrease) or "right" (increase).
  *                             Only relevant if method is "button".
  */
-function change_int(file_root, num_ints, available_ints, method, direction = 'right') {
+function change_int(file_root, num_ints, available_ints, method, direction='right') {
```
BradleySappington: Stylistic input, calling it "change_integration" will be more clear for future developers.  "Int" is a particularly tricky abbreviation that forces a docstring glance.

jwql/website/apps/jwql/static/js/jwql.js
```diff
@@ -931,54 +1196,71 @@ function update_sort_options(data, base_url) {
...
+                      '" data-exp_start="' + file.expstart + '" data-look="' + viewed + '" data-exp_type="' + exp_type + '">';
+        content += '<div class="thumbnail-group">'
+        content += '<a class="thumbnail-link" href="#" data-image-href="/' +
+                   instrument + '/' + rootname + '/" data-group-href="/' +
+                   instrument + '/exposure/' + filename_dict.group_root +  '">';
+        content += '<span class="helper"></span>'
+
+        // Make sure thumbnail img always has a src and alt
+        content += '<img id="thumbnail' + i +
+                   '" src="/static/img/default_thumb.png" ' +
```
BradleySappington, line 1286: I think we use the imagenotfound instead of the plain blue.

```suggestion
                   "/static/img/imagenotfound.png";
```

BradleySappington (commented):
Most of these changes are stylistic, this is a ton of great work!

One overall issue is that the default thumbnail (which I believe should be changed to imagenotfound.png) is not being used for your groups or files if the thumbnail does not exist. 

Example here if loaded on test:
https://jwql-test.stsci.edu/NIRCam/archive/2057/obs003/

jwql/website/apps/jwql/static/js/jwql.js
melanieclarke: For this one -- I added the plain blue image for thumbnails that weren't found.  Prior to this change, the img tag had no src or alt added if there was an error, so it appeared as a plain white square indistinguishable from background.  I was hoping the blue square would be both more visible than no image at all, and more legible on mouseover than adding the imagenotfound.png.

The example on test had an issue with permissions on the new image, for some reason, so it wasn't showing up.  It is there now (https://jwql-test.stsci.edu/NIRCam/archive/2057/obs003/).  Can you please let me know if you still think I should swap it back to the other not found image?

jwql/website/apps/jwql/static/js/jwql.js
melanieclarke: Good point!  I'll make that change.

jwql/website/apps/jwql/static/js/jwql.js
BradleySappington: Mouseover is certainly more accessible with the blue than with "imagenotfound" (so I think I support this change).  
I'd defer to @bhilbert4 as the PO.

Once I click on a group or file, there is no longer a placeholder for a missing preview_image (I think this situation should use imagenotfound).

melanieclarke: There is still a default not found image on the image pages, e.g.
https://jwql-test.stsci.edu/NIRCam/jw02057003001_02104_00004_nrcb1/

The exposure page handles it differently -- it just hides viewers for images it can't find, whether because there is no such FITS file or because the preview wasn't found.  I can look into adding a fallback image instead if the file exists but there is no preview image.

jwql/website/apps/jwql/static/js/jwql.js
melanieclarke: Sure, I can add that.

jwql/website/apps/jwql/static/js/jwql.js
bhilbert4: Yeah, the text in the mouseover is certainly more legible with the blue compared to white. 

I think keeping the Image Not Found fallback preview image makes sense. Some people do tend to jump onto JWQL shortly after their data are taken, but before preview images have been made. It seems like they'd tend to freak out less if they can see a message that there's no preview image yet.

jwql/website/apps/jwql/static/js/jwql.js
melanieclarke: Done!

jwql/website/apps/jwql/static/js/jwql.js
melanieclarke: I added a function to at least keep the magic numbers in one place.  Some day, we may want to make it parse filenames in a more sophisticated way, like the Python utility does.

jwql/website/apps/jwql/static/js/jwql.js
melanieclarke: Done!

jwql/website/apps/jwql/static/js/jwql.js
melanieclarke: Done!

jwql/website/apps/jwql/static/js/jwql.js
melanieclarke: I pushed a modification to address this.  There is now a fallback div/image for each detector in the exposure that is shown on load.  If any real preview images get shown for the detector, the fallback is hidden and doesn't come back.

I think this should allow show/hide for FITS files that don't exist (like s3d for mirimage, i2d for mirifushort/long), but should show default images and links in the case where there are no previews available yet.

The change is on test. Let me know what you think.

bhilbert4: That looks good! And the filenames are available in the drop down as well as when hovering over the fallback preview image, so users have a way of seeing which detectors were used. And I don't see any option for mirimage s3d, like you said.

BradleySappington: This looks good in chrome, but a little wonky in safari.  Not sure if safari requires more deliberate sizing of the imgnotfound preview images.
https://jwql-test.stsci.edu/NIRCam/exposure/jw02057003001_02102_00001/

That said, it looks great in chrome, so if its not an incredibly easy fix I wouldn't worry about it for now, especially since the user can still tell they're temporary placeholders.  I'd rather get this all merged in.

bhilbert4: The layout in safari looks good, with the LW image centered below both columns of the SW. Everything's just a little stretched. 😆  In Firefox everything looks similar to how it does in Chrome.

BradleySappington: I'm a-ok with not addressing the safari stretch!

melanieclarke: Sorry, I noticed that at the same time you did -- the Safari stretch is fixed.

melanieclarke: I just pushed one more fix, to keep the view header/explore image links available in the case where there are no preview images yet.  I think that should be it.

BradleySappington: Just checked it out on test. looks great!

BradleySappington (approved):

mfixstsci:
Thank you for submitting this @melanieclarke and thank you @BradleySappington for the review. Merging!

PR #1217, opened 2023-03-17
[WIP] Updates and Fixes to EDB Monitor

mfixstsci:
When the EDB Monitor attempts to plot every data point returned from the mnemonic, we are currently running into a column mismatch error from astropy. This is because the median time values do not match the length of the median eu_values obtained from the mnemonic. Here is a quick fix that assigns the proper data to the `multiday_table`.


mfixstsci:
Updating EDB Monitor:
- Change column from mean to median
- Figure out multiple numpy errors and crashes

PR #1218, opened 2023-03-17
Set file permissions based on the account name

### LINKED ISSUE #1220: Web app not displaying prev images, with permission error
There are some preview images that the web app is failing to display, while reporting a permission error. 

Investigation shows that these files have rw-r----- permission. By not being world readable, the web app cannot access them. It seems that perhaps the umask policy on the server has changed. Previously files were world readable. And in fact, set_permissions() was not doing anything, since it was looking for jwqladm as the owner files to make world readable. But none of our files have that owner name.

We need to fix set_permissions() such that it does work on files created on the servers (as well as locally). We need to make sure that it makes files world readable.

bhilbert4:
Recent files created by e.g. the preview image generator have owner read and write permissions, and group read permissions. Previously these files were also world readable, but not any more. We need them to be world readable in order for the web app to read them in and display them. This PR updates our set_permissions() function such that:

For files created by one of the server accounts, files will be set to be rwxr-xr--. Files created by a local account (i.e. running a monitor or web app locally), will be set to rwxrwxr--. The latter change is due to the fact that there are multiple people who can run the web app locally but output files to the same locations (central store), and they therefore need to be able to overwrite an existing file that someone else may have created.

Also note that up until now, set_permissions() was not doing anything to any file, since the default owner listed in that file, 'jwqladm', is not the name of any of the service accounts. So things were working only because the umask for the servers happened to be set to what we needed.


bhilbert4:
This still needs to be tested on the test or dev servers but in the meantime, @melanieclarke @BradleySappington @mfixstsci do you agree with the logic behind these changes, and the implementation here? I want to be careful not to create more file permission problems while trying to solve this one.

melanieclarke:
I think this looks right, except that files should not have 'x' permission by default and directories should. Do the mode codes as written make that distinction?

bhilbert4:
Good catch. I've updated the default mode to assume a file, and then if it is found to be a directory, the codes for owner and group executable are added.

melanieclarke:
I think it needs to be world executable too, or the browser won't be able to see into program directories.

bhilbert4:
Done. Good catch.

mfixstsci:
@bhilbert4 looks like we have some issues with the tests for this module. Do you have an idea of what could be causing the failures?

bhilbert4:
I'm working on it. It's due to the dir vs file differences in the mode.

jwql/tests/test_permissions.py
```diff
@@ -51,7 +51,7 @@ def test_directory(test_dir=TEST_DIRECTORY):
 
     yield test_dir
     if os.path.isdir(test_dir):
-        os.remove(test_dir)
```
bhilbert4, line 54: Tests are now passing locally. When testing this on one of the servers, it would be helpful to run the tests there as well, in addition to making sure that file/directory permissions end up being what we expect.

bhilbert4: I ran pytest on the test server, and all tests pass. I've confirmed that running set_permissions() with no keyword values on a file created on the server results in world readable files.

bhilbert4:

mfixstsci (approved):
Thanks @bhilbert4 !

PR #1229, opened 2023-03-28
Use Model Database for Anomaly form

BradleySappington:
No longer use the postgres db, now populate the form from the Anomalies Model, and update the same location accordingly.

### DIFF
--- a/jwql/tests/test_data_containers.py
+++ b/jwql/tests/test_data_containers.py
@@ -31,6 +31,9 @@ import numpy as np
 import pandas as pd
 import pytest
 
+from django.conf import settings
+settings.configure()
+
 # Skip testing this module if on Github Actions
 ON_GITHUB_ACTIONS = '/home/runner' in os.path.expanduser('~') or '/Users/runner' in os.path.expanduser('~')
 from jwql.website.apps.jwql import data_containers
--- a/jwql/tests/test_data_containers.py
+++ b/jwql/tests/test_data_containers.py
@@ -159,25 +159,27 @@ def test_get_available_suffixes(untracked, input_suffixes, expected):
 
 def test_get_current_flagged_anomalies(mocker):
     # get a sample query group with 2 files
-    rootname = 'jw02589006001_04101_00001-seg001'
-    instrument = 'NIRSpec'
+    # TODO update this test to use the new django models
 
-    # mock a single shared anomaly type
-    mocker.patch.object(data_containers.di, 'session', MockSessionGroupAnomaly())
+    # rootname = 'jw02589006001_04101_00001-seg001'
+    # instrument = 'NIRSpec'
 
-    result = data_containers.get_current_flagged_anomalies(
-        rootname, instrument, n_match=2)
-    assert result == ['persistence']
+    # # mock a single shared anomaly type
+    # mocker.patch.object(data_containers.di, 'session', MockSessionGroupAnomaly())
 
-    # get a sample query for 1 file
-    rootname = 'jw02589006001_04101_00001-seg001_nrs1'
+    # result = data_containers.get_current_flagged_anomalies(
+    #     rootname, instrument, n_match=2)
+    # assert result == ['persistence']
 
-    # mock two anomalies for this file
-    mocker.patch.object(data_containers.di, 'session', MockSessionFileAnomaly())
+    # # get a sample query for 1 file
+    # rootname = 'jw02589006001_04101_00001-seg001_nrs1'
 
-    result = data_containers.get_current_flagged_anomalies(
-        rootname, instrument, n_match=1)
-    assert result == ['persistence', 'crosstalk']
+    # # mock two anomalies for this file
+    # mocker.patch.object(data_containers.di, 'session', MockSessionFileAnomaly())
+
+    # result = data_containers.get_current_flagged_anomalies(
+    #     rootname, instrument, n_match=1)
+    # assert result == ['persistence', 'crosstalk']
 
 
 @pytest.mark.skipif(ON_GITHUB_ACTIONS, reason='Requires access to django models.')
--- a/jwql/tests/test_data_containers.py
+++ b/jwql/tests/test_data_containers.py
@@ -31,8 +31,7 @@ import numpy as np
 import pandas as pd
 import pytest
 
-from django.conf import settings
-settings.configure()
+os.environ.setdefault("DJANGO_SETTINGS_MODULE", "jwql.website.jwql_proj.settings")
 
 # Skip testing this module if on Github Actions
 ON_GITHUB_ACTIONS = '/home/runner' in os.path.expanduser('~') or '/Users/runner' in os.path.expanduser('~')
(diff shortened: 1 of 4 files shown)


jwql/website/apps/jwql/data_containers.py
```diff
@@ -85,22 +87,20 @@
...
     configs = get_config()
     auth_mast = configs['auth_mast']
     mast_flavour = '.'.join(auth_mast.split('.')[1:])
     from astropy import config
     conf = config.get_config('astroquery')
     conf['mast'] = {'server': 'https://{}'.format(mast_flavour)}
-
-__location__ = os.path.realpath(os.path.join(os.getcwd(), os.path.dirname(__file__)))
-if not ON_GITHUB_ACTIONS and not ON_READTHEDOCS:
     FILESYSTEM_DIR = configs['filesystem']
```
bhilbert4, line 97: I think the "if not ON_GITHUB_ACTIONS and not ON_READTHEDOCS:" is still required here, right?

BradleySappington:
@bhilbert4  - I just commented out `test_get_current_flagged_anomalies` as it is now model based.  We will need to implement django model testing into our test suite as a whole via a separate issue.

BradleySappington:
@mfixstsci @melanieclarke @bhilbert4 - Ready for Review

jwql/tests/test_data_containers.py
```diff
@@ -159,25 +159,27 @@ def test_get_available_suffixes(untracked, input_suffixes, expected):
 
 def test_get_current_flagged_anomalies(mocker):
     # get a sample query group with 2 files
-    rootname = 'jw02589006001_04101_00001-seg001'
-    instrument = 'NIRSpec'
+    # TODO update this test to use the new django models
```
bhilbert4: Are you waiting to update this in a future PR?

bhilbert4 (approved):
LGTM. Just a couple of comments.

bhilbert4:
Looks like test_data_containers.py has a failure. Maybe it just needs to have the Django env variable set and config to be run?

jwql/website/apps/jwql/data_containers.py
BradleySappington: Those following lines of code get bumped up into the preceding group which has the same check in place.  I just merged all elements instead of having two groups with the same `if not ON_GITHUB_ACTIONS and not ON_READTHEDOCS:`

jwql/tests/test_data_containers.py
BradleySappington: Yup, I tagged you elsewhere regarding it.  We'll need to implement Django Model testing which should be its own issue.

melanieclarke: This test was one of my new additions, to test the group anomaly behavior and fill out coverage in data_containers.  I can refactor it later.

BradleySappington:
> Looks like test_data_containers.py has a failure. Maybe it just needs to have the Django env variable set and config to be run?

@bhilbert4 - Had to move the importing of the AnomalyForm in datacontainers into the `if not GithubActions` block.

BradleySappington:
@mfixstsci - TODO AFTER MERGE TO DEV AND MAIN

this code finalizes the anomaly form work as it will both read and write to the django models.  We need to update the django database for any anomalies that have recently been added.  so on dev/test/prod servers after merging to development we will need to run:
`$ python anomaly_db_transfer.py`

When dev is merged to main, and if/when production is actually running in main we will need to run it again there too.

BradleySappington:
@mfixstsci - ready to merge

jwql/website/apps/jwql/data_containers.py
```diff
@@ -435,59 +434,43 @@ def get_available_suffixes(all_suffixes, return_untracked=True):
...
+            anomalies_set.append(rootfileinfo.anomalies.get_marked_anomalies())
+        except (ObjectDoesNotExist, AttributeError):
+            empty_anomaly_found = True
+
+    if not empty_anomaly_found:
+        # If all RootFileInfos have anomalies, calculate which anomalies exist in every RootFileInfo
+        flat_list = [anomaly for sublist in anomalies_set for anomaly in sublist]
+        for anomaly in all_anomalies:
+            if flat_list.count(anomaly) == len(rootfileinfo_set):
+                current_anomalies.append(anomaly)
```
melanieclarke: Question - how does this handle multiple entries in the anomaly table for the same root file?  For example, the case where an anomaly was marked but later removed.  I don't see where it is taking the latest one.

BradleySappington: Good question, All of the `Anomalies` models are a one to one relationship with a `RootFileInfo` model.  Each `RootFileInfo` model has a unique rootname, so we are no longer be using multiple entries for anomaly updates. We are just updating the `Anomalies` model associated with the single `RootFileInfo`.  The only time rootfileinfo_set will have more than one rootfileinfo in it, is for updating groups (and each will have a unique `RootFileInfo` associated with it).

melanieclarke: Got it, thanks!

jwql/website/apps/jwql/data_containers.py
```diff
@@ -435,59 +434,43 @@ def get_available_suffixes(all_suffixes, return_untracked=True):
...
-        for col, val in all_records.sum(axis=0, numeric_only=True).items():
-            # ignore rootname, flag_date, etc
-            if val == n_match and col in ANOMALIES_PER_INSTRUMENT:
-                current_anomalies.append(col)
+    empty_anomaly_found = False
+    for rootfileinfo in rootfileinfo_set:
+        try:
+            anomalies_set.append(rootfileinfo.anomalies.get_marked_anomalies())
+        except (ObjectDoesNotExist, AttributeError):
+            empty_anomaly_found = True
```
melanieclarke, line 467: This section is raising an error for me when trying to edit a group anomaly from an exposure page:

```
  File "data_containers.py", line 462, in get_current_flagged_anomalies
    for rootfileinfo in rootfileinfo_set:
TypeError: 'RootFileInfo' object is not iterable
```

jwql/website/apps/jwql/data_containers.py
BradleySappington: Although because you highlighted that code, I see I forgot my `break` after setting `empty_anomaly_found`!

jwql/website/apps/jwql/data_containers.py
melanieclarke: It looks like line 512 needs to be `[rootfileinfo]` instead of `rootfileinfo`

BradleySappington: weird, I tested that locally, will investigate, thanks!

BradleySappington: Clearly my local test was flawed, thanks for finding this!  Suggested fix works dev.

melanieclarke: Works for me locally too, now. Thanks!

melanieclarke (approved):

PR #1230, opened 2023-03-30
Remove bokeh templating from Bias Monitor

bhilbert4:
Replace bokeh templating for figure production with straight bokeh.


bhilbert4:
Looks good when run locally. Need to test on the dev or test server.

bhilbert4:
Hmmm. The results page for NIRCam is very slow to load (~20 seconds, on the test server). I'm going to look into switching this to be like the bad pixel monitor, where at the end of the monitor run, the plots will be created and saved into an html file.

bhilbert4:
Similar to the bad pixel plots, I'm successfully creating an html file with the bias plots, but when I try to open the file within the app, I'm getting an error:

Encountered unknown tag 'd'. Jinja was looking for the following tags: 'endblock'. The innermost block that needs to be closed is 'block'.",

For the bad pixel plots, the problem was a clash of bokeh datetime format specifiers being interpreted by jinja as tags in the html files. To get around this, we need to create a list of string representations of the datetime values, and include those in the ColumnDataSources for the plots. 

I'm just documenting this here, since it was difficult to track down the answer last time and we may see the error again for other monitors.

jwql/instrument_monitors/common_monitors/bias_monitor.py
```diff
@@ -60,6 +60,7 @@
 from jwql.utils.monitor_utils import update_monitor_table  # noqa: E402 (module import not at top)
 from jwql.utils.permissions import set_permissions  # noqa: E402 (module import not at top)
 from jwql.utils.utils import copy_files, ensure_dir_exists, filesystem_path, get_config  # noqa: E402 (module import not at top)
+from jwql.website.apps.jwql.monitor_pages.monitor_bias_bokeh import BiasMonitorPlots  # noqa: E402 (module import not at top)
```
bhilbert4, line 63: After merging, we'll have to create the initial copy of the instrument-specific bias pages manually. Then they will be updated each time the bias monitor runs. To create them manually, we just need to do:

from jwql.website.apps.jwql.monitor_pages.monitor_bias_bokeh import BiasMonitorPlots
for inst in ['nircam', 'niriss', 'nirspec']:
    BiasMonitorPlots('niriss')

bhilbert4:
@mfixstsci this is ready for review. GA seems to be acting a little strange, claiming that stdatamodels is not installed, even though it should be. I'll try re-running the tests tomorrow and see if the results are any different.

bhilbert4:
Ok, looks like tests are failing because we're using a version of stdatamodels from January (0.4.5). Once we update to a more recent version (1.3.0 is the latest), the tests should pass.

bhilbert4:
Actually, per Howard, we should probably update to: stdatamodels==1.3.1 and stcal==1.3.5

mfixstsci:
@bhilbert4 that sounds good. We need to perform the monthly updates this week anyways. We can merge this in once we get the updates in if that sounds good.

jwql/instrument_monitors/common_monitors/bias_monitor.py
mfixstsci: @bhilbert4 sounds good. I will make sure to do this once we get this PR in.

jwql/website/apps/jwql/bokeh_utils.py
```diff
@@ -0,0 +1,32 @@
+"""Various Bokeh-related utility functions for the ``jwql`` project.
```
mfixstsci, line 1: Very big fan of this. This will come in handy when removing bokeh templating/building new monitors 👍

mfixstsci (approved):

jwql/website/apps/jwql/bokeh_utils.py
bhilbert4: I'm kind of bummed, because bokeh-related functions are split between bokeh_containers and bokeh_utils. I had to create bokeh_utils here in order to avoid a circular import that was affecting the dark monitor pages.

bhilbert4:
With the updated version of stdatamodels in place, tests are now passing. Hurrah!

jwql/website/apps/jwql/bokeh_utils.py
mfixstsci: It would be nice to be able to consolidate all of it into one module. From slack, it sounds like you have an idea of how to solve this issue. Do we have an open issue for this right now?

bhilbert4: Not yet, but I can create one. I think it would be pretty straightforward to move the code that makes the figure tabs for each monitor from bokeh_containers.py. All that would be left then is a collection of utils-type functions. Maybe at that point we can rename it to bokeh_utils or something more accurate than bokeh_containers.

mfixstsci: Sounds good to me. I think the `_containers` lingo comes from django. I am cool with that, let's move that discussion there once the issue is made. I will go ahead and close this conversation so we can get this in! Thanks again!

mfixstsci:
> With the updated version of stdatamodels in place, tests are now passing. Hurrah!

WOOOOO! Awesome, thank you @melanieclarke for submitting the package version updates. Merging!

PR #1234, opened 2023-04-03
Updates for sqlalchemy>=2, django>=3.2

melanieclarke:
This pull request updates code for compatibility with SQLAlchemy > 2.0 and Django > 3.2.  Both of these versions introduced backwards incompatibilities that affected the JWQL application.


jwql/instrument_monitors/common_monitors/bias_monitor.py
```diff
@@ -430,7 +430,8 @@ def process(self, file_list):
                 bias_db_entry[key] = float(amp_medians[key])
 
             # Add this new entry to the bias database table
-            self.stats_table.__table__.insert().execute(bias_db_entry)
+            with engine.begin() as connection:
```
mfixstsci, line 433: I noticed that you are replacing insertion statements and adding a `with` beforehand. If I had to guess the connection is made and then closed once the `with` block is exited?

mfixstsci (approved):
Hey @melanieclarke -- These changes look great. Thank you for handling them, I am happy with the getter/setting naming conventions you implemented and the updated ways we connect and interface with the database.

jwql/instrument_monitors/common_monitors/bias_monitor.py
melanieclarke: That's right.  This is a change to the required syntax in SQLAlchemy 2.0: it no longer allows implicit connections like the previous .insert().execute() call.  

If called as `with engine.begin()`, the connection is closed at the end of the with block and any changes are committed.  If called as `with engine.connect()`, the connection is still closed at the end of the block, but changes are not automatically committed: in this case, you have to call `connection.commit()` explicitly, to commit changes as you go.

BradleySappington (approved):
Great work!

PR #1235, opened 2023-04-03
Bad pixel monitor plots: add missing import and remove unused function

### LINKED ISSUE #1231: Bad pixel monitor missing an import
The bad pixel monitor calls BadPixelPlots to create the html file that shows the monitor results, but the import statement for BadPixelPlots is missing.

It should be:
from jwql.website.apps.jwql.monitor_pages.monitor_bad_pixel_bokeh import BadPixelPlots

bhilbert4:
Small but important tweak for the bad pixel monitor. After #1202, the bad pixel monitor plots were changed such that they are generated each time the monitor runs, and the results are saved to an html file. But that update forgot to include an import statement in the monitor for the plots.

This PR adds that missing import statement, and also deletes the previous function that had been used to generate the plots. This function is no longer used.


bhilbert4:
@mfixstsci this should be ready for review.

mfixstsci:
@bhilbert4 I will go ahead and start my review. Looks like we have an issue with some of the CI test from your changes. I looked at the pipeline and stdatamodels versions and they looked fine to me. It also looks like we have a conflict that needs to be resolved as well.

mfixstsci (approved):
LGTM

bhilbert4:
Huh. Tests are failing because it's looking for the config file. Strange, since none of the tests have been updated here.

melanieclarke:
> Huh. Tests are failing because it's looking for the config file. Strange, since none of the tests have been updated here.

@bhilbert4 - I was just looking at this because I noticed the missing import and tried to add it in.  I think the import for BadPixelPlots needs to be not ON_GITHUB_ACTIONS because it needs config and filesystem info.

bhilbert4:
Ah, good catch @melanieclarke! I'll update it.

bhilbert4:
@mfixstsci looks like all the issues are resolved.

mfixstsci (approved):

Issue #1244, opened 2023-04-12
codecov has been deprecated


bhilbert4:
codecov has been pulled from pypi, and the repository is now archived and not supported.

We install codecov via conda in our environment files, so the removal from pypi hasn't caused a problem for us. But we should probably migrate to the new codecov uploader.

Here is a blog post that gives migration instructions:
https://docs.codecov.com/docs/deprecated-uploader-migration-guide#python-uploader

bhilbert4:
We no longer use the codecov package, but we still do have a codecov.yml file in the repo. We should remove that.

PR #1245, opened 2023-04-13
remove usages of deprecated `codecov` package

zacharyburnett:
the `codecov` package is now deprecated and has been yanked from PyPI

### DIFF
--- a/environment_python_3_10.yml
+++ b/environment_python_3_10.yml
@@ -26,7 +26,6 @@ dependencies:
   - bokeh=2.4.3
   - beautifulsoup4=4.11.2
   - celery=5.2.7
-  - codecov=2.1.12
   - cryptography=39.0.1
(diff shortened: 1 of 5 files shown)


mfixstsci (approved):
Cool -- thanks @zacharyburnett is there a recommended coverage tool we could use?

bhilbert4:
See #1244 for some notes on migrating to the new tool

zacharyburnett:
> Cool -- thanks @zacharyburnett is there a recommended coverage tool we could use?

Yes indeed; `pytest-cov` already takes care of coverage testing, and Codecov has a new `codecov-uploader`; they've just deprecated their previous Python package, so to upload coverage to Codecov we now have to either download the new binary that they distribute, or use the GitHub action in the CI workflow that does all of that for us

PR #1248, opened 2023-04-17
use `micromamba` to build environment and split job into check and tests

zacharyburnett:
the `provision-with-micromamba` action uses the small Micromamba binary to build the environment


zacharyburnett:
The `Required Status Checks` in the branch protections will need to be updated with the new job names after merge

`Python - 3.9` -> `test (Python 3.9, ubuntu-latest)`

image

zacharyburnett:
It looks like this might be the issue with using the 3.10 version:
https://github.com/spacetelescope/jwql/actions/runs/4732498267/jobs/8398751066#step:5:12
```
Error: ENOENT: no such file or directory, open '/home/runner/work/jwql/jwql/environment_python_3_0.yml'
```
It is confusing because the parsing command does use `tail -c 2` which should grab `10`

PR #1249, opened 2023-04-17
drop support for Python 3.8

### LINKED ISSUE #1098: Drop support for python 3.8
astropy will be dropping support for python 3.8 in March 2023. I'm assuming the jwst package will follow not long after. Given that, it seems that jwql should also shift to >= python 3.9 in that timeframe.

zacharyburnett:
Numpy is imminently dropping support for Python 3.8


bhilbert4:
Thanks for working on this @zacharyburnett! It's been on our To Do list for a while, but we haven't managed to get to it.

jwql/tests/test_pipeline_tools.py
```diff
@@ -67,7 +67,6 @@ def test_completed_pipeline_steps():
     assert completed_steps == true_completed
 
 
-@pytest.mark.skipif(ON_GITHUB_ACTIONS, reason='Drizzle import issue with python 3.8')
```
mfixstsci, line 70: @bhilbert4 so this test has issues with drizzle, do we know if this was solved in python versions `> 3.9`

mfixstsci (approved):

jwql/tests/test_pipeline_tools.py
bhilbert4: I don't recall what this import issue is/was. I guess we'll see what happens when the test runs. :)

zacharyburnett:
`Python - 3.8` will need to be removed from the `Required Status Checks` in the branch protections after merge:

image

zacharyburnett:
It also looks like running against the Python 3.10 environment fails due to not being able to parse `jwql-3.10` (?)

https://github.com/spacetelescope/jwql/actions/runs/4731959550/jobs/8397513039#step:8:53
```
SpecNotFound: Invalid name 'jwql-3.10', try the format: user/package
```

mfixstsci:
> It also looks like running against the Python 3.10 environment fails due to not being able to parse `jwql-3.10` (?)
> 
> https://github.com/spacetelescope/jwql/actions/runs/4731959550/jobs/8397513039#step:8:53
> 
> ```
> SpecNotFound: Invalid name 'jwql-3.10', try the format: user/package
> ```

@zacharyburnett yeah I am confused because the name in the yml file: https://github.com/spacetelescope/jwql/blob/develop/environment_python_3_10.yml#L18 looks totally fine. So I am confused. I know GHA doesn't like 3.10 but it has to be parsed a string so I wonder if that has anything to do with the parsing.

mfixstsci (approved):
Thanks again @zacharyburnett

PR #1250, opened 2023-04-17
move build configuration from `setup.py` to `pyproject.toml`

zacharyburnett:
`setuptools` now supports the `[project]` table, which is defined by [PEP621](https://peps.python.org/pep-0621).

Additionally, `setuptools` now supports its own entry in `pyproject.toml` called `[tool.setuptools]` (https://github.com/pypa/setuptools/issues/1688, https://setuptools.pypa.io/en/latest/userguide/pyproject_config.html#setuptools-specific-configuration); however, it comes with the following disclaimer:
> Support for declaring configurations not standardized by [PEP 621](https://peps.python.org/pep-0621/) (i.e. the `[tool.setuptools]` table), is still in **beta** stage and might change in future releases.

Reading `toml` is supported natively in Python 3.11 with `tomllib`

Given this, we can attempt to consolidate the build configuration into a single `pyproject.toml` file that can possibly be read by other build systems in the future.


mfixstsci (approved):
Looks good, tests are passing 👍

pyproject.toml
```diff
@@ -1,6 +1,74 @@
...
+  "jinja2",
+  "jsonschema",
+  "jwst",
+  "jwst_reffiles",
+  "matplotlib",
+  "nodejs",
+  "numpy",
+  "numpydoc",
+  "pandas",
+  "psycopg2-binary",
```
zacharyburnett, line 34: @mfixstsci Just as a heads up, I did change this dependency from `psycopg2` to `psycopg2-binary` so that an active pg installation isn't required to build the package; it seems to be a transparent change, but just in case something breaks in the future this can be changed back

mfixstsci: @zacharyburnett Thank you for the heads up! If anything comes up I will reach out.

PR #1256, opened 2023-04-19
Remove mast monitor

bhilbert4:
Remove the MAST monitor since it duplicates work done by the filesystem monitor, and the results have never been displayed.

I've removed the bokeh templating-related files. I've also turned off the cron job that runs the monitor on the servers. At the moment, I haven't deleted the monitor code itself. Looking it over, it may be useful for future MAST query work? Is it worth keeping around as a resource? @mfixstsci what do you think?


bhilbert4:

bhilbert4:
The stand-alone functions that were in monitor_mast.py I moved into mast_utils.py. Those functions were called by a number of other monitors. 

The RTD build was failing (as seen in some other PRs also). I went in and updated the names of the modules to build docs for (there were a couple references to files that don't exist.) I also had to add sphinx_rtd_theme to rtd_requirements.txt. Without it, RTD was grabbing a copy of sphinx_rtd_theme from 2019, which is what was causing the build to fail.

jwql/instrument_monitors/common_monitors/cosmic_ray_monitor.py
```diff
@@ -557,7 +557,7 @@ def process(self, file_list):
             files
         """
         for file_chunk in grouper(file_list, 100):
```
mfixstsci, line 559: Looks like some weird white space changes here.

rtd_requirements.txt
```diff
@@ -10,6 +10,7 @@ pygments==2.14.0
 pytest==7.2.1
 redis==4.5.1
 sphinx>=2
+sphinx_rtd_theme==1.2.0
```
mfixstsci, line 13: Is this something new to get the builds working?

mfixstsci (approved):
Hey @bhilbert4 everything looks great. Thanks for submitting this, just a couple of comments and we can get this in.

rtd_requirements.txt
bhilbert4: Yeah, the RTD builds were failing. Without this, the CI was installing a very old version of sphinx_rtd_theme, which was causing a crash. I saw this in a few of our other PRs as well.

jwql/instrument_monitors/common_monitors/cosmic_ray_monitor.py
bhilbert4: I think that's just from sublime text. When I save a file, it automatically removes extra whitespace at the ends of lines. 

I'm hoping it won't be too much longer before we can delete the CR monitor anyway....

rtd_requirements.txt
mfixstsci: Awesome, thank you for catching that.

jwql/instrument_monitors/common_monitors/cosmic_ray_monitor.py
mfixstsci: Agreed about removing the monitor. I was unsure if they were editor based or intentional. Thanks!

mfixstsci:
@bhilbert4 I ran the tests locally and everything passed. Thanks for responding to my comments, merging!

PR #1258, opened 2023-04-25
Fix path to pyproject.toml

bhilbert4:
Fixes an issue with #1250 
In that PR, the full path to setup.py was replaced with pyproject.toml. But without the full path, the logging function was not finding pyproject.toml, and crashing. This PR restores the full path to pyproject.toml


bhilbert4:
@mfixstsci this is ready for review. This should fix the failing monitors on the servers.

BradleySappington (approved):
looks good!

BradleySappington:
Thanks for the quick fix @bhilbert4!

PR #1259, opened 2023-04-25
Dark monitor threshold and cadence updates

### LINKED ISSUE #1252: Dark monitor enhacements
```[tasklist]
### Tasks
- [ ] https://github.com/spacetelescope/jwql/issues/1253
- [ ] https://github.com/spacetelescope/jwql/issues/1254
```

### LINKED ISSUE #1253: Count dark integrations rather than exposures

### LINKED ISSUE #1254: Add ability to specify date range and have the monitor run using data from those dates

bhilbert4:
Make some updates to the thresholds used to trigger a run of the dark monitor. Previously the monitor would only run when N files of a given instrument/aperture were found. This PR updates that to be N integrations (because some teams take dark calibration data with multiple integrations). 

There are also some updates to the way the files are grouped, in order to support a case where thresholds have changed enough that the monitor will be re-run across data spanning a broad range in time. The date of the input files are first examined, and the files are split into groups based on the amount of time between consecutive files. A gap larger than N days is interpreted as an epoch boundary, and the monitor will not run on files that span more than 1 epoch. The files for each epoch are then further subdivided by the integration threshold described above. Since we cannot split a single file between monitor runs, the threshold value is used as a minimum, with the exception of the case where it would need to cross epochs. 

For example, if the threshold value is 15 integrations, and the new files for one epoch have these numbers of integrations:
[1, 3, 1, 5, 10, 5, 7, 10, 5, 5, 2]
Then the code will divide up the files as such:
[1, 3, 1, 5, 10] = 20 total integrations, [5, 7, 10] = 22 total integrations, [5, 5, 2] = 12 total integrations
Even though the final of the three groups is under the threshold of 15 integrations, the monitor will run on the 12 integrations.

Also, for MIRI, the first M integrations of each file are skipped, as the dark current is not stable in the initial M integrations. At the moment we use M=1 for full frame data.


bhilbert4:
- [x] For MIRI, we need to ignore the first N integrations in each file

bhilbert4:
Initial test on the dev server looks good. I commented out all database interactions and the actual analysis code. So the test just queried MAST from launch to May 2023, and then divided up the files using the new scheme, and put together what would be the new database entries.

bhilbert4:
Left the monitor running (with no db interactions nor analysis) on the dev server. See log file ending in 2023-05-02-17-02.log

bhilbert4:
Two other small tweaks to make:

- [x] Add a date to the histogram plots describing when the data used to create them is from
- [x] Cut down the list of apertures examined. Keep full frame and just a couple subarrays per instrument?
- [x] Make sure the dates added to the database are those for the used files, rather than the overall query start and stop
- [x] Make sure the amp-specific data is being correctly populated/retrieved from the database

bhilbert4:
What do we do about varying numbers of groups? We probably don't want to be creating a mean dark rate image by averaging 10-group integrations with 150-group integrations. The three options for getting around this problem are:
1. Filter out and ignore all files with ngroups < some threshold
2. Keep the integration threshold set to 1, so that each file is run on its own.
3. Making the sublist generator more complex, such that exposures with very different numbers of groups are kept separate

The third option seems like it would be the most difficult to implement. 

The second option makes the most sense to me. We don't want JWQL to filter out too much data, even if that data gives noisy results. I think as long as users can get a good idea of the number of groups/ints associated with each data point, then having all the data present would be ok. Also, for NIRCam at least, dark files are taken very infrequently, meaning that the monitor is going to run on one exposure at a time regardless of the number of groups or integration threshold.

Filtering out all data with not enough groups would be the easiest to implement. And it would make sense in terms of ignoring the early buffer flush observations and any short darks that users take in order to look for persistence, etc. But again, unless we set the limit very high, we could still end up with cases where integrations with very different numbers of groups are averaged together in order to make a mean dark current rate image. 

So maybe we should do a combination of the first and second options. Filter out exposures with very few groups (e.g. <10), but then keep the one integration threshold, such that each file is run on its own.

jwql/instrument_monitors/common_monitors/dark_monitor.py
```diff
@@ -1020,43 +1091,73 @@ def run(self):
                                 copied_size = os.stat(dark_file).st_size
                                 orig_size = os.stat(filesystem_path(os.path.basename(dark_file))).st_size
                                 if orig_size != copied_size:
```
bhilbert4, line 1093: I'm wondering about this. I want to catch cases where there's some major problem with the copying. I've seen cases where the copying stops partway through these large dark files. In that case we need to know. But I've also seen cases where the copied file size is just slightly different than the size of the original file, and in those cases the calculations all seem to work. Maybe some sort of metadata difference or something, which doesn't affect the actual data. But also, darks are generally taken rarely enough that if there's a copying problem, we probably want to do something to use the file in the future. (The next monitor run? Or try again in the current run?)

jwql/tests/test_dark_monitor.py
```diff
@@ -28,11 +28,151 @@
 from jwql.instrument_monitors.common_monitors import dark_monitor
 from jwql.tests.resources import has_test_db
 from jwql.utils.monitor_utils import mast_query_darks
+from jwql.utils.constants import DARK_MONITOR_BETWEEN_EPOCH_THRESHOLD_TIME
 from jwql.utils.utils import get_config
 
 ON_GITHUB_ACTIONS = '/home/runner' in os.path.expanduser('~') or '/Users/runner' in os.path.expanduser('~')
 
 
+def generate_data_for_file_splitting_test():
```
mfixstsci, line 36: Thanks for writing tests!

jwql/utils/constants.py
```diff
@@ -205,6 +205,18 @@
...
 DARK_MONITOR_BADPIX_TYPES = ["hot", "dead", "noisy"]
 
+# Minimum amount of time, in days, between epochs of dark current observations. If the
+# dark monitor sees this much time, or longer, between two dark current files, it assumes
+# that the two files are part of separate epochs. This means the monitor will run separately
+# on these files, rather than bundling them together into a batch, where they would have
+# been combined into a mean dark rate
+DARK_MONITOR_BETWEEN_EPOCH_THRESHOLD_TIME = {'nircam': 10.,
+                                             'niriss': 10.,
+                                             'miri': 0.00001,  # Treat each MIRI exposure separately
```
mfixstsci, line 217: Just looked this up, MIRI epochs are `0.864` seconds apart 😆

jwql/instrument_monitors/common_monitors/dark_monitor.py
```diff
@@ -432,14 +441,30 @@ def exclude_existing_badpix(self, badpix, pixel_type):
...
         logging.info("\t\tKeeping {} {} pixels".format(len(new_pixels_x), pixel_type))
-#             pixel = (x, y)
-#             if pixel not in already_found:
-#                 new_pixels_x.append(x)
-#                 new_pixels_y.append(y)
 
         session.close()
         return (new_pixels_x, new_pixels_y)
 
+    def exclude_too_few_groups(self, result_list):
```
mfixstsci, line 448: This is really good, thanks for this addition.

jwql/instrument_monitors/common_monitors/dark_monitor.py
mfixstsci: Agreed, I remember when we found this problem, it's a strange one. I think it's something we keep in mind in the future.

mfixstsci (approved):
Hey @bhilbert4 thanks for testing this out and contributing all of these updates. Looks like a lot of work went into this. From what I see, everything looks good to me. Thanks for adding documentation and tests!

jwql/utils/constants.py
bhilbert4: MIRI said that each file should always be considered on its own. No averaging multiple files. This seemed like the easiest way to do that. For NIRCam, we only take one exposure in each observation, so the time here doesn't really matter much. It'll be one file at a time there as well.

mfixstsci: Definitely makes sense, I was just curious what the time in seconds would be so I am sure we will catch each individual file with this rate!

PR #1262, opened 2023-05-08
Pipeline skip already run steps

### LINKED ISSUE #1260: Add ability to call/omit specific steps of the pipeline
In shared_tasks/run_pipeline.py, the code assumes that all steps of calwebb_detector1 should be run on a given input file (with the exception of ramp fitting). 

For the dark monitor, we want to skip the dark_current step. Further, there are _dark.fits files in the filesystem. Starting from these, only the jump and ramp-fitting steps would have to be run, which would save time compared to running the entire pipeline on the uncal version of these files. 

It would be helpful to tweak the code such that lists of pipeline steps to skip/run could be provided, and the pipeline call would be updated accordingly. 

Finally, at the moment the code is using the run() method of the pipeline. We should switch to using the call() method so that the pipeline will use the appropriate parameter reference files for parameters that we don't explicitly specify. This would involve switching the parameters we do specify to use a nested dictionary.

e.g.:
parameters = {'dark_current': {"skip": True}}
p = Detector1Pipeline.call(uncal_file, steps=parameters}

FYI @york-stsci

bhilbert4:
This PR is motivated by updates to the dark monitor. In input data for the dark monitor are dark.fits files. However, the pipeline calls in our run_pipeline.py skip at the moment all assume that the input file is an uncal.fits file, and that all calwebb_detector1 pipeline steps should be run.

This PR updates the pipeline calls such that for dark.fits input files, only the jump and ramp-fitting steps are run. This will save time and disk space, since the pipeline call for the dark data at the moment saves the output from each step of the pipeline.


bhilbert4:
This PR also switches the pipeline calls to use the call() method, rather than the run() method of the pipeline. This will allow the values of unspecified parameters to be retrieved from parameter reference files, rather than falling back to the defaults that are hardcoded.

bhilbert4:
Testing on dev appears successful. The dark monitor successfully grabbed *_dark.fits files, ran the appropriate pipeline steps, and saved the rate.fits files. This PR also adds logic to allow the rateints files to be saved. Earlier testing showed that to be working. That aspect of the code is turned off in this PR though. A future PR that re-works how the dark monitor groups input files will also turn that on.

@york-stsci since you are the most familiar with the celery functions, it might be helpful if you gave the changes a quick look over as well.

jwql/shared_tasks/run_pipeline.py
```diff
@@ -43,27 +44,65 @@ def run_pipe(input_file, short_name, work_directory, instrument, outputs, max_co
...
         copy_files([input_file], work_directory)
         set_permissions(uncal_file)
-    
+
         steps = get_pipeline_steps(instrument)
+
+        # If the input file is a file other than uncal.fits, then we may only need to run a
+        # subset of steps. Check the completed steps in the input file. Find the latest step
+        # that has been completed, and skip that plus all prior steps
+        if 'uncal.fits' not in input_file:
```
mfixstsci, line 64: This is clever, thanks for checking this. Hopefully it will save us some time!

jwql/shared_tasks/run_pipeline.py
```diff
@@ -157,51 +207,73 @@ def run_save_jump(input_file, short_name, work_directory, instrument, ramp_fit=T
...
+        params = {}
 
         # Always true
         if instrument == 'nircam':
-            model.refpix.odd_even_rows = False
+            params['refpix'] = dict(odd_even_rows=False)
 
         # Default CR rejection threshold is too low
-        model.jump.rejection_threshold = 15
+        #params['jump'] = dict(rejection_threshold=15)
```
mfixstsci: Do we need to keep this commented line here?

jwql/shared_tasks/run_pipeline.py
```diff
@@ -157,51 +207,73 @@ def run_save_jump(input_file, short_name, work_directory, instrument, ramp_fit=T
...
+        # Include any user-specified parameters
+        for step_name in step_args:
+            if step_name in params:
+                params[step_name] = params[step_name].update(step_args[step_name])
+            else:
+                params[step_name] = step_args[step_name]
+
         if run_jump or (ramp_fit and run_slope) or (save_fitopt and run_fitopt):
-            model.run(datamodel)
+            model.call(datamodel, output_dir=work_directory, steps=params)
```
mfixstsci, line 275: Just to make sure I have my head on straight, we aren't changing the `detectorOnePipeline` model parameters directly but instead gathering the parameters from the data/using our knowledge of what steps have been run based on file extension and then passing the steps we need to run as an argument at call time?

mfixstsci (commented):
Hey @bhilbert4 everything is looking great! Some comments here, I am waiting on approval because I would like to hear from @york-stsci first.

jwql/shared_tasks/shared_tasks.py
```diff
@@ -528,7 +569,7 @@ def prep_file(input_file, in_ext):
     input_path, input_name = os.path.split(input_file)
     logging.info("\tPath is {}, file is {}".format(input_path, input_name))
 
-    if "uncal" not in in_ext:
```
york-stsci, line 531: I feel as if we can probably remove this completely, now that the pipeline has been modified to (as far as I can see) be callable starting with any input file.

jwql/shared_tasks/run_pipeline.py
```diff
@@ -244,6 +317,7 @@ def run_save_jump(input_file, short_name, work_directory, instrument, ramp_fit=T
     parser.add_argument('input_file', metavar='FILE', type=str, help=file_help)
     parser.add_argument('short_name', metavar='NAME', type=str, help=name_help)
     parser.add_argument('max_cores', metavar='CORES', type=str, help=cores_help)
+    parser.add_argument('--step_args', metavar='STEP_ARGS', type=json.loads, default='{}', help=step_args_help)
```
york-stsci, line 319: In the shared tasks log, it looks like there's a consistent warning message:

```
07/13/2023 15:45:59 PM WARNING: step_args: {'dark_current': {'skip': True}}
07/13/2023 15:45:59 PM WARNING: step_args_str: '{"dark_current":{"skip":"True"}}'
07/13/2023 15:45:59 PM WARNING: expected: {"dark_current": {"skip";"True"}}
```

So, based on the above, I'm not sure if the arguments are getting through un-mangled. Which in turn raises the questions:

- Is skipping the dark step actually working right now?
- If it is, is it working 'by accident' (e.g. because the string 'true' evaluates to the bool True, but the string 'false' would also evaluate to True, so we're not actually sending a boolean here)

jwql/shared_tasks/shared_tasks.py
```diff
@@ -224,9 +224,43 @@ def collect_after_task(**kwargs):
     gc.collect()
 
 
-def run_subprocess(name, cmd, outputs, cal_dir, ins, in_file, short_name, res_file, cores):
-    command = "{} {} {} '{}' {} {} {} {}"
-    command = command.format(name, cmd, outputs, cal_dir, ins, in_file, short_name, cores)
+def convert_step_args_to_string(args_dict):
```
york-stsci, line 227: Actually, looking at this function, I have what might be a naive question – if we're using `json.loads()` to parse the command-line argument, is there a reason we're not using `json.dumps()` to encode it?

york-stsci:
I think the above has my thoughts on this. In general it looks good, and the shared tasks log shows that it's running through celery in the way that we would expect.

jwql/shared_tasks/run_pipeline.py
bhilbert4: I'm unsure about this one. The default value for the parameter is 3. If you talk to some people, they say that mathmatically, 3 should be correct. On the other hand, I've seen a few help desk calls where people say the jump step is flagging too many pixels as jumps. That it's flagging noise rather than real jumps. And I assume that most users don't even check their data to see what was flagged.

jwql/shared_tasks/run_pipeline.py
bhilbert4: That's right. The list of what steps to skip and what steps to run is determined at run time based on what's in the header of the file.

jwql/shared_tasks/run_pipeline.py
york-stsci: I know that the cosmic ray monitor used this version of the pipeline. I think the bad pixel monitor is the other one that uses it. Whether it's the right change to make depends on the goal of the monitor, and I don't actually know *why* the bad pixel monitor was doing this, in particular.

jwql/shared_tasks/run_pipeline.py
bhilbert4: I'll check on this. The last two of those warning lines were manual logging entires that I put into shared tasks to track that step_args was being passed correctly. I did spend time playing around with this, because there were 2 options for passing the parameters. json.loads seemed to need to have quotes around the `True` in order to interpret it as a boolean once it had been passed. The other option was to use ast.literal_eval, in which case `True` had to not have quotes around it. I'll re-confirm though.

york-stsci: Fair. If it is working, then probably we should remove the logging lines, because they give the impression that something unexpected is happening, and I don't want someone to be tracking down an unrelated issue and have to figure out where this warning message is coming from and why it's there.

jwql/shared_tasks/run_pipeline.py
bhilbert4: For the bad pixel monitor, I think it was to avoid over-flagging as well. Given that this PR makes it possible for instrument monitors to pass parameter values to the pipeline calls, maybe it makes more sense to remove this parameter here, and have each monitor pass whatever parameters are appropriate? And the CR monitor is coming back at some point, but in a very different format that doesn't involve any pipeline calls.

jwql/shared_tasks/shared_tasks.py
bhilbert4: Good question. I can compare the output from that and remove this function if it's not needed.

jwql/shared_tasks/run_pipeline.py
york-stsci: That's a good question. I guess the question is whether the save_jump pipeline is asking for different things, or just asking for custom parameters. I think, at the time, that it was using `call()` when the standard pipeline function in JWQL was using `run()` made me reluctant to try to combine together, but if we're at a point where we can just have one pipeline function, it will make the code path a lot easier (I think).

jwql/shared_tasks/shared_tasks.py
bhilbert4: I didn't test with any input file other than dark.fits. Actually though, the only possible input files, since they all ultimately come from MAST, are uncal and dark. Beyond those, any other input files (rate, cal, etc) would never need to have the calwebb_detector1 pipeline called at all.

jwql/shared_tasks/run_pipeline.py
bhilbert4: It does seem like the only difference is the custom parameters. Using `run()` is frowned upon because parameters that are not explicitly set by the user fall back to default values that are not optimal. With `call()`, unspecified parameters are grabbed from parameter reference files, which contain better values.

york-stsci: Fair. I expect that the functions in the JWST pipeline_tools.py file may have been written before all of that was specified.

jwql/shared_tasks/run_pipeline.py
bhilbert4: I pulled out the logging lines a few days ago. You'll see in the latest few runs of the monitor that they are not present.

bhilbert4: Hmm, wait. I do see them. I'll look into it.

jwql/shared_tasks/shared_tasks.py
bhilbert4: Looks like `json.loads()` works. I'll remove this function and switch to using that.

jwql/shared_tasks/shared_tasks.py
bhilbert4: Thinking a little more deeply about this, I agree. I think it's safe to remove this block completely.

jwql/shared_tasks/run_pipeline.py
```diff
@@ -43,27 +44,65 @@ def run_pipe(input_file, short_name, work_directory, instrument, outputs, max_co
...
+        for step, step_dict in step_args.items():
+            if 'skip' in step_dict:
+                print(f'SKIP the {step}!!')
+                if step_dict['skip']:
+                    steps[step] = False
+
+        # Run each specified step
         first_step_to_be_run = True
         for step_name in steps:
-            sys.stderr.write("Running step {}\n".format(step_name))
```
bhilbert4, line 60: @york-stsci does it make sense to add an entry about running a pipeline step to stderr?

york-stsci: I'm not sure it needs to be a write to stderr, but it should write using one of STDOUT or STDERR, because the shared tasks function(s) get log information from the subprocess by redirecting STDOUT, STDERR, or both to a pipe, and then redirecting that pipe to a log (all in the shared_tasks.py file).

jwql/shared_tasks/shared_tasks.py
bhilbert4: I take that back. The previous results were deceiving because I hadn't restarted the celery worker. When I do that, `json.dumps()` does not appear to work for translating the dictionary to the necessary string.

york-stsci: That's annoying. Also, it kind of feels like a python issue if taking a dictionary, turning it into a string with `json.dumps()`, and turning it back into a dictionary with `json.loads()` *doesn't* give back the same dictionary.

bhilbert4: I'm starting with the dictionary:
step_args = {"dark_current": {"skip":True}}

json.dumps turns that into: 
'{"dark_current": {"skip": true}}'

But then in the command that is pieced together to call the pipeline, it looks like that is being broken up into multiple strings:
'--step_args', '{dark_current:', '{skip:', 'true}}'

I might just go back to the function I had written, which was doing the conversion correctly.

jwql/shared_tasks/run_pipeline.py
york-stsci: Okay, looking at it, in the file "shared_tasks.py",

- At line 202, function `log_subprocess_output()` adds to the log file anything that comes out of the pipe produced by calling `popen()` on the subprocess
- In `run_subprocess()`, lines 235-237, when calling `subprocess.Popen()`, the standard error output from the subprocess is sent to pipe, and `log_subprocess_output()` is called on `process.stderr`

Now, the reason that I'm doing this is to get the pipeline logging, because the JWST pipeline prints all of its output, whatever the source, to STDERR rather than STDOUT, presumably for *some* reason, but I don't know what the reason actually *is*.

jwql/shared_tasks/shared_tasks.py
york-stsci: Looking at the file, it seems as if, in `run_subprocess()`, you're calling `json.loads()` on the dictionary, which seems as if it might the the opposite of what you want to be doing?

bhilbert4: Ignore that. That's a typo. I've been testing on the server using json.dumps()

york-stsci:
I feel as if, given the set of changes we've been making, it's probably worth taking this PR and putting it on test and doing a run there, just to be trying again on a clean system and make sure that we're not missing anything.

jwql/shared_tasks/shared_tasks.py
bhilbert4: Removed.

jwql/shared_tasks/run_pipeline.py
bhilbert4: After restarting celery, these logging entries are now gone, as expected.

bhilbert4:
I agree. I've re-implemented my function for the dictionary -> string conversion and am going to run another test on the dev server. Assuming that is successful, then I think testing this on the test server makes sense.

jwql/shared_tasks/run_pipeline.py
york-stsci: To me, this seems like a good change to make, but probably not in this PR. I feel as if it would add another big test burden, because we'd also have to start testing a bunch of files with the bad pixel monitor to make sure things are running the same. Assuming you approve, my inclination is to file an issue to combine these functions into a single function, and then resolve the conversation here. Then, once we've merged this in, I can work that issue.

bhilbert4:
@mfixstsci @york-stsci Ok, everything seems to be working well on the dev server. I've re-added the line with the jump step threshold. It would be nice to remove that in a separate PR in the not too distant future. We could always update the bad pixel monitor at the same time to have it pass the same parameter to its pipeline call, so nothing would change there. I just don't like the idea of forcing all monitors that use the current pipeline functions to not be able to adjust the parameter value.

york-stsci:
> @mfixstsci @york-stsci Ok, everything seems to be working well on the dev server. I've re-added the line with the jump step threshold. It would be nice to remove that in a separate PR in the not too distant future. We could always update the bad pixel monitor at the same time to have it pass the same parameter to its pipeline call, so nothing would change there. I just don't like the idea of forcing all monitors that use the current pipeline functions to not be able to adjust the parameter value.

I don't think we're currently forcing anything, because as far as I can tell the only monitor that calls the save_jump pipeline is the bad pixel monitor, and the parameters that are set are specifically the ones that the bad pixel monitor wants to set. That said, I agree that we should go down to a single pipeline function, especially since you now have a way to set custom parameters on a per-step basis when doing that.

I'm creating an issue now to do this, and once we have the capability to set per-monitor pipeline parameters (which also wasn't really something we had before we started using celery, because the directly-calling-the-pipeline version *also* didn't have a free way to set pipeline parameters), we might want to actually go through the common monitors (since those are the ones that call the pipeline in the first place), and ask whether we could get better results if we *did* start to use custom parameters.

york-stsci:
As just a last question, when you say that rateints support is present but not turned on, do you mean that "if you ask for a rateints file, you'll get one, but the dark monitor doesn't currently ask for them?" or "you can't currently get one, but we can uncomment some lines, and then you'll be able to"?

bhilbert4:
@york-stsci, the former. I made some updates that allow rateints to be specified as an output that is saved and moved into the output directory. Those updates worked in my dark monitor testing last week where I temporarily changed the dark monitor to ask for them. I have since changed the dark monitor back to ask for rate files, as it has been doing. In #1259 I have made more substantial changes to the dark monitor to allow it to actually open and work with rateints files. My plan is to update that branch with the changes in this branch (once this PR is merged), and to switch the dark monitor to ask for rateints files in there. Then everything will be set for an end to end test using rateints files.

jwql/shared_tasks/run_pipeline.py
mfixstsci: Thanks @york-stsci so for logging purposes should we go ahead and add the line back in?

mfixstsci: Ahh nevermind, I just took a look at the file. Looks like the line is currently there from the most recent changes.

york-stsci: Yes, I would say leave it in.

york-stsci:
In that case, we probably want to add to the general notes that, once this PR goes in, *other* monitors can also ask for rateints files, in case that would be useful.

york-stsci (approved):
As (extensively) annotated in the conversation, looks good to me.

mfixstsci (approved):

PR #1267, opened 2023-05-22
May/June Environment Updates

mfixstsci:


melanieclarke (commented):
The new environment files look good to me!  Both of them build, look fine in spot checks, and mostly pass local testing.

Two things though...
1. I think the requirements.txt and rtd-requirements.txt files need to be updated to match the updated dependency versions.
2. There are a couple minor local testing issues related to the recent packaging changes (below). They're unrelated to the environment changes here, so I can add a separate PR for it if you want.

Testing issues:
- test_setup.py fails due to changes in the way versioning is done
- pyproject.toml is not setup correctly to exclude the static directory, so test discovery is not working right -- 'pytest jwql' attempts to recurse the entire filesystem to find unit tests.

mfixstsci:
> The new environment files look good to me! Both of them build, look fine in spot checks, and mostly pass local testing.
> 
> Two things though...
> 
>     1. I think the requirements.txt and rtd-requirements.txt files need to be updated to match the updated dependency versions.
> 
>     2. There are a couple minor local testing issues related to the recent packaging changes (below). They're unrelated to the environment changes here, so I can add a separate PR for it if you want.
> 
> 
> Testing issues:
> 
>     * test_setup.py fails due to changes in the way versioning is done
> 
>     * pyproject.toml is not setup correctly to exclude the static directory, so test discovery is not working right -- 'pytest jwql' attempts to recurse the entire filesystem to find unit tests.

Thanks for the review @melanieclarke! I forgot about the requirement files 😓 I will go ahead and update those. As for the the `pyproject.toml` files I am unaware of how to set them up so if you have some tips you can give me, that would be excellent. I am still learning about the new packaging configuration we are working with!

melanieclarke:
> As for the the `pyproject.toml` files I am unaware of how to set them up so if you have some tips you can give me, that would be excellent. I am still learning about the new packaging configuration we are working with!

Still learning too!  I don't know if this is the best way, but testing locally, it looks like it works.  In pyproject.toml, remove the `addopts` line under pytest and replace it with a `norecursedirs` option, like this:
```
[tool.pytest]
junit_family = "xunit2"

[tool.pytest.ini_options]
norecursedirs = ["jwql/website/apps/jwql/static"]
```

mfixstsci:
@melanieclarke thanks, I went ahead and made the changes.

melanieclarke (approved):
Looks good now.  Thanks for the updates!

PR #1276, opened 2023-05-31
Update lock_module decorator to handle abandoned locks

### LINKED ISSUE #1275: Automated check for failed jobs
We've seen cases where one of the JWQL cron jobs was interrupted or failed in some way, and the corresponding lock file was not removed. This means that subsequent runs of the job won't be able to start. Currently when this happens, the developers are not alerted in any way, and only find out when someone goes looking for the outputs from the job.

It would be nice to implement an automated check for interrupted/failed jobs, and alert the developers when it happens. 

For archive_database_update.py and generate_preview_images.py (i.e. jobs that do not use a celery/redis lock, but instead a manually constructed lock file), we could set up a job to look at any lock files in the directory and check their age. Any lock files older than some threshold may indicate a problem and should trigger a notification to the developers.

What would be a good threshold to use for this? It seems like even just 24 hours would be a conservative estimate. For normal operations, archive_database_update and generate_preview_images both complete in significantly less time than 24 hours. And in special cases where those tasks are catching up on a backlog of data and do take 24 hours, the developers would already know about the situation and could ignore the notification.

What about tasks that use the redis lock? I'm assuming there is some way to check the age of a given lock...
@york-stsci 

@melanieclarke @mfixstsci

BradleySappington:
Update @lock_module decorator

When any protected module kicks off, it will check if the lock exists.
If it DOES exist, it will extract the PID from the file and do the following
If the PID is currently running
    The command will exit because its locked. 
If the PID isn’t running
    It will delete the erroneous lock file
    It will email [jwql@stsci.edu](mailto:jwql@stsci.edu) to let us know that something happened with the previous run (no 
action needed on our end, but will help us answer questions from scientists as to why their data is available)
    It will start a fresh run
If there is a bad lock file without a PID in it
    We will receive an email with instructions to manually investigate and the process will NOT currently run (this allows us to manually lock something for any reason we want by creating an empty lock file)  We will also get emailed every time until the bad lock file is removed so we wont forget about it.


BradleySappington:
Ready for review

melanieclarke (commented):
Testing locally, changing the notification address to my email, this looks good.  It behaved appropriately for all conditions I checked:
- missing PID sent email
- running after a killed process sent email
- a successfully completed process removed its lock file
- running again while previous process was running did nothing

However, the test in test_protect_module.py needs to be modified.  It currently sends a "No PID found" email every time it runs.

BradleySappington:
@melanieclarke thanks for the test reminder!

BradleySappington:
@melanieclarke - tests updated

jwql/tests/test_protect_module.py
```diff
@@ -13,13 +13,19 @@
     These tests can be run via the command line (omit the -s to
     suppress verbose output to stdout):
 
+    !!!WARNING IF TESTING LOCALLY!!!
+    These tests will call protect_module.py::lock_module.py and will sent emails to jwql.stsci.edu!
+    Please locally change the `message['To']` recipient to an appropriate address while testing to
+    avoid spamming/confusing the team.
+
+
```
melanieclarke: I like these changes, but I think this warning isn't sufficient to prevent regularly spamming the jwql list.  Can we make the email address configurable instead of hard-coded, and change it to a null address for testing?   Maybe make it a constant value at the top of the protect_module file? Or in the general jwql constants module?

melanieclarke: Looks good now! Thanks for the fix.

melanieclarke (approved):

mfixstsci:
Thanks for reviewing and approving @melanieclarke. @BradleySappington are you okay with me updating the branch and merging? Do you have any other features to add?

BradleySappington:
@mfixstsci - This is ready for merge

PR #1240, opened 2023-04-07
Reformat query page functionality

BradleySappington:
Reformat how the query page works to access using the django model DB

PR #1281, opened 2023-06-06
Improve general JWQL queries

### LINKED ISSUE #1247: Add viewed status and proposal category to query page
Now that the query page is working better, it would be nice to add a few more possible fields.  In particular, it would be nice to show only new/viewed files and filter by proposal category.  These are already available in the archive interface, but would be nice to have in the query page too.

melanieclarke:
This PR adds some new general query fields to the top of the JWQL query page: proposal category and look status

I also added some basic server-side pagination for broad queries.  Pagination is controlled by the sort type and results per page fields, also at the top of the query page.


melanieclarke:
This should resolve the "insufficient resources" issue with broad queries, noted in https://github.com/spacetelescope/jwql/pull/1240.

BradleySappington (approved):
Great Work!

BradleySappington (approved):

PR #1282, opened 2023-06-14
Improve explore image

### LINKED ISSUE #1266: Add row/column plots based on mouse position in Explore Image
A request from NIRSpec: 

Similar to what is implemented in ds9, add plots of the signal in the row and column under the mouse when on the Explore Image page.

Seems like a case for Bokeh js callback.

melanieclarke:
A few new improvements for the explore image page:

- configure zoom and figure size to make sure pixels stay square
- move scale controls to client side for faster response
- implement some useful default scale settings
- move hover tool display to a separate div and add RA/Dec if applicable
- add optional column and row plots alongside the image, showing a cross-section of the data at a clicked position in the image

### DIFF
--- a/jwql/utils/interactive_preview_image.py
+++ b/jwql/utils/interactive_preview_image.py
@@ -454,7 +454,13 @@ class InteractivePreviewImg:
                     val = s.data['dq'][0][y][x].join(', ');
                     label = "Value";
                 } else {
-                    var val = s.data['image'][0][y][x];
+                    // get the data from the array of arrays
+                    val = s.data['image'][0][y][x];
+                    if (val === undefined) {
+                        // uncal images have to be addressed with the flat index
+                        val = s.data['image'][0][flat];
+                    }
+                    // report any non-number as NaN
                     if (!Number.isFinite(val)) {
                         val = 'NaN';
                     } else {
--- a/jwql/utils/interactive_preview_image.py
+++ b/jwql/utils/interactive_preview_image.py
@@ -16,6 +16,7 @@
         Required Arguments:
         ''filename'' - Name of a fits file containing a JWST observation
     """
+import datetime
 from copy import deepcopy
 import os
 
@@ -23,12 +24,13 @@ from astropy.io import fits
 from astropy.visualization import ZScaleInterval, MinMaxInterval, PercentileInterval
 import numpy as np
 from bokeh.embed import components
-from bokeh.layouts import layout
+from bokeh.layouts import gridplot, layout
 from bokeh.models import (
-    BasicTicker, BoxZoomTool, Button, ColorBar, CustomJS, Div, HoverTool,
-    LinearColorMapper, LogColorMapper, LogTicker, RadioGroup, Row, Select, Spacer,
-    Spinner, WheelZoomTool)
+    BasicTicker, BoxZoomTool, Button, ColorBar, ColumnDataSource,
+    CustomJS, CustomJSTransform, DataRange1d, Div, HoverTool, LinearColorMapper, LogColorMapper,
+    LogTicker, RadioGroup, Range1d, Row, Select, Spacer, Spinner, WheelZoomTool)
 from bokeh.plotting import figure, output_file, show, save
+from bokeh.transform import transform
 
 from jwst.datamodels import dqflags
 
@@ -37,8 +39,8 @@ class InteractivePreviewImg:
     """Class to create the interactive Bokeh figure.
     """
 
-    def __init__(self, filename, low_lim=None, high_lim=None, scaling='log', contrast=None, extname='SCI',
-                 group=None, integ=None, mask=None, save_html=None, show=False):
+    def __init__(self, filename, low_lim=None, high_lim=None, scaling='lin', contrast=None, extname='SCI',
+                 group=None, integ=None, mask=None, line_plots=False, save_html=None, show=False):
         """Populate attributes, read in data, and create the Bokeh figure
         Parameters
         ----------
@@ -68,6 +70,9 @@ class InteractivePreviewImg:
         mask : numpy.ndarray
             Mask to use in order to avoid some pixels when auto-scaling. Pixels with a value other than 0 will
             be ignored when auto-scaling.
+        line_plots : bool
+            If set, column and row plots are added to the layout, to be updated on click in the main figure.
+            These take some time to create, so are off by default.
         save_html : str
             Name of html file to save the figure to. If None, the components are returned instead.
         show : bool
@@ -80,6 +85,7 @@ class InteractivePreviewImg:
         self.contrast = contrast
         self.extname = extname.upper()
         self.mask = mask
+        self.show_line_plots = line_plots
         self.show = show
         self.save_html = save_html
 
@@ -105,6 +111,8 @@ class InteractivePreviewImg:
         self.get_data()
         if 'DQ' in self.extname:
             self.get_bits()
+            # col/row plots not available for dq values
+            self.show_line_plots = False
         self.script, self.div = self.create_bokeh_image()
 
     def create_bokeh_image(self):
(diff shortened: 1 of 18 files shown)


jwql/utils/interactive_preview_image.py
```diff
@@ -19,22 +19,26 @@
...
 
 
-class InteractivePreviewImg():
+class InteractivePreviewImg:
     """Class to create the interactive Bokeh figure.
     """
 
-    def __init__(self, filename, low_lim=None, high_lim=None, scaling='log', contrast=None, extname='SCI',
-                 group=None, integ=None, mask=None, save_html=None, show=False):
+    def __init__(self, filename, low_lim=None, high_lim=None, scaling='lin', contrast=None, extname='SCI',
```
bhilbert4, line 40: Did you find that defaulting to a linear rather than log scale gives a better initial image more often?

jwql/utils/interactive_preview_image.py
```diff
@@ -98,63 +106,485 @@ def __init__(self, filename, low_lim=None, high_lim=None, scaling='log', contras
...
         self.integ = integ
+
+        self.data = None
+        self.signal_units = None
+        self.wcs_coord = None
         self.get_data()
         if 'DQ' in self.extname:
             self.get_bits()
+            # col/row plots not available for dq values
+            self.show_line_plots = False
```
bhilbert4, line 117: Good catch.

jwql/utils/interactive_preview_image.py
```diff
@@ -98,63 +106,485 @@ def __init__(self, filename, low_lim=None, high_lim=None, scaling='log', contras
...
-            hover_tool = HoverTool(tooltips=[("(x,y)", "($x{0.2f}, $y{0.2f})"),
-                                             ('Value', '@image{0.4f}')
-                                             ], mode='mouse', renderers=[img])
+
+        # fix figure aspect from data aspect
+        # bokeh throws errors if plot is too small, so make sure
+        # the smaller dimension has reasonable size
+        max_dim, min_dim = 700, 400
+        if xd > yd:
+            plot_width = max_dim
```
bhilbert4, line 160: Nice!

jwql/website/apps/jwql/views.py
```diff
@@ -1008,7 +1008,7 @@ def explore_image_ajax(request, inst, file_root, filetype, scaling="log", low_li
         FITS file_root of selected image in filesystem
     filetype : str
         Type of file (e.g. ``uncal``)
-    scaling : str
+    line_plots : str
         Scaling to implement in interactive preview image ("log" or "lin")
```
bhilbert4: This parameter description line is still referring to the 'scaling' parameter

jwql/utils/interactive_preview_image.py
```diff
@@ -98,63 +106,485 @@ def __init__(self, filename, low_lim=None, high_lim=None, scaling='log', contras
...
+                              "</div>"
+                }
+                d.text += "<div style='display:table-row'>" + 
+                          "<div style='display:table-cell; text-align:right'>Value (" + u + ") =</div>" +
+                          "<div style='display:table-cell'>" + val + "</div></div></div></div>";
+            } else {
+                d.text = "";
+            }
+        """)
+        if 'DQ' not in self.extname:
```
bhilbert4: Is there a difference between what's done in the "if" and "else" blocks?

jwql/utils/interactive_preview_image.py
```diff
@@ -213,6 +643,17 @@ def get_data(self):
         except KeyError:
             self.signal_units = ''
 
+        ny, nx = self.data.shape
+        col_idx, row_idx = np.indices((ny, nx))
+        try:
+            wcs = WCS(header)
```
bhilbert4, line 666: If you switch out the astropy.wcs for the jwst datamodels wcs, it should enable you to display RA, Dec info for more types of data (e.g. spectroscopic data). I believe that the astropy.wcs is only looking at the fits standard WCS information, right? That would mean the reported RA, Dec values would have a slight inaccuracy compared to the values the datamodel would give. It may be that the inaccuracy is small enough we don't care though.

bhilbert4 (approved):
This is great! I like the improved speed when rescaling and the new layout of the hover information. The line plots look great. Pre-computing them all can be slow, but defaulting to not showing the plots should keep loading time down for those who don't care about the plots. 

One detail I'm on the fence about is forcing the pixels to stay square, just because it limits how you can zoom in on details a bit. I think in most cases it's nice. But just to show a case on a fairly extreme aperture: https://jwql-test.stsci.edu/NIRCam/jw02084001001_04103_00001-seg001_nrca1_rate/explore_image/
I found it a little difficult to zoom in on the sources I wanted in this case. But, there are very few NIRCam observations taken with this aperture. It might be that a little difficulty in looking at these data is ok in return for the convenience with all the other observations taken through apertures that are closer to square. And it might be that with a little practice, working with these extreme apertures is just as easy as with others.

I should also mention that my JS/html knowledge is limited enough that I'm not able to provide a very deep review. It all looks reasonable to me, and my functional testing on the test server shows that everything is behaving as expected though.

melanieclarke:
@bhilbert4 - Thanks for your comments!  I'll address the specific threads above.

For the square pixels, I completely agree that it makes it harder to zoom to a rectangle, and I was quite surprised bokeh didn't have any better solution.  I think it's necessary, though, to make sure images stay useful for analysis, especially for the subarray cases.

For comparison, here's a spectral image I look at a lot for testing.  Compare the [current version](https://jwql.stsci.edu/NIRSpec/jw02589006001_04101_00001-seg001_nrs1_rate/explore_image/), which does not enforce square pixels to the [test version](https://jwql-test.stsci.edu/NIRSpec/jw02589006001_04101_00001-seg001_nrs1_rate/explore_image/) that does.  Once the pixels have gone oblong, the image becomes very difficult to interpret, and it can't be recovered from.

For what it's worth, for small subarrays, zooming via mouse scrolling seems to work better than zooming to a rectangle.

jwql/utils/interactive_preview_image.py
melanieclarke: Generally I found that when the log scale image is good, it's very good, but when it's bad, it's bad in a way that's disturbing to look at.  For images with many negative values, for example, log scale doesn't make much sense.  I think the linear scale is a safer default.

jwql/website/apps/jwql/views.py
melanieclarke: Oops, sorry, I'll fix that.

jwql/utils/interactive_preview_image.py
melanieclarke: Good catch! It looks like I forgot to update DQ handling when I refactored the hover tool.  It should display the DQ description string, but currently is displaying NaN.  I'll get that fixed.

jwql/utils/interactive_preview_image.py
melanieclarke: It hadn't occurred to me to use the datamodels. I'll look into it.  Thanks!

bhilbert4:
I see what you mean about that oblong spectrum. It definitely makes sense to force square pixels in that case.

jwql/website/apps/jwql/views.py
melanieclarke: Fixed.

jwql/utils/interactive_preview_image.py
melanieclarke: This should be fixed now.

jwql/utils/interactive_preview_image.py
melanieclarke: I looked into this briefly, but there are some complexities in using the datamodels for different data products that I don't understand very well.  Maybe we can leave this for a future improvement?  There's a lot we could do with the datamodel metadata -- spectral WCS, for example -- but I think it should be a separate PR.  

I can pull out the RA/Dec info for now, if you prefer.

bhilbert4: Either way seems fine. I agree that down the line we could have it show wavelength as well as RA/Dec in spectroscopic cases.

jwql/utils/interactive_preview_image.py
bhilbert4: Looks good to me.

bhilbert4:
From my perspective, this is ready to merge. @mfixstsci

mfixstsci:
@bhilbert4 thank you for the review. I will go ahead and get it in!

PR #1290, opened 2023-07-14
Update tab_service search for all intents

BradleySappington:
Proposal list wasn't getting fully populated from TAPService request.  It was omitting proposals with `intent` that wasn't `science`.  We need all proposals to be viewable from JWQL so updating query using recommendation from @dr-rodriguez


mfixstsci (approved):
Changes look good, thanks for digging into this @BradleySappington! API is tested, merging and will wait for changes on production application.

PR #1278, opened 2023-06-01
[WIP] TA Monitor

mengesser:
Beginnings of TA monitor. Moved to new feature branch to avoid merge conflicts.

PR #1292, opened 2023-07-21
[WIP]: Add TA monitor for NIRCam, NIRISS, MIRI

bhilbert4:
Add the TA monitor, with requirements specified by our meeting with the instrument teams and coronagraphy working group. The code is still under heavy development. At the moment the general layout for the non-coronagraphy TA monitor is nearly there. The coronagraphic TA monitor work will copy and extend the non-coronagraphic functionality. 

Development on plot creation also has not started. Plots need to be made using straight Bokeh, rather than Bokeh templating, which has been effectively deprecated for a while.

This PR replaces #1278


bhilbert4:
Note that the edb code will need to be tweaked to look for rowCen and colCen, rather than rowCentroid and colCentroid, per the email from Mario and Neville. See JWSTOSS-8045 for details.

PR #1297, opened 2023-08-01
Bias monitor failure from recent shared tasks update

### LINKED ISSUE #1298: Bias monitor failing after shared tasks update
The bias monitor is failing after the shared_tasks update in #1262.

The log file shows error statements such as: 
Pipeline was unable to process jw01453018001_02201_00001_nrcb1_uncal_0thgroup.fits

The "0thgroup" substring being after the uncal string is not anticipated in the updates to shared_tasks. The updates assumed that the calibration suffic (uncal, cal, etc) would be the final substring in the filename.

bhilbert4:
The bias monitor is failing after the recent shared tasks updates in #1262 

One problem that popped up immediately is that the pipeline calls for the bias monitor appeared to be feeding the wrong short filename into the pipeline code. It turns out that the bias monitor extracts the 0th group from each input file and places it in a new file with the suffix _uncal_0thgroup.fits. Having the "0thgroup" string after uncal was screwing up the new line in shared tasks that was supposed to return the "short name" of the file (e.g. jw01068001001_02101_00001_nrcb2). This fixes that issue. 

Still checking to see if there are any other fixes that need to be made.

### DIFF
--- a/jwql/instrument_monitors/common_monitors/bias_monitor.py
+++ b/jwql/instrument_monitors/common_monitors/bias_monitor.py
@@ -432,7 +432,13 @@ class Bias():
             # Add this new entry to the bias database table
             with engine.begin() as connection:
                 connection.execute(self.stats_table.__table__.insert(), bias_db_entry)
-            logging.info('\tNew entry added to bias database table: {}'.format(bias_db_entry))
+
+            # Don't print long arrays of numbers to the log file
+            log_dict = {}
+            for key in bias_db_entry:
+                if key not in ['collapsed_rows', 'collapsed_columns', 'counts', 'bin_centers']:
+                    log_dict[key] = bias_db_entry[key]
+            logging.info('\tNew entry added to bias database table: {}'.format(log_dict))
 
             # Remove the raw and calibrated files to save memory space
             os.remove(filename)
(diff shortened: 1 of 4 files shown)


bhilbert4:
Ok, I think this is working. My limited test using one nircam aperture worked well. I'm currently running a full test across all instruments and apertures, to be sure I'm not missing anything that is instrument-dependent. So far that test is running well (on the dev server), so I'm optimistic that everything will be fine. In the meantime, @mfixstsci I think this is ready for review.

bhilbert4:
The full test on the dev server completed successfully, and the resulting bias monitor pages look good. Now officially ready for review. @mfixstsci

jwql/instrument_monitors/common_monitors/bias_monitor.py
```diff
@@ -432,7 +432,13 @@ def process(self, file_list):
             # Add this new entry to the bias database table
             with engine.begin() as connection:
                 connection.execute(self.stats_table.__table__.insert(), bias_db_entry)
-            logging.info('\tNew entry added to bias database table: {}'.format(bias_db_entry))
+
+            # Don't print long arrays of numbers to the log file
```
mfixstsci, line 436: Thank you!

mfixstsci (approved):
Thanks @bhilbert4 this looks great!

PR #1360, opened 2023-09-21
Supply web driver instance when saving png files

### LINKED ISSUE #1357: Dark monitor crashing due to reported missing env packages
_Issue [JWQL-138](https://jira.stsci.edu/browse/JWQL-138) was created on JIRA by [Bryan Hilbert](https://jira.stsci.edu/secure/ViewProfile.jspa?name=hilbert):_

On Sept 13, the dark monitor crashed while trying to create png files. The error statement reports:

"RuntimeError: Neither firefox and geckodriver nor a variant of chromium browser and chromedriver are available on system PATH. You can install the former with 'conda install -c conda-forge firefox geckodriver'."

However, I've confirmed that both firefox and geckodriver are present in the jwql-3.9 environment. The environment was just deleted and re-created from scratch last week. Could it be that the web service is working with an old version of the environment? Except that both packages were present in the old version of the environment as well.

bhilbert4:
Resolves #1357 

With selenium > 4.9.1 it looks like we need to supply a web driver instance to bokeh's export_png() when saving png files. This PR creates a small wrapper function around Bokeh's export_png() that creates the web driver and passes it to export_png.


bhilbert4:
Successful testing here as well. @mfixstsci this is ready for review.

jwql/tests/test_utils.py
```diff
@@ -479,6 +483,17 @@ def test_filesystem_path():
     assert check == location
 
 
+def test_save_png():
```
mfixstsci, line 486: Thanks for the test!

jwql/utils/utils.py
```diff
@@ -749,13 +749,34 @@ def read_png(filename):
...
         # with a lower-left origin
-        view[:,:,:] = np.flipud(np.asarray(rgba_img))
+        view[:, :, :] = np.flipud(np.asarray(rgba_img))
     else:
         view = None
     # Return the 2D version
     return img
 
 
+def save_png(fig, filename=''):
```
mfixstsci, line 759: Thanks, this looks great. Is this just patching function until we can get the bokeh version updated? Or are we looking at having to keep this around long term?

mfixstsci (approved):

jwql/utils/utils.py
bhilbert4: I haven't checked Bokeh > 3.0 to see, but my guess is that this will be a long term fix.

mfixstsci:
Thanks @bhilbert4

PR #1364, opened 2023-09-28
Include migrations in source control

BradleySappington:
Migrations need to be part of our source control, adding all migrations I have done up to this point.


bhilbert4 (approved):
It's perfect! Don't change a thing.

PR #1370, opened 2023-10-03
update deps in environment files

BradleySappington:
Update the yml and requirements files with new dependency checks


BradleySappington:
@bhilbert4 @mfixstsci  Ready for review

bhilbert4 (changes requested):
A couple general comments:
pyproject.toml references setuptools versions, but those versions are behind the versions in the updated env files. Looks like they've been behind for a while though, and I haven't seen any issues related to it.
rtd_requirements.txt should be brought in line with the updated env files
jwst released 1.12.3 a few days ago. So far the 1.12.x releases are all release candidates, and I did see at least one person who found an obvious bug in 1.12.2.
A new version of yaml??? They've been on v 6.0.0 forever...

BradleySappington:
> A couple general comments: pyproject.toml references setuptools versions, but those versions are behind the versions in the updated env files. Looks like they've been behind for a while though, and I haven't seen any issues related to it. 

@mfixstsci is there a reason setuptools is set to less than 65.5.0?

bhilbert4:
According to the handy Blame button, that setuptools line was copied straight from setup.py into pyproject.toml back in April. So I'm assuming those versions were appropriate in April, but probably not now?

bhilbert4 (approved):
Looks good to me! 
The RTD log indicates that sphinx_rtd_theme is installed as a dependency of stsci_rtd_theme.

Issue #1378, opened 2023-10-12
Identify databases containing full paths


stscijgbot-jwql:
_Issue [JWQL-148](https://jira.stsci.edu/browse/JWQL-148) was created on JIRA by [Bryan Hilbert](https://jira.stsci.edu/secure/ViewProfile.jspa?name=hilbert):_

Determine which JWQL databases currently contain full paths to output files.

stscijgbot-jwql:
_Comment by [Brian York](https://jira.stsci.edu/secure/ViewProfile.jspa?name=york) on [JIRA](https://jira.stsci.edu/browse/JWQL-148?focusedCommentId=669775#comment-669775):_

Based on my checks, we're looking at the following:
 * There are file names and/or paths in:
 ** <INSTRUMENT>BadPixelStats
 ** <INSTRUMENT>DarkDarkCurrent
 ** <INSTRUMENT>DarkPixelStats
 ** <INSTRUMENT>ReadnoiseStats

Of those, we have the following fields that contain direct file information:
 * BadPixelStats
 ** source_files: contains an array of bare filenames
 ** baseline_file: contains an absolute path to a location on central storage
 * DarkDarkCurrent
 ** source_files: contains an array of bare filenames
 ** baseline_file: contains a bare filename
 * DarkPixelStats
 ** source_files: contains an array of bare filenames
 ** mean_dark_image_file: bare filename
 ** baseline_file: bare filename
 * ReadnoiseStats
 ** uncal_filename: absolute path on /internal
 ** readnoise_filename: absolute path on /internal
 ** readnoise_diff_image: absolute path on /internal

My recommendation is, obviously, to change the ReadnoiseStats table to be paths relative to the configured location. The question is whether we want to make more changes for the sake of consistency. My general thought is that yes, things should be consistent, so my overall recommendation is:
 * Change the ReadnoiseStats entries to be only the file name, and generate the path at runtime in the same way that the other monitors do
 * Generally adopt the system that, if we have a file name string in the database, it's **only** the file name, not the path, and we can generate the path at runtime
 * In the **specific** case where we have a reference file name (currently only BadPixelStats.baseline_file), then providing the file's path on central storage is fine.

Issue #1377, opened 2023-10-12
Update code to prevent full paths in the databases


stscijgbot-jwql:
_Issue [JWQL-147](https://jira.stsci.edu/browse/JWQL-147) was created on JIRA by [Bryan Hilbert](https://jira.stsci.edu/secure/ViewProfile.jspa?name=hilbert):_

Make the code updates to prevent storing full paths in the JWQL databases

stscijgbot-jwql:
_Comment by [Bryan Hilbert](https://jira.stsci.edu/secure/ViewProfile.jspa?name=hilbert) on [JIRA](https://jira.stsci.edu/browse/JWQL-147?focusedCommentId=677752#comment-677752):_

The readnoise monitor code was been updated to prevent storing full paths in the database tables, via [#&#x2060;1394](https://github.com/spacetelescope/jwql/pull/1394).

stscijgbot-jwql:
_Comment by [Bryan Hilbert](https://jira.stsci.edu/secure/ViewProfile.jspa?name=hilbert) on [JIRA](https://jira.stsci.edu/browse/JWQL-147?focusedCommentId=677755#comment-677755):_

None of the other current monitors need any updating for this issue.

Issue #1376, opened 2023-10-12
Determine code changes needed to prevent full paths in database


stscijgbot-jwql:
_Issue [JWQL-146](https://jira.stsci.edu/browse/JWQL-146) was created on JIRA by [Bryan Hilbert](https://jira.stsci.edu/secure/ViewProfile.jspa?name=hilbert):_

We need to figure out which monitors (and any other code) need to be updated in order to stop storing full paths in JWQL database tables.

stscijgbot-jwql:
_Comment by [Bryan Hilbert](https://jira.stsci.edu/secure/ViewProfile.jspa?name=hilbert) on [JIRA](https://jira.stsci.edu/browse/JWQL-146?focusedCommentId=677754#comment-677754):_

Only the readnoise monitor code needed updating. Those changes were completed in [#&#x2060;1394](https://github.com/spacetelescope/jwql/pull/1394).

PR #1152, opened 2023-01-10
NIRCam Claw Monitor

bsunnquist:
This PR makes the NIRCam Claw Monitor, which creates observation-level stacks of all NIRCam imaging data to detect claws and other scattered light/persistence issues. These stacks can also double as a rough flat field check for those observations with longer exposure times/higher backgrounds.

PR #1394, opened 2023-11-06
Stop storing absolute paths in the database

york-stsci:
I believe the intent summary is on Jira, but the goal of this PR is to update the readnoise monitor so that it doesn't store absolute paths in the database, and it also doesn't expect absolute paths when run as part of the website.

There will also be an associated script to update the existing database entries that have absolute paths in them.


york-stsci:
[update_absolute_paths.py.txt](https://github.com/spacetelescope/jwql/files/13272135/update_absolute_paths.py.txt)

york-stsci:
The above is the database update script (named as a text file, because GitHub apparently can't handle the upload of a .py file)

york-stsci:
Closes #1377 
Closes #1378 
Closes #1376

jwql/website/apps/jwql/monitor_pages/monitor_readnoise_bokeh.py
```diff
@@ -107,13 +110,16 @@ class ReadNoisePlotTab():
...
         self.aperture = aperture
+        self.ins_ap = "{}_{}".format(self.instrument.lower(), self.aperture.lower())
 
         self.db = ReadnoiseMonitorData(self.instrument, self.aperture)
 
         self.plot_readnoise_amplifers()
         self.plot_readnoise_difference_image()
         self.plot_readnoise_histogram()
 
+        self.file_path = os.path.join(OUTPUT_DIR, "readnoise_monitor", "data", self.ins_ap)
```
mfixstsci: This  is great @york-stsci along with line 177, this makes sense for serving the static outputs in the app. I think we can use this as a model for moving forward with the other monitors to avoid the issues we were seeing.

mfixstsci (approved):
@york-stsci thanks for the updates! We will need to perform similar updates to #1152. Have the databases been updated yet?

york-stsci:
The script has not yet been run in active mode. I’m saving that until after we review everything.

BradleySappington (approved):

PR #1397, opened 2023-11-08
Moving monitor outputs used by the server to central storage

york-stsci:
This is the code change part of issue 1395. Before it is merged, all in-use config.json files should be updated to contain a "working" key, whose value should initially be set to the same value as the "outputs" key.


jwql/instrument_monitors/common_monitors/readnoise_monitor.py
```diff
@@ -253,7 +253,7 @@ def image_to_png(self, image, outname):
             The full path to the output png file.
         """
 
-        output_filename = os.path.join(self.data_dir, '{}.png'.format(outname))
+        output_filename = os.path.join(self.output_data_dir, '{}.png'.format(outname))
```
BradleySappington, line 256: change docstring of "data_dir"

jwql/instrument_monitors/common_monitors/dark_monitor.py
```diff
@@ -132,7 +132,7 @@ class Dark():
 
     Attributes
     ----------
-    output_dir : str
+    working_dir : str
```
BradleySappington: this doesn't line up with other uses of working dir.  
outputs where outputs are placed (now on central store), working where data is kept and gathered to do calculations.  Is that understanding correct?

jwql/instrument_monitors/common_monitors/dark_monitor.py
```diff
@@ -132,7 +132,7 @@ class Dark():
...
         Path into which outputs will be placed
 
     data_dir : str
```
BradleySappington: see above comment, but also, data_dir is renamed to `output_data_dir` in other modules. We should maintain naming conventions.

jwql/instrument_monitors/common_monitors/bad_pixel_monitor.py
```diff
@@ -1110,6 +1110,7 @@ def run(self):
         logging.info('Begin logging for bad_pixel_monitor')
 
         # Get the output directory
+        self.working_dir = os.path.join(get_config()['working'], 'bad_pixel_monitor')
```
BradleySappington, line 1113: typo bug - should be `working_data_dir`

BradleySappington (changes requested):
A few small changes.  Also please keep naming conventions common through all monitors (when appropriate).  It looks like `working_data_dir` and `output_data_dir` are the most used so would be easiest to stick with that.  I did not mark every instance where naming standard wasn't used.

jwql/instrument_monitors/common_monitors/readnoise_monitor.py
york-stsci: Change made.

jwql/instrument_monitors/common_monitors/dark_monitor.py
york-stsci: That understanding is correct, with one nuance (on which more below).

jwql/instrument_monitors/common_monitors/dark_monitor.py
york-stsci: Change made.

jwql/instrument_monitors/common_monitors/bad_pixel_monitor.py
york-stsci: This is the nuance – some of the common monitors create a "data" directory underneath the working/output directory, usually to separate files by amplifier/detector/whatever, and it's *that* directory that's the working_data_dir. The directory specific to the monitor, by itself, is the working_dir. So essentially, working_data_dir is built from working_dir by appending on /data/ and potentially a further specifier.

BradleySappington: My bad! I just opened up the whole file in a codespace, I see what you're doing now. Apologies for the false alarm!

york-stsci:
@BradleySappington I think I've made all the changes you requested, but I can't figure out how to tell the change request that.

BradleySappington (approved):

BradleySappington:
I approved, I think that should fix it aside from resolving conflicts and making sure the tests pass

PR #1403, opened 2023-11-27
use build workflow from OpenAstronomy

zacharyburnett:
adds a build workflow that also does publishing to PyPI (but only on a release)

image

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -1,0 +1,15 @@
+name: build
+
+on:
+  release:
+    types: [ released ]
+  pull_request:
+  workflow_dispatch:
+
+jobs:
+  build:
+    uses: OpenAstronomy/github-actions-workflows/.github/workflows/publish_pure_python.yml@v1
+    with:
+      upload_to_pypi: ${{ (github.event_name == 'release') && (github.event.action == 'released') }}
+    secrets:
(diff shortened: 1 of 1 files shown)


mfixstsci:
Hey @zacharyburnett thanks for the submission, our team is currently discussing how long JWQL will remain on pypi. The tool is open source (visible to the community through github) but is not intended to be an installable package for the public. I think the intent behind publishing to pypi was more of a formality than something of practical use in the past. I will make sure to keep you in the loop here on what we decide to do.

.github/workflows/build.yml
```diff
@@ -0,0 +1,15 @@
...
+  pull_request:
+  workflow_dispatch:
+
+jobs:
+  build:
+    uses: OpenAstronomy/github-actions-workflows/.github/workflows/publish_pure_python.yml@v1
+    with:
+      upload_to_pypi: ${{ (github.event_name == 'release') && (github.event.action == 'released') }}
+    secrets:
+      pypi_token: ${{ secrets.PYPI_PASSWORD_STSCI_MAINTAINER }}
```
zacharyburnett, line 15: @mfixstsci ok! then if you do decide to no longer publish to PyPI, you can also use this workflow just as a build check (if you think that's useful):
```suggestion
    with:
      upload_to_pypi: false
```

zacharyburnett:
superseded by #1568

PR #1425, opened 2024-01-10
Hotfix: Crashing archive_database_update

BradleySappington:
Archive Database Update is crashing with a key error when past does not return all values from a `get_proposals_by_category` call in data_containers.py

Add simple try except and logging


BradleySappington:
@mfixstsci @bhilbert4 - Ready for review, verified all works on test

mfixstsci (approved):
Looks good to me @BradleySappington thanks!

PR #1438, opened 2024-01-22
remove special case anomalies and make all caps

BradleySappington:
Anomaly form was not being updated properly due to special case naming. 
Making all uppercase to match query page, keep code clean, and avoid special scenario code for small number of anomalies.


mfixstsci (approved):
Looks good @BradleySappington this is a tricky one to catch. Thanks for getting this in so quickly.

PR #1442, opened 2024-01-24
bug fix for values_list

BradleySappington:
per django documentation,
values_list should be able to take a single column name ("root_name") with flat=True parameter.  This isn't taking in our current situation and is throwing an error sayin `'' is not an acceptable field`  Adding a bogus but existing second column of data resolves the issue.  Pushing forward to resolve the bug with this release but can re-evaluate issue with django if priorities permit.


BradleySappington:
https://jira.stsci.edu/browse/JWQL-166

bhilbert4 (approved):
Looks good to me. Should we create a new JIRA issue to fix this, and include links to the relevant DJango doc or stackoverflow question?

BradleySappington:
> Looks good to me. Should we create a new JIRA issue to fix this, and include links to the relevant DJango doc or stackoverflow question?

yeah, I think thats a good idea.  We are soon going to update to django 5.0.  in the next release, so the work should be done after that.  It'll be low priority since it at least works this way.

PR #1446, opened 2024-01-24
Update Path for Claws and Background

mfixstsci:
Missed the static_dir portion. Part of #1397


BradleySappington (approved):
Are there any other potential locations that may have been missed?

PR #1455, opened 2024-01-29
Retrieve program number and category efficiently

### LINKED ISSUE #1428: Better way to get a list of instrument proposals and categories
_Issue [JWQL-162](https://jira.stsci.edu/browse/JWQL-162) was created on JIRA by [Bryan Hilbert](https://jira.stsci.edu/secure/ViewProfile.jspa?name=hilbert):_

In data_containers.get_proposals_by_category(), the code retrieves a list of proposals and associated categories for a given instrument. It does this in a very inefficient way, by querying MAST for all exposures for a given instrument, and then picking out the unique proposal names. But for NIRCam, we are currently bumping up against our (already bumped up) MAST query limit of 500,000 results. (NIRCam currently returns 540,000 results.) Missing results are leading to missing proposals in the list, which causes problems downstream. Related to JWQL-160, we have bumped up the MAST query limit in order to solve the problem in the short term. But in the long term, we need a better way to get a list of the proposals and categories for a given instrument. 

I have asked Dick Shaw, as part of [Help Desk ticket INC0196521](https://stsci.service-now.com/stars?id=ticket&table=incident&sys_id=3448820947ef79d0289c4f32736d4309) if there is a better way to get this information.

bhilbert4:
Resolves #1428 

This PR fixes a problem where our MAST query to retrieve program numbers + categories was really inefficient, and was running into the maximum number of rows that MAST would return (even with our increased limit). 

With this, get_proposals_by_category() now gets information by querying pyvo, rather than MAST itself. I have confirmed that the dictionary returned with the new function matches that from the old function (and the new function executes many times faster than the old function).

### DIFF
--- a/jwql/website/apps/jwql/data_containers.py
+++ b/jwql/website/apps/jwql/data_containers.py
@@ -1391,7 +1391,7 @@ def get_preview_images_by_rootname(rootname):
 
 
 def get_proposals_by_category(instrument):
-    """Return a dictionary of program numbers based on category type
+    """Return a dictionary of program numbers and category type
     Parameters
     ----------
     instrument : str
--- a/jwql/website/apps/jwql/data_containers.py
+++ b/jwql/website/apps/jwql/data_containers.py
@@ -1400,24 +1400,16 @@ def get_proposals_by_category(instrument):
     Returns
     -------
     category_sorted_dict : dict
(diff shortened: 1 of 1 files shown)


mfixstsci (approved):
Not only is the code more elegant than what I submitted, it will be more consistent. Thanks for the changes @bhilbert4

PR #1456, opened 2024-01-29
Add script to automatically delete old log files

### LINKED ISSUE #1434: Create script to delete old log files
_Issue [JWQL-165](https://jira.stsci.edu/browse/JWQL-165) was created on JIRA by [Bryan Hilbert](https://jira.stsci.edu/secure/ViewProfile.jspa?name=hilbert):_

With instrument monitors and other scripts running via cron job, JWQL builds up a lot of log files very quickly. We should create a script that will periodically clean out old log files, in order to save disk space and to limit the number of files present in the log file directories.

It probably makes sense to have the script only examine and delete files that were created by the servers and ignore those created by other machines, since the latter will be coming from development work and older logs may be useful.

bhilbert4:
Resolves #1434 

This PR introduces a script that will delete all log files older than a given threshold age. We can run this via cron job, which would then prevent the build-up of tons of log files, especially in the archive_database_update directory.

@mfixstsci @BradleySappington I think this is ready for review. (The line that does the actual deletion is commented out at the moment in favor of a print statement saying what it would delete.)

I tested on my local machine, and the log files it said it would delete were correct.

### DIFF
--- a/jwql/website/apps/jwql/clean_old_log_files.py
+++ b/jwql/website/apps/jwql/clean_old_log_files.py
@@ -74,10 +74,8 @@ def run(time_limit=timedelta(days=14)):
                     last_modified_time = datetime.fromtimestamp(stat_result.st_mtime)
                     age = now - last_modified_time
                     if age > time_limit:
-                        #os.remove(item.name)
                         full_path = os.path.join(log_dir, logtype, item)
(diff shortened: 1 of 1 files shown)


mfixstsci (approved):
Looks good, tested on `dev` and everything looks to be in order.

PR #1479, opened 2024-02-08
Get subarray lists from datamodels schema

### LINKED ISSUE #1389: Add missing subarrrays to constants
While searching for NIRCam data, I noticed that there are a number of subarrays that are missing from the query form. These values come from utils/constants.py, in the SUBARRAYS_PER_DETECTOR entry.

The full lists of possible subarrays for all the instruments are given in the schema for the stdatamodels:  https://github.com/spacetelescope/stdatamodels/blob/master/src/stdatamodels/jwst/datamodels/schemas/subarray.schema.yaml

We should update our lists in constants.py to match these lists.

bhilbert4:
Resolves #1389 

Rather than relying on manually provided subarray lists (which had been missing some entries), get subarray lists from the subarray schema file in the stdatamodels package. This ensures that all available subarrays will be present on the JWQL query page.

One downside to this is the way the schema file is read in. Since the instrument names are in commented lines in the schema file, the subarray lists are accessed through hard-coded indexing. If the order of the instruments' lists ever change in the schema file, we would need to update this code. That seems pretty unlikely though.

Local testing looks good.

### DIFF
--- a/jwql/utils/constants.py
+++ b/jwql/utils/constants.py
@@ -921,7 +921,7 @@ SUBARRAYS_PER_INSTRUMENT = {
     "nirspec": ['FULL'] + sorted(schema["properties"]["meta"]["properties"]["subarray"]["properties"]["name"]["anyOf"][6]['enum']),
     "miri": ['FULL'] + sorted(schema["properties"]["meta"]["properties"]["subarray"]["properties"]["name"]["anyOf"][1]['enum']),
     "fgs": ['FULL'] + sorted(schema["properties"]["meta"]["properties"]["subarray"]["properties"]["name"]["anyOf"][0]['enum'])
-                           }
+    }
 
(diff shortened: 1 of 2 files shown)


bhilbert4:
Ready for review.

mfixstsci (approved):
Thanks @bhilbert4 let's hope the schema doesn't change 🤞

mfixstsci:
Just checked locally, subarrays are appearing for FGS and NIRSPEC:
Screen Shot 2024-02-08 at 3 23 37 PM
Screen Shot 2024-02-08 at 3 23 55 PM

@bhilbert4 If these look correct, I will merge.

bhilbert4:
Those look right to me! 
Here's the schema file: https://github.com/spacetelescope/stdatamodels/blob/main/src/stdatamodels/jwst/datamodels/schemas/subarray.schema.yaml

PR #1480, opened 2024-02-08
set up Dependabot to keep CI actions updated

zacharyburnett:


mfixstsci:
Very late to this.. Sorry @zacharyburnett does this have any overlap with #1646?

zacharyburnett:
> Very late to this.. Sorry @zacharyburnett does this have any overlap with #1646?

No worries! yeah, this should be superseded by #1646

PR #1481, opened 2024-02-12
Initial migration to django management

york-stsci:
This PR includes the following:

- Ran `python manage.py inspectdb --database monitors` to generate the `monitor_models.py` file
- Edited the file to remove the `managed=False` lines, and re-order the databases
- Ran `python manage.py migrate --database monitors` to generate an initial migration file

In addition, I adjusted the config.json file to point django to both its own default databases and the jwqldb database, but have not yet removed any of the other SQL code.

If I understand correctly what's going on here, it should now be possible to access and update the monitor databases either via Django or via SQLAlchemy (where by "update" I mean "add or edit rows" not "change the composition of tables").


BradleySappington:
@york-stsci - verifying that you ran `python manage.py makemigrations` to actually create the migration file (and if so, you will need to include that migration file in this PR).  If you are not including the initial migration, then you can remove that from the PR summary.

BradleySappington (changes requested):
You will need to include the initial migration file (does that go in anew folder?) 

Please also include the updated example config file that has the new monitors database setup information in it.

jwql/website/jwql_proj/settings.py
```diff
@@ -95,20 +95,21 @@
...
+}
 
 WSGI_APPLICATION = 'jwql.website.jwql_proj.wsgi.application'
 
 
 # Database
 # https://docs.djangoproject.com/en/2.0/ref/settings/#databases
 DATABASES = {
-    'default': get_config()['django_database']
+    'default': get_config()['django_databases']['default'],
```
mfixstsci, line 111: in my local config the key is `django_database` without the `s`. Did we update the key at some point? Also are `monitors` and `default` new keys we need to add?

jwql/website/manage.py
```diff
@@ -47,11 +47,12 @@
         'outputs': 'outputs',
         'preview_image_filesystem': 'preview_images',
         'thumbnail_filesystem': 'thumbnails'
-        }
+    }
 
     for directory in ['filesystem', 'outputs', 'preview_image_filesystem', 'thumbnail_filesystem']:
         symlink_location = os.path.join(os.path.dirname(__file__), 'apps', 'jwql', 'static', directory_mapping[directory])
         if not os.path.exists(symlink_location):
+            print("{} does not exist".format(symlink_location))
```
mfixstsci: This is fine, it looks like it still attempts to create a symlink. If the `symlink_location` doesn't exist, this print statement will alert the user, but does `os.symlink` not throw an error?

mfixstsci (approved):
Everything looks okay from a coding perspective. Just a couple of questions from me.

jwql/website/jwql_proj/settings.py
york-stsci: So, I changed it from `django_database` to `django_databases` because there are now 2 django databases – the existing one (both previously and now called "default"), and the JWQLDB one, called "monitors".

jwql/website/manage.py
york-stsci: So, this was a modification that I made to the auto-generated file to find out what it was failing on, and why, because I couldn't figure out why my local JWQL (with config file) couldn't run the django manage script, and it turns out that if you create a symlink, but it's an incorrect symlink (as in it points to something that doesn't exist), then os.path.exists() will return False on the symlink, even though the symlink is there in the filesystem, and this helped me diagnose what the issue actually was. I didn't want to actually change the way the file behaved, because again it's auto-generated by django so I wasn't comfortable fiddling with it too much.

Basically, the issue that I had was that:
- If you have directories (e.g.) internal/ and actual/
- If you create actual/data/
- If you create the symlink internal/data/ but you point it at actual/Data

Then calling os.path.exists() on internal/data will return False, but trying to create the symlink internal/data pointing to actual/data will fail because internal/data already exists. I'm not actually sure why the script wasn't working (why it didn't create things correctly the first time, that is) but I was able to solve it by making sure all the symlinks were set up and valid before running, and then I forgot to remove that diagnostic line.

york-stsci:
Enclosed is the updated config file.
[config.json](https://github.com/spacetelescope/jwql/files/14243058/config.json)

york-stsci:
@BradleySappington apparently when you're adding a new database to an existing installation, it uses the existing file for makemigrations, and migrate.py doesn't (yet) *create* any new migrations for the new database, so there isn't a new migration file to include.

This is a slight weirdness with django. More or less every django manage.py command takes the argument `--database=BLAH` and, if you don't include that argument, then it will operate on the default database (only), but `makemigrations` operates on *all* databases, and can't be restricted to a single database, but will only check for past inconsistencies and errors in the default database *unless* you create a router class for the new tables that returns a valid value for its `allow_migrate()` function. As such, I'm going to create such a file, and also split out the existing monitor_models.py file by monitor, and at that point I'll make another push to the PR (hopefully quite soon).

BradleySappington:
@york-stsci , just want to be clear.  `python manage.py migrate` will never create any new migration files.  That command is the actual act of taking the already made migration files (from makemigrations) and applying them to the appropriate database.

BradleySappington:
@york-stsci , just want to be clear.  `python manage.py migrate` will never create any new migration files.  That command is the actual act of taking the already made migration files (from makemigrations) and applying them to the appropriate database.

jwql/website/manage.py
mfixstsci: Cool, thanks for the explanation. Let's just make sure to remove this line before merging, thank you!

mfixstsci (approved):

mfixstsci:
@york-stsci all of the model files look good. I spoke with @BradleySappington and we can make sure we dissect all of them when we add the monitors one at a time to make sure everything looks good. I approved it, tests are passing, at this point what's next?

york-stsci:
I think the only thing remaining is that @BradleySappington requested a code change, which was updating the example config, which I've now done, but I think he has to approve that I've done it before merging is unlocked.

jwql/example_config.json
```diff
@@ -2,21 +2,31 @@
     "admin_account" : "",
     "auth_mast" : "",
     "connection_string" : "",
-    "database" : {
+    "databases" : {
```
BradleySappington, line 5: Why is database[s] needed at all anymore? Isn't everything now nested under "django_databases"?

BradleySappington (commented):
Looks good with this one question

jwql/example_config.json
york-stsci: The main reason is that all of the monitors are still using the existing methods of accessing the databases, and there’s no reason to remove that interface until the clean up when we also remove the sqlalchemy code

BradleySappington: perfect, thanks!

BradleySappington (approved):

PR #1482, opened 2024-02-14
Update readnoise monitor to use django database models

york-stsci:
This PR is a test of updating one of the monitors (readnoise) to use the django database model, both for insertion and retrieval. It also adds general documentation describing the process.

### DIFF
--- a/docs/source/database.rst
+++ b/docs/source/database.rst
@@ -2,14 +2,137 @@
 database
 ********
 
-database_interface.py
----------------------
-.. automodule:: jwql.database.database_interface
-    :members:
-    :undoc-members:
+Introduction
+------------
 
-reset_database.py
------------------
-.. automodule:: jwql.database.reset_database
-    :members:
-    :undoc-members:
+JWQL uses the django database models for creating tables, updating table fields, adding 
+new data to tables, and retrieving data from tables. For instrument monitors, in particular,
+there are a number of issues that may be relevant.
+
+In general, django model documentation can be found 
+`on the django website <https://docs.djangoproject.com/en/5.0/#the-model-layer>`_. 
+Unfortunately, finding a particular bit of documentation in django can be a challenge, so
+a few quick-reference notes are provided below.
+
+Retrieving Data
+---------------
+
+Django retrieves data directly from its model tables. So, for example, if you want to 
+select data from the `MIRIMyMonitorStats` table, you must first import the relevant 
+object:
+
+.. code-block:: python
+
+    from jwql.website.apps.jwql.monitor_models.my_monitor import MIRIMyMonitorStats
+
+Then, you would access the database contents via the `objects` member of the class. For 
+example, to search the `MIRIMyMonitorStats` table for all entries matching a given 
+aperture, and to sort them with the most recent date at the top, you might do a query like
+the following:
+
+.. code-block:: python
+
+    aperture = "my_miri_aperture"
+    
+    records = MIRIMyMonitorStats.objects.filter(aperture__iexact=aperture).order_by("-mjd_end").all()
+
+In the above code,
+
+* The `filter()` function selects matching records from the full table. You can use 
+  multiple filter statements, or a single filter function with multiple filters. `filter()`
+  statements are always combined with an implicit AND.
+* If you have a long filter statement and want to separate it from the query statement,
+  you can create a dictionary and add it in with the `**` prepended. The dictionary 
+  equivalent to the above would be `{'aperture__iexact': aperture}`
+* The text before the double underscore is a field name, and the text afterwards describes
+  the type of comparison. `iexact` indicates "case-insensitive exact match". You can also
+  use a variety of standard SQL comparisons (`like`, `startswith`, `gte`, etc.)
+* If you want to get only records that *don't* match a pattern, then you can use the 
+  `exclude()` function, which otherwise operates exactly the same as `filter()`.
+* In the `order_by()` function, the `-` at the start is used to reverse the sort order, 
+  and the `mjd_end` is the name of the field to be sorted by.
+* The `all()` statement indicates that you want all the values returned. `get()` returns
+  a single value and can be iterated on, `first()` returns only the first value, etc.
+
+As an example of multiple filters, the code below:
+
+.. code-block:: python
+
+    records = MIRIMyMonitorStats.objects.filter(aperture__iexact=ap, mjd_end__gte=60000)
+    
+    filters = {
+        "aperture__iexact": ap,
+        "mjd_end__gte": 60000
+    }
+    records = MIRIMyMonitorStats.objects.filter(**filters)
+
+show two different ways of combining a search for a particular aperture *and* only data 
+taken more recently than MJD=60000.
+
+Note that django executes queries lazily, meaning that it will only actually *do* the 
+query when it needs the results. The above statement, for example, will not actually 
+run the query. Instead, it will be run when you operate on it, such as
+
+* Getting the length of the result with e.g. `len(records)`
+* Printing out any of the results
+* Asking for the value of one of the fields (e.g. `records[3].aperture`)
+
+Q Objects
+=========
+
+In order to make more complex queries, Django supplies "Q Objects", which are essentially 
+encapsulated filters which can be combined using logical operators. For more on this, see
+`the django Q object documentation <https://docs.djangoproject.com/en/5.0/topics/db/queries/#complex-lookups-with-q-objects>`_.
+
+Storing Data
+------------
+
+Django also uses the model tables (and objects) directly for storing new data. For example,
+if you have a monitor table defined as below:
+
+.. code-block:: python
+
+    from django.db import models
+    from django.contrib.postgres.fields import ArrayField
+
+    class NIRISSMyMonitorStats(models.Model):
+        aperture = models.CharField(blank=True, null=True)
+        mean = models.FloatField(blank=True, null=True)
+        median = models.FloatField(blank=True, null=True)
+        stddev = models.FloatField(blank=True, null=True)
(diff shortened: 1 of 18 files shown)


jwql/instrument_monitors/common_monitors/readnoise_monitor.py
```diff
@@ -47,14 +47,16 @@
...
-from jwql.database.database_interface import FGSReadnoiseQueryHistory, FGSReadnoiseStats  # noqa: E348 (comparison to true)
-from jwql.database.database_interface import MIRIReadnoiseQueryHistory, MIRIReadnoiseStats  # noqa: E348 (comparison to true)
-from jwql.database.database_interface import NIRCamReadnoiseQueryHistory, NIRCamReadnoiseStats  # noqa: E348 (comparison to true)
-from jwql.database.database_interface import NIRISSReadnoiseQueryHistory, NIRISSReadnoiseStats  # noqa: E348 (comparison to true)
-from jwql.database.database_interface import NIRSpecReadnoiseQueryHistory, NIRSpecReadnoiseStats  # noqa: E348 (comparison to true)
-from jwql.database.database_interface import session, engine  # noqa: E348 (comparison to true)
+
+# Need to set up django apps before we can access the models
+from django import setup
+os.environ.setdefault("DJANGO_SETTINGS_MODULE", "jwql.website.jwql_proj.settings")
```
bhilbert4: I think these lines need to be embedded in an if clause, so that they are not run when readthedocs or github actions are used? See data_containers.py for an example.

bhilbert4 (commented):
This looks ok to me, as someone who is not a Django model expert. And I'm assuming there's no issue with defining the models for all of the monitors in this PR, even though the monitors themselves are not yet switching over to use them.

I do think the Django environment variable setting and call to setup() probably need to be in an if clause so that they're not run on RTD or GA.

york-stsci:
The models were actually all defined in a previous (and merged) PR. This PR touches all of those files because, when I found out that I needed to change the (incorrect) data types that django had assigned to array fields, I just did it for all the monitors rather than stop with readnoise.

docs/source/database.rst
```diff
@@ -2,14 +2,137 @@
...
+    
+    entry = NIRISSMyMonitorStats(**values)
+    entry.save()
+
+There are (as usual) a few things to note above:
+
+* Django doesn't have a built-in array data type, so you need to import it from the 
+  database-compatibility layers. The ArrayField takes, as a required argument, the type
+  of data that makes up the array.
+* In the Meta sub-class of the monitor class, the `all_label = 'monitors'` statement is
```
BradleySappington:
```suggestion
* In the Meta sub-class of the monitor class, the `app_label = 'monitors'` statement is 
```

BradleySappington (commented):
This looks good to me with only a mispelling in the doc.
Aside from readnoise.py, I did not dive deep into the monitor_models as they will be changing.

Especially excellent work with the documentation update, thanks for taking the time to do that!

docs/source/database.rst
york-stsci: Change made. I realized that the right place to put this documentation was a different part of readthedocs, which is why it's marked as outdated.

jwql/instrument_monitors/common_monitors/readnoise_monitor.py
york-stsci: Done

jwql/instrument_monitors/common_monitors/readnoise_monitor.py
```diff
@@ -385,17 +388,18 @@ def most_recent_search(self):
...
             where the readnoise monitor was run.
         """
-
-        query = session.query(self.query_table).filter(and_(self.query_table.aperture == self.aperture,
-                self.query_table.run_monitor == True)).order_by(self.query_table.end_time_mjd).all()  # noqa: E712 (comparison to True)
+        filter_kwargs = {
+            'aperture__iexact': self.aperture,
+            'run_monitor__exact': True
+        }
+        query = self.query_table.objects.filter(**filter_kwargs).order_by("-end_time_mjd").all()
```
mfixstsci, line 395: I like how these queries are structured and like that we define the payload separately as well.

jwql/instrument_monitors/common_monitors/readnoise_monitor.py
```diff
@@ -385,17 +388,18 @@ def most_recent_search(self):
...
 
         if len(query) == 0:
             query_result = 59607.0  # a.k.a. Jan 28, 2022 == First JWST images (MIRI)
             logging.info(('\tNo query history for {}. Beginning search date will be set to {}.'.format(self.aperture, query_result)))
         else:
-            query_result = query[-1].end_time_mjd
+            query_result = query[0].end_time_mjd
```
mfixstsci, line 401: So here we are trying to get the last time the monitor was run, right? Instead of ordering chronologically by time and grabbing the last item in the list using `-1` indexing we sorting in reverse chronological order and grab the first item in the list?

jwql/instrument_monitors/common_monitors/readnoise_monitor.py
```diff
@@ -512,24 +516,24 @@ def process(self, file_list):
                                   'readnoise_filename': os.path.basename(readnoise_outfile),
                                   'full_image_mean': float(full_image_mean),
                                   'full_image_stddev': float(full_image_stddev),
-                                  'full_image_n': full_image_n.astype(float),
-                                  'full_image_bin_centers': full_image_bin_centers.astype(float),
+                                  'full_image_n': list(full_image_n.astype(float)),
```
mfixstsci, line 519: Do you mind reminding me why we need to type cast these objects into lists? Was this something that sqlalchemy did for us that the django models do not?

jwql/tests/test_bad_pixel_monitor.py
```diff
@@ -163,6 +163,7 @@ def test_get_possible_apertures(instrument, expected_list):
     assert ap_list == expected_list
 
 
+@pytest.mark.skipif(ON_GITHUB_ACTIONS, reason='Requires access to central storage.')
```
mfixstsci, line 166: Curious at what happened here, does using the django models not allow us to identify what tables we need to use? How did we do it before?

jwql/website/apps/jwql/monitor_models/bad_pixel.py
```diff
@@ -210,11 +211,11 @@ class Meta:
         app_label = 'monitors'
 
 
-class NirspecBadPixelStats(models.Model):
+class NIRSpecBadPixelStats(models.Model):
     entry_date = models.DateTimeField(unique=True)
     detector = models.CharField(blank=True, null=True)
-    x_coord = models.TextField(blank=True, null=True)  # This field type is a guess.
-    y_coord = models.TextField(blank=True, null=True)  # This field type is a guess.
+    x_coord = ArrayField(models.IntegerField())
```
mfixstsci, line 217: Much better than guessing, thank you for updating this.

mfixstsci (approved):
Hey @york-stsci just a few questions but this is huge improvement, thank you for submitting this!

jwql/instrument_monitors/common_monitors/readnoise_monitor.py
york-stsci: Correct. I just needed to figure out how to reverse the sort order, which makes sense but was not the obvious way to do it until I saw an example.

jwql/instrument_monitors/common_monitors/readnoise_monitor.py
york-stsci: Exactly. Before I did this, django was throwing `django.db.utils.ProgrammingError: can't adapt type 'numpy.ndarray'`, which seems like a direct translation of an error from postgresql, `psycopg2.ProgrammingError: can't adapt type 'numpy.ndarray'`. My guess is that SQLAlchemy had a basic adaptation where it would automatically turn something into a list if it wasn't a known type but it had a built-in way to do that (I'm pretty sure that in python you can have an object property that defines whether it's list-like, so my guess is that django just doesn't translate object types as they go to the database.

jwql/tests/test_bad_pixel_monitor.py
york-stsci: So basically the issue is that, if we want to import the django models (anything under jwql.website.apps.jwql.monitor_models), django will try to run setup in order to configure itself, and it can't do that without a config.json file. So we haven't changed the way that we select what objects to use, it's just that the sqlalchemy model doesn't need django to be set up in order to import it, but the django model does, so the identify tables function, which chooses which table objects to use based on what instrument is being looked at, can't run because the tables themselves haven't been imported.

jwql/website/apps/jwql/monitor_models/bad_pixel.py
york-stsci: Not to mention that the guess doesn't work (by which I mean if PostgreSQL has a field that's an array of integers, and you give it a string '[1, 2]', it doesn't know what to do with it, but it's pretty sure it's not an array of integers, so it will refuse to do the insert.

jwql/instrument_monitors/common_monitors/readnoise_monitor.py
mfixstsci: Cool, I figured there was a good reason thank you!

jwql/tests/test_bad_pixel_monitor.py
mfixstsci: Ah, using the config file 😎

jwql/website/apps/jwql/monitor_models/bad_pixel.py
mfixstsci: Yeah, that's also a very valid point 😅 thank you!

PR #1490, opened 2024-02-19
Moved ON_GITHUB_ACTIONS and ON_READTHEDOCS to constants.py

york-stsci:
Added import statements as needed.


jwql/edb/engineering_database.py
```diff
@@ -64,14 +64,13 @@
...
+from jwql.utils.constants import ON_GITHUB_ACTIONS
 from jwql.utils.credentials import get_mast_base_url, get_mast_token
 from jwql.utils.utils import get_config
 
 MAST_EDB_MNEMONIC_SERVICE = 'Mast.JwstEdb.Mnemonics'
 MAST_EDB_DICTIONARY_SERVICE = 'Mast.JwstEdb.Dictionary'
 
-# Temporary until JWST operations: switch to test string for MAST request URL
-ON_GITHUB_ACTIONS = '/home/runner' in os.path.expanduser('~') or '/Users/runner' in os.path.expanduser('~')
 if not ON_GITHUB_ACTIONS:
```
bhilbert4, line 75: I'm surprised this hasn't been failing on RTD. Maybe RTD isn't looking here for some reason.

jwql/tests/test_preview_image.py
```diff
@@ -30,11 +30,9 @@
 from jwst.datamodels import dqflags
 
 from jwql.utils.preview_image import PreviewImage, crop_to_subarray
+from jwql.utils.constants import ON_GITHUB_ACTIONS, ON_READTHEDOCS
 from jwql.utils.utils import get_config, ensure_dir_exists
 
-# Determine if tests are being run on Github Actions
-ON_GITHUB_ACTIONS = '/home/runner' in os.path.expanduser('~') or '/Users/runner' in os.path.expanduser('~')
-
 # Determine if the code is being run as part of a Readthedocs build
```
bhilbert4: Looks like this block should be removed as well.

jwql/utils/constants.py
```diff
@@ -963,3 +964,11 @@ class QueryConfigKeys:
     "nirspec": "https://jwst-docs.stsci.edu/jwst-near-infrared-spectrograph",
     "nircam": "https://jwst-docs.stsci.edu/jwst-near-infrared-camera",
 }
+
```
bhilbert4: Generally we've been trying to keep the constants in alphabetical order. It looks like we're not at 100% for this, but it could help to move these definitions up in the file.

bhilbert4 (commented):
It looks like there are a few files where ON_READTHEDOCS is not used, where these changes now import it. I can't claim to understand why RTD is not failing in these cases (maybe RTD does not look at unit test files?), but if we don't need ON_READTHEDOCS, then we shouldn't import it.

jwql/edb/engineering_database.py
york-stsci: Yeah, not sure. It looks as if it *should* be, but possibly there's been a change where ON_GITHUB_ACTIONS ends up resolving to True when the readthedocs build is running, at least in cases where there isn't a config.

jwql/tests/test_preview_image.py
york-stsci: Yes. Oops. Done.

jwql/utils/constants.py
york-stsci: Fixed.

mfixstsci (approved):
Thank you @york-stsci easy to review, hope it wasn't too hard to find all of the instances of `ON_GITHUB_ACTIONS`.

PR #1491, opened 2024-02-19
Fix Broken Dashboard - ReImplement Anomaly Dash

### LINKED ISSUE #1457: Update dashboard to look at new anomaly db tables
_Issue [JWQL-168](https://jira.stsci.edu/browse/JWQL-168) was created on JIRA by [Bryan Hilbert](https://jira.stsci.edu/secure/ViewProfile.jspa?name=hilbert):_

The dashboard produces a plot showing the number of each type of anomalies in the database for each instrument. Since we switched the anomaly database into the Django models (djangomodeldb), the dashboard is now looking at the wrong location for the anomaly info. 

We should point the dashboard to the correct location. The jwqldb page also looks in the wrong place when asked to show an <instrument>_anomaly table, so we should make the same update there.

BradleySappington:
Fix broken dashboard due to references to unused anomaly tables
Re write anomaly dashboard functionality to use new table structure.

### DIFF
--- a/jwql/website/apps/jwql/bokeh_containers.py
+++ b/jwql/website/apps/jwql/bokeh_containers.py
@@ -24,7 +24,7 @@ import os
 
 from bokeh.embed import components
 from bokeh.layouts import layout
-from bokeh.models import TabPanel, Tabs
+from bokeh.models import TabPanel, Tabs, DatetimeTickFormatter
 from bokeh.plotting import figure, output_file
 import numpy as np
 import pysiaf
@@ -282,10 +282,10 @@ def generic_telemetry_plot(times, values, name, nominal_value=None, yellow_limit
     if nominal_value is not None:
         fig.line(times, np.repeat(nominal_value, len(times)), line_dash='dashed')
 
-    fig.xaxis.formatter = DatetimeTickFormatter(hours=["%d %b %H:%M"],
-                                                days=["%d %b %H:%M"],
-                                                months=["%d %b %Y %H:%M"],
-                                                years=["%d %b %Y"],
+    fig.xaxis.formatter = DatetimeTickFormatter(hours="%d %b %H:%M",
+                                                days="%d %b %H:%M",
+                                                months="%d %b %Y %H:%M",
+                                                years="%d %b %Y"
                                                 )
     fig.xaxis.major_label_orientation = np.pi / 4
 
--- a/jwql/website/apps/jwql/bokeh_containers.py
+++ b/jwql/website/apps/jwql/bokeh_containers.py
@@ -401,7 +401,7 @@ def standard_monitor_plot_layout(instrument, plots):
     elif instrument.lower() == 'niriss':
         full_frame_lists = [
             [plots['NIS_CEN']]
-            ]
+        ]
     elif instrument.lower() == 'miri':
         full_frame_lists = [
             [plots['MIRIM_FULL']]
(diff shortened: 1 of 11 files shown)


york-stsci:
Is it worth changing any of the database queries that use database_interface to use the django models instead?

jwql/website/apps/jwql/bokeh_containers.py
```diff
@@ -282,10 +282,10 @@ def generic_telemetry_plot(times, values, name, nominal_value=None, yellow_limit
     if nominal_value is not None:
         fig.line(times, np.repeat(nominal_value, len(times)), line_dash='dashed')
 
-    fig.xaxis.formatter = DatetimeTickFormatter(hours=["%d %b %H:%M"],
-                                                days=["%d %b %H:%M"],
-                                                months=["%d %b %Y %H:%M"],
-                                                years=["%d %b %Y"],
+    fig.xaxis.formatter = DatetimeTickFormatter(hours="%d %b %H:%M",
```
bhilbert4, line 286: How did this function ever work if DateTimeTickFormatter was never imported?

jwql/website/apps/jwql/bokeh_dashboard.py
```diff
@@ -37,7 +37,7 @@
 
 from bokeh.layouts import column
 from bokeh.models import Axis, ColumnDataSource, DatetimeTickFormatter, HoverTool, OpenURL, TapTool
-from bokeh.models import TabPanel, Tabs
+from bokeh.models.layouts import TabPanel, Tabs
```
bhilbert4, line 40: In bokeh_containers.py these imports do not include "layouts"

bhilbert4 (approved):
I haven't tried looking at the dashboard yet, but LGTM code-wise.

jwql/website/apps/jwql/bokeh_containers.py
BradleySappington: This function never worked and is not called.  bokeh_containers.py is NOT in good shape, but this isn't the PR to fix it.

bhilbert4:
The page looks about as good as it can get when running locally.

BradleySappington:
> Is it worth changing any of the database queries that use database_interface to use the django models instead?

@york-stsci 
I considered this and chose not to just because they will continue working with their current setup and I wanted to get the dashboard functional again.  We can discuss if its worth the time commitment to have django do them, and if so, can have it be part of the epic of the migration to the model system.

jwql/website/apps/jwql/bokeh_dashboard.py
BradleySappington: TabPanel and Tabs now exists in layouts as of bokeh 3.0

york-stsci:
Ultimately, at the end of the epic, the database_interface code will have been removed, so the answer for "is it worth the time commitment" is going to be "yes" eventually, but it certainly doesn't need to be a part of this PR specifically.

BradleySappington:
Well that makes that decision easy :)

jwql/website/apps/jwql/bokeh_dashboard.py
bhilbert4: So then should the import statement in bokeh_containers.py be updated to include "layouts"?

BradleySappington: 100% - Good catch

PR #1492, opened 2024-02-20
Claw monitor migrate

york-stsci:
Migrates the claw monitor to using the django models. This is an initial PR, with a second to follow which will add the new columns to the claw monitor.

### DIFF
--- a/jwql/instrument_monitors/nircam_monitors/claw_monitor.py
+++ b/jwql/instrument_monitors/nircam_monitors/claw_monitor.py
@@ -137,7 +137,7 @@ class ClawMonitor():
             measured vs model trending.
         """
 
-        columns = ['filename', 'filter', 'pupil', 'detector', 'effexptm', 'expstart_mjd', 'entry_date', 'mean', 'median', 
+        columns = ['filename', 'filter', 'pupil', 'detector', 'effexptm', 'expstart_mjd', 'entry_date', 'mean', 'median',
                    'stddev', 'frac_masked']  # , 'total_bkg']
 
         # Get all of the background data.
@@ -145,13 +145,6 @@ class ClawMonitor():
         df_orig = pd.DataFrame.from_records(background_data)
         # remove any duplicate filename entries, keep the most recent
         df_orig = df_orig.drop_duplicates(subset='filename', keep="last")
-
-        try:
-            jbt.get_background(ra, dec, wv, thisday=doy, plot_background=False, plot_bathtub=False,
-                               write_bathtub=True, bathtub_file='background_versus_day.txt')
-            bkg_table = Table.read('background_versus_day.txt', names=('day', 'total_bkg'), format='ascii')
-        except:
-            bkg_table = None
 
         # Get label info based on plot type
         if plot_type == 'bkg':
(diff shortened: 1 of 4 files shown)


jwql/instrument_monitors/nircam_monitors/claw_monitor.py
```diff
@@ -128,13 +137,21 @@ def make_background_plots(self, plot_type='bkg'):
...
-                              NIRCamClawStats.median, NIRCamClawStats.stddev, NIRCamClawStats.frac_masked, NIRCamClawStats.total_bkg).all()
-        df_orig = pd.DataFrame(query, columns=['filename', 'filter', 'pupil', 'detector', 'effexptm', 'expstart_mjd',
-                                               'entry_date', 'mean', 'median', 'stddev', 'frac_masked', 'total_bkg'])
-        df_orig = df_orig.drop_duplicates(subset='filename', keep="last")  # remove any duplicate filename entries, keep the most recent
+        background_data = NIRCamClawStats.objects.all().values(*columns)
+        df_orig = pd.DataFrame.from_records(background_data)
+        # remove any duplicate filename entries, keep the most recent
+        df_orig = df_orig.drop_duplicates(subset='filename', keep="last")
+
+        try:
```
mfixstsci: Where did this `try` block come from?

mfixstsci (approved):
Hey @york-stsci this looks good to me.

jwql/instrument_monitors/nircam_monitors/claw_monitor.py
york-stsci: It came from my trying to get the monitor to run. I had forgotten that the existing code in develop had the two new columns that don't actually exist yet, and so I had to either figure out how to get those values before they were used or I had to comment them out. I eventually decided that I probably couldn't get them, but forgot to remove the block I'd copied from lower down in the file in my attempt to see if I could figure out the background in time to divide by it.

Either way I've removed it now.

bsunnquist (approved):
This looks great to me - thanks a lot for getting this switched over to django models, the documentation with examples on how to work with them is especially useful.

Perhaps more relevant for the next PR, but I do store nans for the total_bkg if the JBT tool fails for whatever reason - do you know if django can store nans ok in the float columns, or should I switch it to something else?

jwql/instrument_monitors/nircam_monitors/claw_monitor.py
mfixstsci: Thank you @york-stsci just wanted to double check!

york-stsci:
As I recall, the "nan" value is explicitly defined in the IEEE double-precision floating point standard (which is the floating-point value that all of our database tables use), so it's possible that we might have to convert it from `np.nan` to `float.nan` (where that latter one isn't *actually* a thing in python as far as I know, but there's undoubtedly a way to generate a floating-point nan value).

jwql/instrument_monitors/nircam_monitors/claw_monitor.py
```diff
@@ -319,13 +331,12 @@ def process(self):
                                  'stddev': float(stddev),
                                  'frac_masked': len(segmap_orig[(segmap_orig != 0) | (dq & 1 != 0)]) / (segmap_orig.shape[0] * segmap_orig.shape[1]),
                                  'skyflat_filename': os.path.basename(self.outfile),
-                                 'doy': float(doy),
-                                 'total_bkg': float(total_bkg),
+                                 # 'doy': float(doy),
```
mfixstsci, line 334: Any reason for commenting these out?

jwql/website/apps/jwql/monitor_views.py
```diff
@@ -158,8 +157,9 @@ def claw_monitor(request):
     template = "claw_monitor.html"
 
     # Get all recent claw stack images from the last 10 days
-    query = session.query(NIRCamClawStats.expstart_mjd, NIRCamClawStats.skyflat_filename).order_by(NIRCamClawStats.expstart_mjd.desc()).all()
-    df = pd.DataFrame(query, columns=['expstart_mjd', 'skyflat_filename'])
+    query = NIRCamClawStats.objects  # .filter(expstart_mjd__gte=(Time.now().mjd - 10))
```
mfixstsci, line 160: I am assuming we will want to remove the comment this after the first run?

mfixstsci (approved):

PR #1494, opened 2024-02-20
create pull_jwql_branch.sh

BradleySappington:
create a bash script that will automate pulling new branches or updating current ones.
This is intended to be used during our release procedure but can also be used for devs to verify that their migrations are up to date when pulling from develop.

### DIFF
--- a/jwql/pull_jwql_branch.sh
+++ b/jwql/pull_jwql_branch.sh
@@ -1,8 +1,22 @@
 #!/bin/bash
 
 function echo_format {
-    echo "WARNING! the optional parameters should only be used during a JWQL release in production"
+    echo ""
     echo "Usage: $0 <branch> [-r|--reset_service] [-n|--notify <email@stsci.edu>]"
+    echo ""
+    echo "WARNING! the optional parameters should only be used during a JWQL release in production"
+    echo "branch: the git branch to pull from"
+    echo "[-r|--reset_service]: Reset the jwql service"
+    echo "[-n|--notify <email@stsci.edu>]: Notify via provided email"
+    echo ""
+    echo "Local:"
+    echo "$ bash pull_jwql_branch.sh develop"
+    echo ""
+    echo "Test:"
+    echo "$ bash pull_jwql_branch.sh v1.2 -r"
+    echo ""
+    echo "Production:"
+    echo "$ bash pull_jwql_branch.sh v1.2 -r -n group_email_address@stsci.edu"
 }
 
 # Check if the required number of arguments are provided
@@ -54,7 +68,7 @@ git fetch origin $branch_name
 git pull origin $branch_name
 git fetch origin --tags
 
-# 2. Bring the server down
+# 2. Bring the service down
 if [ "$reset" = true ]; then
     sudo /bin/systemctl stop jwql.service
 fi
@@ -65,7 +79,7 @@ pip install -e ..
 # 4. Merge Any Migrations
 python ./website/manage.py migrate
 
-# 5. Bring the server back up
+# 5. Bring the service back up
 if [ "$reset" = true ]; then
     sudo /bin/systemctl start jwql.service
 fi
--- a/jwql/pull_jwql_branch.sh
+++ b/jwql/pull_jwql_branch.sh
@@ -54,7 +54,7 @@ git fetch origin $branch_name
 git pull origin $branch_name
 git fetch origin --tags
 
-# 2. Bring the server down and back up
+# 2. Bring the server down
 if [ "$reset" = true ]; then
     sudo /bin/systemctl stop jwql.service
 fi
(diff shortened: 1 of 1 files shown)


BradleySappington:
@bhilbert4 @mfixstsci @york-stsci - This is the adaptable script that we were using for releases.  Now with some options handle via parameters so it can be used for local development.  Do you think anything else would be handy to add?  If not review appreciated.

mfixstsci:
Personally, since I have the least experience with django migrations, anything to make the process more streamlined and less prone to error is extremely useful. Let me take a look 👍

jwql/pull_jwql_branch.sh
```diff
@@ -0,0 +1,83 @@
...
+git pull origin $branch_name
+git fetch origin --tags
+
+# 2. Bring the server down
+if [ "$reset" = true ]; then
+    sudo /bin/systemctl stop jwql.service
+fi
+
+# 3. Install jwql
+pip install -e ..
```
mfixstsci, line 77: Looks like running the script is location dependent ie we have to be in the same directory as the script.

jwql/pull_jwql_branch.sh
```diff
@@ -0,0 +1,83 @@
...
+
+# 4. Merge Any Migrations
+python ./website/manage.py migrate
+
+# 5. Bring the server back up
+if [ "$reset" = true ]; then
```
mfixstsci, line 83: From what I see here, this script is also still useful on the servers as well. Locally we don't need to reset the servers nor notify anyone. Although I spamming your email might be fun @BradleySappington 😉

mfixstsci (commented):
@BradleySappington this is great and I think it's useful. Before approving I will see what the others think.

jwql/pull_jwql_branch.sh
BradleySappington: lol, correct, this is still intended to be our "release procedure".  I put a warning at the top to not use the optional parameters for local use.  (I also wanted to prevent us from sending multiple emails if we ever need to run this twice in a row...not that we've ever had to do that...)

jwql/pull_jwql_branch.sh
BradleySappington: Yes, currently set up to be run in place

jwql/pull_jwql_branch.sh
mfixstsci: 🤫 I don't know what you're talking about here...

mfixstsci:
Another question @BradleySappington do I need to be in a specific branch when running this script? Would there be any harm if someone were to set `branch_name` to `develop`? I know these branches are local to a user's forked repo (so do whatever man, it's your life) but I just wanted to double check on the implications that it could have for conflicts etc etc

BradleySappington:
@mfixstsci - if you are in develop and run this, there will just be a harmless failure on the checkouts that can be ignored

york-stsci:
I have only a tiny comment –  in the "echo_format" function, could you add a line explaining what "reset service" means? I'm guessing it means running "systemctl restart jwql" but I'm not quite sure.

BradleySappington:
@york-stsci - NP, I added the full set of usage instructions to echo_format.

jwql/pull_jwql_branch.sh
```diff
@@ -0,0 +1,96 @@
...
+    exit 1
+fi
+
+echo "Branch: $branch_name";
+echo "Reset: $reset";
+echo "Notify: $notify $recipient";
+
+# 1. Pull updated code from GitHub deployment branch (keep second checkout in case its already defined for some weird reason)
+git checkout -b $branch_name --track origin/$branch_name
+git checkout $branch_name
```
bhilbert4, line 66: Is this line needed? Shouldn't the line above create the new branch and switch to it?

jwql/pull_jwql_branch.sh
```diff
@@ -0,0 +1,96 @@
...
+git pull origin $branch_name
+git fetch origin --tags
+
+# 2. Bring the service down
+if [ "$reset" = true ]; then
+    sudo /bin/systemctl stop jwql.service
+fi
+
+# 3. Install jwql
+pip install -e ..
```
bhilbert4, line 77: What does the double period do?

bhilbert4 (approved):
Slick. Looks like this will make the release process easier.

york-stsci:
@BradleySappington Thank you, and looks good to me.

york-stsci (approved):
Looks like a very useful script

jwql/pull_jwql_branch.sh
york-stsci: I'm guessing parent directory. Given that this script is a level below JWQL's setup.py, that statement should install JWQL via pip in editable mode.

BradleySappington: Thats correct @york-stsci

jwql/pull_jwql_branch.sh
BradleySappington: The line is needed in case someone already has a branch defined (ie develop).  That first command will fail and leave them in their existing branch.  So we need to verify (a bit redundantly) that they are where they want to be.

mfixstsci:
@BradleySappington looks like everyone is in support! Once all the conversations are resolved, I am okay with merging!

PR #1495, opened 2024-02-20
Monitors setup migrations

york-stsci:
This PR officially adds the monitors tables to the migrations folder, adds two columns to the NIRCam claw monitor, and filters out NIRCam claw images by date during the database query rather than after retrieval.


jwql/website/apps/jwql/monitor_views.py
```diff
@@ -157,10 +157,9 @@ def claw_monitor(request):
     template = "claw_monitor.html"
 
     # Get all recent claw stack images from the last 10 days
-    query = NIRCamClawStats.objects  # .filter(expstart_mjd__gte=(Time.now().mjd - 10))
-    query = query.order_by('-expstart_mjd').all().values('expstart_mjd', 'skyflat_filename')
-    df = pd.DataFrame.from_records(query)
-    recent_files = list(pd.unique(df['skyflat_filename'][df['expstart_mjd'] > Time.now().mjd - 10]))
+    query = NIRCamClawStats.objects.filter(expstart_mjd__gte=(Time.now().mjd - 10))
```
mfixstsci, line 160: I see the changes now @york-stsci thank you!

jwql/instrument_monitors/nircam_monitors/claw_monitor.py
```diff
@@ -331,8 +331,8 @@ def process(self):
                                  'stddev': float(stddev),
                                  'frac_masked': len(segmap_orig[(segmap_orig != 0) | (dq & 1 != 0)]) / (segmap_orig.shape[0] * segmap_orig.shape[1]),
                                  'skyflat_filename': os.path.basename(self.outfile),
-                                 # 'doy': float(doy),
-                                 # 'total_bkg': float(total_bkg),
+                                 'doy': float(doy),
```
mfixstsci, line 333: Awesome, changes are here from last PR. Thanks!

jwql/website/apps/jwql/migrations/0017_centralstorage_fgsanomaly_filesystemcharacteristics_and_more.py
```diff
@@ -0,0 +1,622 @@
...
+                ('used', models.FloatField()),
+                ('available', models.FloatField()),
+            ],
+            options={
+                'db_table': 'filesystem_general',
+                'db_table_comment': 'monitors',
+                'managed': True,
+            },
+        )
+        migrations.CreateModel(
```
mfixstsci: I see some anomaly tables in here, they won't over write or reset anything when the migrations are performed, right?

jwql/website/apps/jwql/migrations/0018_centralstorage_fgsanomaly_filesystemcharacteristics_and_more.py
```diff
@@ -0,0 +1,612 @@
+# Generated by Django 4.2.5 on 2024-02-20 20:19
+
+import django.contrib.postgres.fields
+from django.db import migrations, models
+
+
+class Migration(migrations.Migration):
```
mfixstsci: I see the name is very similar to 0017, what is this migration doing differently from the first?

mfixstsci (commented):
Everything looks good to me @york-stsci thank you for this. Just a few questions and comments before I approve.

jwql/website/apps/jwql/monitor_views.py
york-stsci: Yeah. There's still something about how the data frame is built from the query that means that, even though the view only uses the filename field (given that it's already filtering by expstart), it seems like I have to retrieve both fields in order to then filter by uniqueness.

jwql/instrument_monitors/nircam_monitors/claw_monitor.py
york-stsci: Indeed, and I think the value (in existing entries) will end up being NAN, as the float equivalent to a NULL table value.

jwql/website/apps/jwql/migrations/0017_centralstorage_fgsanomaly_filesystemcharacteristics_and_more.py
york-stsci: Correct. These are the existing anomaly tables, by instrument, in the JWQLDB database, and they won't have any effect on the by-exposure anomaly table in DJANGODB.

jwql/website/apps/jwql/migrations/0018_centralstorage_fgsanomaly_filesystemcharacteristics_and_more.py
york-stsci: Nothing at all. Same with 19 and 20. For whatever reason, things went as follows:

- I create a migration with makemigration
- I run it with migrate.py and it causes an error or doesn't actually implement the changes, and it suggests running makemigration again
- I do so, and it makes a second pretty much identical migration
- This one works

I have, frankly, no idea what happened.

jwql/instrument_monitors/nircam_monitors/claw_monitor.py
mfixstsci: That's okay, the claw monitor doesn't take too long to run and only has ~1 month of data in it so we can just run it to get all of the fields we need populated.

jwql/website/apps/jwql/migrations/0017_centralstorage_fgsanomaly_filesystemcharacteristics_and_more.py
mfixstsci: Thank you!

jwql/website/apps/jwql/migrations/0018_centralstorage_fgsanomaly_filesystemcharacteristics_and_more.py
mfixstsci: Strange, maybe when @BradleySappington does his deep dive he might see something we aren't seeing.

jwql/website/apps/jwql/migrations/0017_centralstorage_fgsanomaly_filesystemcharacteristics_and_more.py
```diff
@@ -0,0 +1,622 @@
...
+                ('used', models.FloatField()),
+                ('available', models.FloatField()),
+            ],
+            options={
+                'db_table': 'central_storage',
+                'db_table_comment': 'monitors',
+                'managed': True,
+            },
+        )
+        migrations.CreateModel(
```
BradleySappington: I assume these anomaly models are being included because they will be initially created, then we will have a migration to delete them once our logic is updated from sqlalchemy to django for each model?

bhilbert4 (approved):
LGTM

jwql/website/apps/jwql/migrations/0018_centralstorage_fgsanomaly_filesystemcharacteristics_and_more.py
BradleySappington: I'm suggesting some overall structural changes, and if agreed upon and implemented we should put the database back to where you began, and re-run make migrations.  If we run into a weird situation like this lets address it when it pops up.  I noticed (from experience) that if something is weird locally, it will not be smooth sailing down the line.

jwql/website/apps/jwql/monitor_models/common.py
```diff
@@ -191,15 +191,15 @@ def db_for_read(self, model, **hints):
         """
```
BradleySappington: The `MonitorRouter` class should be in its own file called `router.py` and not associated with the common monitor functionality.  This router is a system wide addition (apologies, as this comment should have been addressed in the previous PR).  I personally think `router.py` should live parallel to `models.py` however if you choose to keep it in the sub folder thats fine.

jwql/website/apps/jwql/monitor_models/common.py
```diff
@@ -191,15 +191,15 @@ def db_for_read(self, model, **hints):
...
+        if model._meta.db_table_comment in self.route_app_labels:
             return "monitors"
         return None
 
     def db_for_write(self, model, **hints):
         """
         Attempts to write  monitor models go to monitors db.
         """
-        if model._meta.app_label in self.route_app_labels:
+        if model._meta.db_table_comment in self.route_app_labels:
```
BradleySappington: If we aren’t using `db_table_comment` to align the models with their respective db, we should use the monitor_model names from a list that should be defined in `constants.py`.  This list will be accessed inside the monitor_router class.  An example of how this would be done is: 

```
from constants import monitor_models
def db_for_write(self, model, **hints):
    if model._meta.app_label == ‘jwql’:
        if model.__name__ in monitor_models:
                return ‘monitors’
    return None
```
Note: the above suggestion should be applied to all elements in the router, with the exception of allow_migrate which will use model_name  in its check as that gathered as a parameter.

BradleySappington (changes requested):
Great work so far @york-stsci, I do have some feedback that we can discuss.  Most of these comments are in the code review, but I want to consolidate them here with a few overall additions so my feedback is cohesive.

1. The `MonitorRouter` class should be in its own file called router.py and not associated with the `common_monitor` functionality.  This router is a system wide addition (apologies, as this comment should have been addressed in the previous PR).  I personally think `router.py` should live parallel to `models.py` however if you choose to keep it in the sub folder thats fine. I’ll continue my explanation assuming it is up a level for brevity.

2.  We also need to set django up to see our `router.py` as part of its settings.  Otherwise it will just sit there as a class.  In order to do this we need to update `jwql/website/jwql_proj/settings.py`  by adding the path to our router `DATABASE_ROUTERS = ['jwql.router.monitor_router']`  At this point, all db access defined in `MonitorRouter` should be automatically handled by django.

3. `db_table_comment` is being used functionally to associate models with databases.  This is a clever workaround but not the intended use of this field.  `db_table_comment` should be used “for documenting database tables for individuals with direct database access who may not be looking at your Django code.”  We should put `app_label` back.  I’m also not sure if `app_label` should be `monitors` or `jwql`, as we don’t have a separate monitors app associated with this repository (which is why I’d lean `jwql`).

4. If we aren’t using `db_table_comment` to align the models with their respective db, we should use the monitor model names from a list that should be defined in `constants.py`.  This list will be accessed inside the `MonitorRouter` class.  An example of how this would be done is: 
```
from constants import monitor_models
def db_for_read(self, model, **hints):
    if model._meta.app_label == ‘jwql’:
        if model.__name__ in monitor_models:
                return ‘monitors’
    return None
```
Note: the above suggestion should be applied to all elements in the router, with the exception of allow_migrate which will use the parameter `model_name`  instead of `model.__name__ `

Happy to discuss further if you like!

bsunnquist (changes requested):
@york-stsci Look great! Only thing I see is to change the `plot_type=='model'` cell back to this:
`if plot_type == 'model':`
 `    plot_data = df['median'].values / df['total_bkg'].values`

jwql/website/apps/jwql/migrations/0017_centralstorage_fgsanomaly_filesystemcharacteristics_and_more.py
york-stsci: Correct. They're included because the tables exist, and the initial migration assumes control over all of the existing tables.

BradleySappington: great, thanks!

jwql/website/apps/jwql/monitor_models/common.py
york-stsci: I decided to move it, and have implemented the name-based constant as described, and updated settings accordingly.

jwql/website/apps/jwql/monitor_models/common.py
york-stsci: See above, with the note that the thing that allow_migrate really uses as a parameter is "database", at least for our (single-app) setup.

jwql/website/apps/jwql/router.py
```diff
@@ -0,0 +1,65 @@
...
+            return "monitors"
+        return None
+
+    def allow_relation(self, obj1, obj2, **hints):
+        """
+        Allow relations between tables in the monitors DB.
+        """
+        if (
+            obj1.__class__.__name__ in MONITOR_TABLE_NAMES
+            or obj2.__class__.__name__ in MONITOR_TABLE_NAMES
```
BradleySappington: should this be `and`?

jwql/website/apps/jwql/router.py
```diff
@@ -0,0 +1,65 @@
...
+        ):
+            return True
+        return None
+
+    def allow_migrate(self, db, app_label, model_name=None, **hints):
+        """
+        Make sure the monitors apps only appear in the 'monitors' database.
+        """
+        if app_label in self.route_app_labels or app_label == 'jwql':
```
BradleySappington: We dont want everything marked as `app_label == 'jwql'` to be in the monitors database.  You can still do the model_name check here I believe and just use the parameter `model_name`

BradleySappington (changes requested):
Thanks Brian! Few small bugs I noticed

jwql/website/apps/jwql/router.py
york-stsci: Given Django right now, probably, because it doesn't support object relations between databases. Fixed.

jwql/website/apps/jwql/router.py
york-stsci: Note that this is a two-stage process where the app has to be "jwql" *and* the database has to be "monitors". Essentially, this logic:

- If the app isn't "jwql" or "monitors", returns None, because this router has no opinion on other apps
- If the app *is* "jwql", but the database *isn't* monitors, returns False, because this router shouldn't be used for any other database in the jwql app
- If the app is "jwql" *and* the database is "monitors", returns True because this router should then be used.

york-stsci:
Note that, for databases *other* than dev, we might need to fake migration 17, because we can't use `--fake-initial` on any migration other than the first, and currently when you try to apply migration 17 it quite reasonably complains that the tables already exist.

jwql/website/apps/jwql/router.py
BradleySappington: ah yes, I misread the return.  A few things to consider

1) Isn't app_label no longer going to be set to 'monitors'?  That check may not be necessary.
2) I still believe there is an issue if you dont check the `model_name`.  When you run `migrate`, `allow_migrate` is run on every model and database combo, so if you're not verifying the modelname, then there could be model/db permission mixmatches.

york-stsci: I think I see what you're getting at. Does the code (as just committed) fix this?

BradleySappington: Looks good logically to me

jwql/website/apps/jwql/migrations/0018_centralstorage_fgsanomaly_filesystemcharacteristics_and_more.py
BradleySappington: looks like the migrations got worked out with those changes

BradleySappington (approved):

jwql/instrument_monitors/nircam_monitors/claw_monitor.py
```diff
@@ -193,7 +193,7 @@ def make_background_plots(self, plot_type='bkg'):
                     plot_data = df['stddev'].values
                 if plot_type == 'model':
                     total_bkg = [1. for x in df['median'].values]
```
bsunnquist: I think this line isn't needed anymore, but it looks harmless:
`total_bkg = [1. for x in df['median'].values]`

bsunnquist (approved):

jwql/instrument_monitors/nircam_monitors/claw_monitor.py
york-stsci: You're right, good catch, fixed.

york-stsci:
And this is the part where I realize that I ran the migrations before updating the router class. Hang on whilst I rock the boat (i.e. roll back, then forward, repeat as required)

york-stsci:
Logging this here for reference: It is entirely possible that migration 17 will fail when run with `--database=monitors`. If it does, then migration 17 (and *only* migration 17) should be run as `migrate.py --fake` when run with `--database=monitors`. This probably means we can't use the shell script, either on dev or production, in this specific case, because we have two migrations and may need to apply them one-at-a-time for one (but not both) database(s).

mfixstsci:
I am good with this @BradleySappington seems happy as well as @bsunnquist. @york-stsci let me know when this is ready to merge.

york-stsci:
I'd like @BradleySappington to take another look at the current form, just to backstop me, and then I think we're good.

mfixstsci:
> I'd like @BradleySappington to take another look at the current form, just to backstop me, and then I think we're good.

I think this is a great idea. @BradleySappington let me know when you are finished with your review and we can get this in!

BradleySappington (approved):
Great work Brian

PR #1497, opened 2024-02-22
Update Bad Pixel Monitor to use Django DB Models

mfixstsci:


bhilbert4 (commented):
These changes look good to me. What about in monitor_bad_pixel_bokeh.py? That should be querying the database in order to produce the plots, right?

mfixstsci:
> These changes look good to me. What about in monitor_bad_pixel_bokeh.py? That should be querying the database in order to produce the plots, right?

@bhilbert4 you are right, I just looked at the file. I will go ahead and update it.

mfixstsci:
@bhilbert4 okay, so the updates to `monitor_bad_pixel_bokeh.py` are in. I think should do it, let me know if I missed anything.

jwql/website/apps/jwql/monitor_pages/monitor_bad_pixel_bokeh.py
```diff
@@ -263,40 +265,33 @@ def __init__(self, pixel_table, instrument, detector):
...
+
+            if record is None:
+                self.new_bad_pix[bad_type] = ([], [])
+                self.background_file[bad_type] = ''
+                self.obs_start_time[bad_type] = datetime.datetime.today()
+                self.obs_end_time[bad_type] = datetime.datetime.today()
+                self.num_files[bad_type] = 0
+                self.baseline_file[bad_type] = ''
+            else:
+                self.new_bad_pix[bad_type] = (record.x_coord, record.y)
```
bhilbert4: Should this be y_coord?

mfixstsci:
With @york-stsci 's updates are we prepared to merge? @bhilbert4 would you like to review the changes with Brian's changes in?

bhilbert4 (changes requested):
This looks good to me. I hope you used a tool like Black rather than changing all those single quotes to double quotes manually. :) 

The only issue is that the migration file needs to be included.

jwql/website/apps/jwql/monitor_pages/monitor_bad_pixel_bokeh.py
york-stsci: Yes, it should be, based on the table definition.

bhilbert4 (approved):
LGTM. Thanks for clearing up the empty defaults!

york-stsci:
Looks good

PR #1500, opened 2024-02-26
Fixed the monitors DB router

york-stsci:
With the new router, I have specifically tested that:

- For operations involving monitor tables, migrations generate SQL in the monitors database, and no-ops otherwise
- For operations not involving monitor tables, migrations generate SQL in the default database, and no-ops in the monitors database.


mfixstsci (approved):
Looks good @york-stsci let's wait for the CI to wrap up and I will merge it.

PR #1509, opened 2024-03-07
Migrate dbs script

BradleySappington:
ensure both jwql databases have migrate applied during standard branch pull


BradleySappington:
tested this locally and it works, should keep everything smooth and updated in all databases moving forward starting AFTER this upcoming release

bhilbert4 (approved):
Makes sense to me, assuming that any migration messiness is going to be dealt with locally, and in the end the servers won't need to worry about rolling back/faking migrations.

BradleySappington:
@bhilbert4, servers will need to fake this upcoming 17, but after that no more

PR #1514, opened 2024-03-14
Fix Bokeh `file_html` Call

mfixstsci:
Bokeh's `file_html` method takes multiple optional parameters. Without specifying the parameter names, the method gets confused by the positions of the parameters passed in the method call.


bhilbert4 (approved):
LGTM, thanks!

PR #1516, opened 2024-03-14
Add max length to charfield

BradleySappington:
Define Constants and add Django max_length parameter to CharField


bhilbert4 (approved):
Looks good to me. I was a little nervous about the aperture length, but I checked in pysiaf and the longest aperture name I've found is only 28 characters, so it seems like we're all good there.

BradleySappington:
@bhilbert4, would you like me to increase anyway?  a lot of the numbers are arbitrary

bhilbert4:
@BradleySappington I think it's ok as is. Should we be worried about anyone who creates a filename that's longer than 100 characters?

BradleySappington:
@bhilbert4, 100 was the previous value so I kept it.  Happy to increase if you think its important?

BradleySappington:
@bhilbert4 , scratch that, It's arbitrary, I'll bump it up to 150

mfixstsci (approved):
LGTM

BradleySappington:
bumping up to 1k since most filenames use full path

PR #1517, opened 2024-03-14
Header fix

BradleySappington:
Crashes during parallel observations when not all header information exists


BradleySappington:
NOTE: should go in after #1516

bhilbert4 (approved):
Nice and easy. I like it.

PR #1526, opened 2024-03-20
remove test code

BradleySappington:
BRADTEST is sneaky and must be un-snuck


mfixstsci (commented):
So you're telling me `BRADTEST` is not a valid header keyword? 😎

mfixstsci (approved):

PR #1527, opened 2024-03-22
Fix database empty fields

BradleySappington:
Our databases have been populated using changing methodologies over the years, with changing default values.  
There are also instances that occur where a database ingestion may not complete properly, and not all datafields are entered.

Because of this issue, our PR will do the following:
1) Create a consistent default value that is not currently used by any field (we chose the string 'empty')
2) Create a methodology to consistently check for empty fields in our database and automatically re-ingest that row's information automatically.


jwql/website/apps/jwql/archive_database_update.py
```diff
@@ -387,6 +402,13 @@ def fill_empty_model(model_name, model_field):
     except ValueError:
         pass
 
+    # filter(field__exact=DEFAULT_MODEL_CHARFIELD)
```
mfixstsci, line 405: Did you want to remove this line?

jwql/website/apps/jwql/archive_database_update.py
```diff
@@ -369,10 +380,14 @@ def fill_empty_model(model_name, model_field):
...
+    is_proposal = (model_name == "proposal")
+    is_rootfileinfo = (model_name == "rootfileinfo")
+    rootfile_info_fields_default_ok = ["filter", "grating", "pupil"]
+
     model_field_null = model_field + "__isnull"
     model_field_empty = model_field + "__exact"
 
-    model = apps.get_model('jwql', model_name)
+    model = apps.get_model("jwql", model_name)
     null_models = empty_models = zero_models = model.objects.none()
```
mfixstsci, line 391: Fancy, I like it.

mfixstsci (approved):
Hey @BradleySappington everything looks good here, just a couple of comments/questions.

jwql/website/apps/jwql/archive_database_update.py
BradleySappington: Keeping that there, that's a comment showing what the kwargs expansion does, there is one like that above each try statement since the code isn't the most legible

PR #1535, opened 2024-04-05
add default to read_patt_num

BradleySappington:
add default value of 1 to read_patt_num missed our last migration


bhilbert4 (approved):
Migrate!

PR #1544, opened 2024-04-18
final model define for faking

BradleySappington:
There were some missed modules that have existing tables in them.  
updated init file to see all monitor modules that define mondels so we can fake the table creations.
this will allow future table modification/removal for any of these.

this migration WILL need to be faked for our existing database structure
`$python manage.py migrate jwql --database=monitors --fake`


bhilbert4 (commented):
Should the other bias models be in here as well?
NIRCamBiasStats
NIRISSBiasQueryHistory
NIRISSBiasStats
NIRSpecBiasQueryHistory
NIRSpecBiasStats

And then the other TA models?
MIRITaStats
NIRSpecTaQueryHistory
NIRSpecTaStats

BradleySappington:
> Should the other bias models be in here as well? NIRCamBiasStats NIRISSBiasQueryHistory NIRISSBiasStats NIRSpecBiasQueryHistory NIRSpecBiasStats
> 
> And then the other TA models? MIRITaStats NIRSpecTaQueryHistory NIRSpecTaStats

@bhilbert4, just having one class from each file will load all of the classes in that file.  All of those mentioned above are successfully in the migration.

mfixstsci:
@BradleySappington from my accidental `makemigration` these were the tables made:

- [x] Create model MIRITaQueryHistory
- [x] Create model MIRITaStats
- [x] Create model NIRSpecTaQueryHistory
- [x] Create model NIRSpecTaStats
 
I see them all in the migration file (`0023`)

BradleySappington:
I will handle making them in my PR, so  hold off on making any migrations until 1544 is merged into development

bhilbert4 (approved):

PR #1545, opened 2024-04-18
RC v1.2.8 prep

BradleySappington:
Final monitor models that have not yet been "defined" in our django models
this migration will be faked as the database tables already exist


BradleySappington:
this is wonky

PR #1546, opened 2024-04-25
Update Redis Package Names in Environment Files

mfixstsci:


mfixstsci:
@BradleySappington So the `python-3.9/osx` build is failing because of the astropy dependency requirements to bump up to python 3.10. Want to remove python 3.9 env in this PR? We can then remove the test requirement in the settings as well.

mfixstsci:
@BradleySappington @bhilbert4 there is something going on with the command `conda` with the new `mamba` action we are using. @zacharyburnett I removed the line and the tests pass now 🤷‍♂️ I think keeping track of what our dependencies look like when running the tests is important but am curious why `conda export env` and `mamba export env` are not working.

mfixstsci:
> @BradleySappington @bhilbert4 there is something going on with the command `conda` with the new `mamba` action we are using. @zacharyburnett I removed the line and the tests pass now 🤷‍♂️ I think keeping track of what our dependencies look like when running the tests is important but am curious why `conda export env` and `mamba export env` are not working.

Looks like we needed `micromamba` instead of `conda` or `mamba` in the export statement. Looks good now 👍

BradleySappington (approved):
LGTM

PR #1549, opened 2024-04-30
create generic error page to handle exceptions in views.

BradleySappington:
When our code crashes, we need to have a proper page to display for our users.  Created a page that we can begin implementing in various views during try/except logic.
applied to two known issue areas.


bhilbert4 (approved):
Slick. I like it. Future improvement: a rotating cast of JWQL mascot images.

BradleySappington:
links with https://jira.stsci.edu/browse/JWQL-194

PR #1550, opened 2024-05-01
debug false by default

BradleySappington:
our test and production environments are taking a performance hit by running debug mode.  We need to disable it


bhilbert4 (approved):
huzzah!

BradleySappington:
@bhilbert4 made it configurable

bhilbert4:
Nice. That'll make it much easier to keep debug on when working locally

PR #1586, opened 2024-05-10
Remove filename parser test over filesystem

### LINKED ISSUE #1308: Remove test_filename_parser_whole_filesystem() from test_utils.py
The test above runs the filename parser on ALL files in the filesystem. 

This takes an excessively long time to run, and it's not clear what the value of the test is. The test is skipped on github actions, so it's not run very often, but we should still either limit the files that it runs on, or remove it completely.

bhilbert4:
Resolves #1308 

Remove test_filename_parser_whole_filesystem().  This isn't a very useful test and would take ages to run now that we have a large filesystem.


bhilbert4:
@mfixstsci I came across this old open issue and decided to get in a quick fix. This is ready for review.

mfixstsci (approved):
Agree with the issue that caused this change. If we skip it and it take forever to run locally, then we can remove it.

PR #1591, opened 2024-05-15
Update remote to upstream in pull_jwql_branch.sh

bhilbert4:
Updates the remote used by pull_jwql_branch.sh to 'upstream' from the current 'origin'

This will allow the script to be used locally, since most developers use 'upstream' to refer to the spacetelecope fork of the repo.

In concert with this PR, we'll need to update the remote names on the servers with the same switch.

This resolves Jira issue JWQL-242


bhilbert4:
The remotes on the servers have been changed from origin to upstream, so this updated script should now work.

BradleySappington (approved):

PR #1592, opened 2024-05-23
Customize colorbar location and size based on aperture size

### LINKED ISSUE #1560: Fix preview images where vertical colorbars on subarray thumbnails are unreadable
_Issue [JWQL-195](https://jira.stsci.edu/browse/JWQL-195) was created on JIRA by [Bryan Hilbert](https://jira.stsci.edu/secure/ViewProfile.jspa?name=hilbert):_

e.g. <https://jwql.stsci.edu/nircam/exposure/jw04442067001_03201_00001/>

 

We should find a way to tweak the bokeh code in order to assure that that colorbar is always legible.

bhilbert4:
Resolves #1560 

This improves the readability of the colorbar, especially for cases where the aperture is very non-square (e.g. SUBSTRIPE64). In full frame images, the colorbar has been shrunk to be the same height as the detector image. This effectively allows the image itself to be larger when displayed on the webpage.

I've tested this on all xsize/ysize ratios for all apertures across all instruments. All images look good.
I also ran generate_preview_images.py locally with no problems.


mfixstsci (approved):
Thanks @bhilbert4 this is a clever fix.

PR #1593, opened 2024-05-28
fix release upload step condition to match workflow trigger

zacharyburnett:
the event type in the `if:` condition (`published`) was not the same as the event type in the trigger (`released`). Thus, the workflow would never satisfy the condition 😅


mfixstsci (approved):
Thanks @zacharyburnett

PR #1595, opened 2024-05-28
fix version matching pattern

zacharyburnett:
use `default_unprefixed` SemVer pattern to grab after version


zacharyburnett:
fixed output filename to include suffixes: [image](https://github.com/spacetelescope/jwql/actions/runs/9288470296)

mfixstsci (approved):

PR #1596, opened 2024-05-29
updating freeze matrix to include linux, mac and python 3.12

mfixstsci:


bhilbert4 (approved):
LGTM. So this is now going to create an environment file asset for each OS/python combination?

PR #1600, opened 2024-05-30
Adjust view image and view exposure to allow for larger preview images

### LINKED ISSUE #1584: Make preview images larger
_Issue [JWQL-214](https://jira.stsci.edu/browse/JWQL-214) was created on JIRA by [Bryan Hilbert](https://jira.stsci.edu/secure/ViewProfile.jspa?name=hilbert):_

From the JWQL user survey (NIRISS team):

Make the preview jpegs bigger. It would be great if these were up to 2x bigger, I think it would be much easier to spot anomalies. While I'm on it, might help to scale the colorbar to the plot height, which could then let you increase the overall jpeg size 

The submitted mentioned that this is one of their top 3 improvements, priority wise.

Preview image size probably can't get much larger in the case of exposure level pages for NIRCam and maybe NIRSpec. But at the detector level pages, they could be made larger. Scaling the size of the colorbars would be helpful as well. This is partially captured in [JWQL-195](https://jira.stsci.edu/browse/JWQL-195)

bhilbert4:
Resolves #1584  ([JWQL-214](https://jira.stsci.edu/browse/JWQL-214))

Based on feedback from the user survey.
Make preview images larger, to allow for easier QL checks.

### DIFF
--- a/jwql/website/apps/jwql/static/css/jwql.css
+++ b/jwql/website/apps/jwql/static/css/jwql.css
@@ -365,6 +365,11 @@
       display: inline-block;
   }
 
+  .image_views {
+      margin-left: 2%;
+      margin-right: 2%;
+  }
+
   #loading {
     text-align:center;
     margin: 0 auto;
--- a/jwql/website/apps/jwql/templates/view_exposure.html
+++ b/jwql/website/apps/jwql/templates/view_exposure.html
@@ -8,7 +8,7 @@
 
 {% block content %}
 
-    <main role="main" class="container">
+    <main role="main" class="image_views">
     	<!-- Show exposure group name -->
     	<h3>Exposure {{ group_root }}</h3>
 
@@ -119,7 +119,7 @@
                             <a href="{{ base_url }}/{{ inst }}/{{ group_root }}_{{ detector1 }}/">
                                 <img class="image_preview_viewer" id="image_viewer_{{ detector1 }}"
                                      title="Preview image for {{ group_root }}, detector {{ detector1 }}"
-                                     data-detector="{{ detector1 }}" style="width:100%">
+                                     data-detector="{{ detector1 }}" style="width:125%">
                             </a>
                         </span>
                     </div>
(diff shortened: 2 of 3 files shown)


bhilbert4:
From testing using the current preview images, as well as the new preview images from #1592, the values to use for the image widths:
In view_image:
new preview images: 1000rem looks good. I tried switching to a percentage, but that moved the image to the right far enough that we couldn't take advantage of the smaller margin, and the image had to be smaller.
old preview images: Could go larger, but 1000rem looks fine.

view_exposure:
new preview images: 100% was best in all cases
old_preview images: With 2 detectors per row, 125% looked good. With 4 detectors per row, 110% looked good. 

So it probably makes sense to keep the widths in view_exposure set to 100% (which is what they are set to already) under the assumption that eventually all programs will have the newer preview images created from #1592 

For view_image, 1000rem seems reasonable, at least in terms of size. 

Good test cases when working in the nircam test account:
http://127.0.0.1:8000/NIRCam/exposure/jw01062004001_02101_00001/  - 2 det/row, new prev images
http://127.0.0.1:8000/NIRCam/exposure/jw01068004001_02101_00001/ - 2 det/row, old prev images
http://127.0.0.1:8000/NIRCam/exposure/jw01063001002_02101_00001/ - 4 det/row, new prev images
http://127.0.0.1:8000/NIRCam/exposure/jw01409034001_02102_00001/ - 2 det/row, new, short, wide prev images
http://127.0.0.1:8000/NIRISS/exposure/jw01538065001_02107_00004/ - single det, new, tall, narrow prev image

bhilbert4:
@mfixstsci @BradleySappington I think this is ready for review. My CSS is weak, so I'm not sure if there is a better way to do what I've done here. I basically made a copy of the margin settings from the Dashboard page and applied them to the image level and exposure level pages.

bhilbert4:
Note that in the case of a very narrow but tall preview image, these changes make the preview image quite large. But that actually makes it a lot easier to see what's going on in such a small subarray. Example case when running on the nircam development account: http://127.0.0.1:8000/NIRISS/jw01538065001_02107_00004_nis/

mfixstsci (approved):
Hey @bhilbert4 I just checked the preview images linked in the comment and they look good.

PR #1602, opened 2024-05-31
Add keyword to specify program IDs for preview image creation

bhilbert4:
Add an optional "programs" keyword to generate_preview_images.py. This allows the user to specify a list of program ID numbers (e.g. 1068, 01220) to the call to generate_preview_images.py. If this keyword is given, the script will create preview images only for the specified programs. If no program keyword is given the script will create preview images for all programs, as it did previously.

With this new keyword, we can split up the creation of new preview images into multiple calls (on multiple machines), and speed up the creation of new preview images. Or, if we find we need to update the preview images for only a few, specific programs, this will allow us to do that.


bhilbert4:
This was inspired by #1592 where is seems like eventually we'll want to replace all preview images.

bhilbert4:
@mfixstsci this is also ready for review. Local testing showed no problems.

mfixstsci (approved):
Cool, this looks good to me.

PR #1603, opened 2024-05-31
[SCSB-155] build with Numpy 2.0 release candidate

zacharyburnett:
resolves [SCSB-155](https://jira.stsci.edu/browse/SCSB-155)

Numpy 2.0 is scheduled for release on June 19th. Due to the C ABI updates, Python projects that build C extensions are required to build against Numpy 2.0 in order to use Numpy 2.0 in runtime.
This is backwards-compatible with Numpy versions in runtime as far back as `1.26`

this PR pins `numpy>=2.0.0rc2` in `build-system.requires`
> [!NOTE]
> this PR was generated automatically by ``batchpr`` :robot:

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -71,7 +71,7 @@ content-type = "text/plain"
 [build-system]
 requires = [
     "setuptools>=68.0.0",
-    "numpy>=2.0.0rc2",
+    "numpy>=2.0.0",
     "wheel",
     "setuptools_scm",
 ]
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -71,7 +71,7 @@ content-type = "text/plain"
 [build-system]
 requires = [
     "setuptools>=68.0.0",
-    "numpy",
+    "numpy>=2.0.0rc2",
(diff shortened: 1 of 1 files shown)


mfixstsci:
@zacharyburnett it's been a long wait on this one. We just had a release and need to do another soon and figured if numpy 2.0 isn't causing too much havoc we could merge it in. I am guess since numpy 2.0 it out, we can remove the `rc` from the pin?

pyproject.toml
```diff
@@ -71,7 +71,7 @@ content-type = "text/plain"
 [build-system]
 requires = [
     "setuptools>=68.0.0",
-    "numpy",
+    "numpy>=2.0.0rc2",
```
zacharyburnett:
```suggestion
    "numpy>=2.0.0",
```

mfixstsci:
@zacharyburnett I have been holding off on this because of the dependencies of the JWST pipeline. I would like to bump to numpy 2.0 but am unsure if we are ready yet.

PR #1607, opened 2024-06-13
Remove old presentations from repo

bhilbert4:
These were taking up so much room that Pypi was starting to refuse our overly-large tar file.


BradleySappington (approved):
Thanks!

PR #1608, opened 2024-06-13
Num results fix

BradleySappington:
currently num results per page is not actually used when selected on the query page.
Lets use it (and default to 500 now that the pages will load faster)

### DIFF
--- a/style_guide/example.py
+++ b/style_guide/example.py
@@ -73,7 +73,7 @@ from jwql.utils.logging_functions import configure_logging, log_info, log_fail, 
 
 # Global variables should be avoided, but if used should be named with
 # all-caps
-A_GLOBAL_VARIABLE = 'foo'  # type: str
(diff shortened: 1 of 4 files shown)


BradleySappington:
resolves https://jira.stsci.edu/browse/JWQL-151

style_guide/example.py
```diff
@@ -73,7 +73,7 @@
 
 # Global variables should be avoided, but if used should be named with
 # all-caps
-A_GLOBAL_VARIABLE = 'foo'  # type: str
+A_GLOBAL_VARIABLE = "foo"  # type: str
```
bhilbert4, line 76: Oh, a PEP8 stickler, I see. :)

bhilbert4 (approved):
LGTM!
Thanks for alphabetizing those imported objects.

PR #1609, opened 2024-06-15
Add Environment Update Script

mfixstsci:
This is the final portion of the release process. This script allows user to download the environment used in a release, install it and set it as the environment `jwql-current`.

General Use Case:
`source install-env.sh <tagged_version> <python_version>`

Specific Use Case:
`source install-env.sh v1.2.9-a.1 3.11`


york-stsci:
The script (and its underlying concept) look good to me. Will approve after seeing it in action on dev.

mfixstsci:
@york-stsci We did the demo a few weeks ago and was hoping I could get approval. @BradleySappington and I are going to perform a release next Wednesday at 3PM EST if you would like to join us.

york-stsci (approved):
Looks good to me.

PR #1613, opened 2024-07-09
Add new NIRISS AMI-related suffixes

bhilbert4:
This adds a few new NIRISS AMI-related filename suffixes to the list of recognized suffix values. These are new data products recently added to pipeline outputs.

### DIFF
--- a/jwql/utils/constants.py
+++ b/jwql/utils/constants.py
@@ -785,7 +785,7 @@ NIRCAM_SUBARRAYS_ONE_OR_FOUR_AMPS = [
 ]
 
 # Possible suffix types for AMI files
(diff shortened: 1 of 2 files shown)


bhilbert4:
@mfixstsci this is ready to go. 

Thinking a bit about the filename parser, do you think it's worth changing it such that when it finds a filename it can't parse, it returns None rather than raising an exception? That would prevent things like the preview image generator or the archive database update from stopping completely when there's one unrecognized file. @BradleySappington

mfixstsci (approved):
LGTM @bhilbert4 thanks for the changes.

BradleySappington:
@bhilbert4 yeah, its probably worth doing that work, may want to notify us if there is an unrecognized file just so we can adjust?

PR #1614, opened 2024-07-10
filename_parser() no longer raises an exception for unrecognized files

bhilbert4:
Adjust the filename_parser() such that when it is called on a filename it doesn't recognize, it no longer raises an exception, but instead returns a dictionary with a key indicating that the filename wasn't recognized.


bhilbert4:
@mfixstsci @york-stsci I've cleaned up the changes here to make them a little more consistent amongst the various calls to filename_parser(). Tests are all passing, so I think this is ready for review. I'll take care of the PEP8 stuff momentarily.

jwql/jwql_monitors/generate_preview_images.py
```diff
@@ -212,6 +212,11 @@ def check_existence(file_list, outdir):
         # for the appropriately named jpg of the mosaic, which depends
         # on the specific detectors in the file_list
         file_parts = filename_parser(file_list[0])
+
+        # If filename_parser() does not recognize the filename, return False
+        if not file_parts['recognized_filename']:
```
york-stsci, line 217: This code (combined with the above code at line 214) implies that if you have an invalid file, you will still get a dictionary back, it's just that the dictionary will have a "recognized_filename" key that is, presumably, a boolean value, with True for a recognized file and False for a non-recognized file.

jwql/jwql_monitors/generate_preview_images.py
```diff
@@ -253,7 +258,11 @@ def create_dummy_filename(filelist):
     modules = []
     for filename in filelist:
         indir, infile = os.path.split(filename)
-        det_string = filename_parser(infile)['detector']
+        try:
+            det_string = filename_parser(infile)['detector']
+        except KeyError:
+            # If filename_parser() does not recognize the file, skip it
+            continue
```
york-stsci, line 270: These lines, on the other hand, imply the filename_parser() will throw an exception for an unrecognized file. Or, possibly, that you still get the dictionary back, but rather than checking for whether it's a recognized filename, you instead just try to access the keyword and continue on an exception.

If feels to me as if, whether or not we want to do the same thing in both cases, we at least want to use the same idiom in calling filename_parser(). Here, that might look like

```
file_parts = filename_parser(infile)
if not file_parts['recognized_filename']:
    continue
det_string = file_parts['detector']
```

york-stsci: My second note is that, on failure, we absolutely need to log an error (or warning), so that we know that there's something that's causing the failures. We don't want to crash, but we *do* want to know.

jwql/jwql_monitors/generate_preview_images.py
york-stsci: Also, because I forgot, we need to log a warning or error at this point so that we know what happened, and can look through the logs to find it, and to know if we need to update the parsing code.

jwql/jwql_monitors/generate_preview_images.py
```diff
@@ -307,7 +316,11 @@ def create_mosaic(filenames):
         else:
             diff_im = image.data
         data.append(diff_im)
-        detector.append(filename_parser(filename)['detector'].upper())
+        try:
+            detector.append(filename_parser(filename)['detector'].upper())
+        except KeyError:
+            # If filename_parser() does not recognize the file, skip it.
+            pass
```
york-stsci, line 331: Repeating that this needs to be logged.

york-stsci: Unless it's already being logged in the parser itself, of course.

jwql/jwql_monitors/generate_preview_images.py
```diff
@@ -650,10 +663,8 @@ def group_filenames(filenames):
         subgroup = []
 
         # Generate string to be matched with other filenames
-        try:
-            filename_dict = filename_parser(os.path.basename(filename))
-        except ValueError:
-            logging.warning('Could not parse filename for {}'.format(filename))
+        filename_dict = filename_parser(os.path.basename(filename))
+        if not filename_dict['recognized_filename']:
             break
```
york-stsci, line 678: Same consistency note here as above.

jwql/jwql_monitors/generate_preview_images.py
```diff
@@ -747,7 +763,8 @@ def process_program(program, overwrite):
         # Determine the save location
         try:
             identifier = 'jw{}'.format(filename_parser(filename)['program_id'])
-        except ValueError:
+        except KeyError:
+            # In this case, the filename_parser failed to recognize the filename
             identifier = os.path.basename(filename).split('.fits')[0]
```
york-stsci, line 783: Same consistency and logging notes.

jwql/jwql_monitors/monitor_filesystem.py
```diff
@@ -137,9 +138,10 @@ def gather_statistics(general_results_dict, instrument_results_dict):
...
 
                         # Parse out filename information
-                        try:
-                            filename_dict = filename_parser(filename)
-                        except ValueError:
+                        filename_dict = filename_parser(filename)
+                        if filename_dict['recognized_filename']:
+                            filename_type = filename_dict['filename_type']
+                        else:
                             break
```
york-stsci, line 147: Same logging note

jwql/utils/utils.py
```diff
@@ -596,11 +600,9 @@ def filename_parser(filename):
...
     # Raise error if unable to parse the filename
     except AttributeError:
-        jdox_url = 'https://jwst-docs.stsci.edu/understanding-jwst-data-files/jwst-data-file-naming-conventions'
-        raise ValueError(
-            'Provided file {} does not follow JWST naming conventions.  '
-            'See {} for further information.'.format(filename, jdox_url)
-        )
+        filename_dict = {'recognized_filename': False}
+        logging.exception((f'\nFile; {filename} was not recognized by filename_parser(). Update parser or '
+                           'constants.py if it should be recognized.\n'))
```
york-stsci, line 606: I think my worry here is where this log entry ends up. Are we set up so that calling logging.exception() will go to the log file for whatever monitor is running? If not, should we pass in the log file, pass in the logger instance itself that's already set up to use the file, or do the logging after the function returns?

@bhilbert4 have you run this to see if the log entries show up in the appropriate file?

york-stsci (commented):
There are a lot of redundant comments, but my basic point is that we should probably choose a specific way of getting output from the file parser, testing that output for validity, and reacting to it. I don't mind if we're doing try/except or if/then, but we should probably pick one and use it.

My second note is that we need to make sure that, when the parser fails, we log that failure in a way that's connected not only to the specific monitor that's running, but to the specific place within that monitor where the failure happened.

The combination of the above suggests to me that, annoying as it is, wrapping the call to `filename_parser()` in a try/except block, and logging the exception (along with what stage of the monitor it is, what file it was trying to parse, and how it got that file (i.e. from user input for a search, from a self-generated list, from some other source) is probably the cleanest way to make sure we have the necessary context.

jwql/utils/utils.py
bhilbert4: I haven't tried running it from a monitor. Or actually, from archive_database_update nor generate_preview_images, since those are the only two monitor-like things that call it. But I think the filename_parser does need to be flexible enough that it can be called from anywhere, including situations where there is no logger set up (e.g. on the fly). That might argue for doing any logging after the function returns.

york-stsci: If we want to pass in a logger, but still be robust, one way to do it is to change the function template to something like:

```
def filename_parser(filename, **kwargs):
    logger = kwargs.get('logger', logging.getLogger(__name__))

```

which would mean that we're guaranteed to have a valid logger whether we pass one in or not. That said, I feel as if it probably makes more sense to log errors after the function returns, because that way we have more context as to what was going on (e.g. a user search string, a new file from a periodic search, etc.) Which means more writing, and more effort to keep things consistent, but also hopefully guarantees that, when we do log, it ends up in the output log for whatever is running.

bhilbert4:
Ok @york-stsci, have a look. I updated the files to add logging statements in the calling functions. I kept the logging statement in filename_parser() itself. Maybe with testing we can see if that statement makes it into the expected log. Running locally, I can try a few of these cases tomorrow.

mfixstsci (approved):
Thanks @bhilbert4 assuming that you've addressed all of @york-stsci comments I am cool with merging this in.

jwql/jwql_monitors/generate_preview_images.py
bhilbert4: Just a note to confirm that in offline conversations, everyone is now happy with the logging statements in the edited code.

PR #1620, opened 2024-07-12
Update Bokeh `filter` Keyword in NRS TA Monitors

mfixstsci:
Most recent versions of Bokeh changed the `filters` argument to `filter` and no longer accepts a lists.


bhilbert4 (approved):
Nice catch!

PR #1622, opened 2024-07-18
filter thumbnails by filter/pupil/grating

### LINKED ISSUE #1573: Add ability to filter thumbnails by filter/pupil
_Issue [JWQL-202](https://jira.stsci.edu/browse/JWQL-202) was created on JIRA by [Bryan Hilbert](https://jira.stsci.edu/secure/ViewProfile.jspa?name=hilbert):_

From the JWQL User Survey. Brought up by NIRCam and NIRISS teams.

Allow thumbnails to be filtered (and sorted?) by filter/pupil value. 
Are there any other pieces of metadata we should think about? 
We can already filter on detector, viewed/unviewed, exposure type (e.g. NRC_IMAGE) and visit number. Perhaps subarray? A single observation will have only a single subarray, but you can get mixed results from the query page.

BradleySappington:
Created filters for "filter", "pupil", and "grating"
These filters will be populated for all results on the following pages
query results
observation archive pages

Default values will be populated ONLY with those relevant to what is available to be filtered.  This does mean that some dropdowns will only have one option, and in some cases, that option may be "empty".

I also aligned non-filtering dropdowns in a following horizontal rule section to avoid page real estate congestion.

https://jira.stsci.edu/browse/JWQL-202
closes #1573

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -19,3 +19,4 @@ jwql/website/apps/jwql/static/filesystem
 jwql/website/apps/jwql/static/outputs
 jwql/website/apps/jwql/static/preview_images
 jwql/website/apps/jwql/static/thumbnails
+jwql-current.yml
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -1,5 +1,46 @@
 ## What's Changed
 
+1.2.10 (2024-07-10)
+===================
+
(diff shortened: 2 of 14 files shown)


BradleySappington:
Screenshot 2024-07-18 at 12 50 42 PM

BradleySappington:
whoops, wrong base branch

PR #1628, opened 2024-08-16
Fix log cleanup script

bhilbert4:
Small tweaks to the script for cleaning up old log files.

Added a dry_run parameter so that the user can see what the tool would delete.
Also made the file executable so that we can call it from the crontab with the same syntax as the other jobs.
Also made the threshold time to determine which logs to delete a float. It was previously an integer, and in units of days. Changing to a float gives a little more flexibility.


bhilbert4:
@mfixstsci this is ready for review. Local testing all worked as expected.

mfixstsci (approved):
LGTM @bhilbert4

PR #1631, opened 2024-08-17
Fix missing data in readnoise monitor plots

bhilbert4:
Looks like the Bokeh plotting.circle function's API was tweaked recently. You now need to specify a radius value when calling circle(). If you don't, there is no default value that it falls back onto, and the data are not plotted.
https://github.com/bokeh/bokeh/issues/13899

This PR adds the radius keywords to the readnoise monitor plots.

### DIFF
--- a/jwql/website/apps/jwql/monitor_pages/monitor_readnoise_bokeh.py
+++ b/jwql/website/apps/jwql/monitor_pages/monitor_readnoise_bokeh.py
@@ -154,7 +154,15 @@ class ReadNoisePlotTab():
                                                    ("ngroups", "@ngroups"),
                                                    ("readnoise", "@readnoise")]))
 
-            amp_plot.circle(x='expstarts', y='readnoise', radius=0.01, radius_dimension='y', source=source)
+            min_rn = np.min(readnoise_vals)
+            max_rn = np.max(readnoise_vals)
+            delta_rn = max_rn - min_rn
+            plot_max = max_rn + 0.5 * delta_rn
+            plot_min = min_rn - 0.5 * delta_rn
+            circle_radius = 0.01 * (plot_max - plot_min)
+
+            amp_plot.circle(x='expstarts', y='readnoise', radius=circle_radius, radius_dimension='y', source=source,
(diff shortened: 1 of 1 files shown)


bhilbert4:
Everything looked ok in my manual testing using the production database table. I set the radius values to match the radii in some screenshots of the readnoise monitor before the plots became empty. It may be more robust to calculate the radius based on the max and min values of the measured readnoise, although it's hard to predict how much Bokeh will extend the axis around these max and min values.

@mfixstsci @bsunnquist this is ready for review. Running it on the test server after merging will give the best idea as to whether the radius values used here are acceptable.

bsunnquist:
Thanks @bhilbert4 for debugging this - should the trending plot be more like radius=0.0005? I found the following A1 readnoise trending plot from back in December and the y range is quite small:
Screenshot 2024-08-19 at 12 13 18 PM

bhilbert4:
This is why I was wondering about using a value based on the min and max data values. But in a previous version of the plot for NIRISS, the values ranged between 10.18 - 10.23, but the plot range was 9.9 to 10.4. In that case, the radii of the points was about 0.01.  So I was anticipating that with more data points, the plot range would expand. Maybe we should just set the plot range explicitly, and from that we can calculate a reasonable circle radius. 
Screenshot 2024-08-19 at 1 30 49 PM

bhilbert4:
@bsunnquist I changed the code to specify the plot range and base the circle radius on that. How does this look?
Do you have good readnoise data in your database tables that you can test locally with? I don't.

bsunnquist:
Ah yeah I hadn't thought of the other instruments. But yes, this new change looks good to me on the NIRCam data.

mfixstsci (approved):
Thanks @bhilbert4

mfixstsci:
I am going to check the WATA/MSATA monitor to see if this is the same issue there 🤔

bhilbert4:
It's not, from what I saw. Those plots use "scatter" rather than "circle"

mfixstsci:
@bhilbert4 I just checked, it looks like something to deal with the databases 😓

PR #1634, opened 2024-08-21
Fix readnoise monitor plot y range

bhilbert4:
Put the y_range keyword in the bokeh figure() call rather than figure.circle() call


bhilbert4:
@mfixstsci this should fix the readnoise monitor. Ready for review.

mfixstsci (approved):

PR #1635, opened 2024-08-22
Added logging configuration to config file, and use it when opening logging

york-stsci:
Note that the config.json files on dev/test/ops have already been updated, since adding the logging configuration to those files shouldn't be an issue because it won't affect anything until it's called.


mfixstsci (approved):
@york-stsci LGTM. Do you think we could tie this into the log viewer in the app? Have a drop down with the different levels that a user can select on to display in the application?

PR #1639, opened 2024-08-26
Filter stage 3 files from get_proposal_info and thumbnails_ajax

bhilbert4:
#1614 accidentally removed a check for level 3 filenames in the functions named above. These checks prevented level 3 files from being considered when getting filename lists and populating rootfileinfo.

This PR returns these checks.


mfixstsci (approved):
LGTM @bhilbert4 thank you for your work on this today!

PR #1642, opened 2024-08-28
Adds filtering of log files by date, log message type, and folder

york-stsci:
Adds entries to the log file dictionary that sort logs based on

- date
- folder (type)
- priority (highest log level recorded)

and adds javascript to hopefully auto-update the list of possible log files based on those filters.


bhilbert4 (approved):
This looks good to me. I'd like to see how it behaves on the test server. I don't see any Cross Reference errors, although my browser does at times list logs from the test server directory, even after deleting cookies. Those logs obviously can't be loaded, but I don't see this as a show stopper.

PR #1643, opened 2024-08-29
Add radius keyword to bokeh.figure.circle calls

bhilbert4:
Similar to the fix for the readnoise monitor plots in #1631, this PR adds the 'circle' keyword to other instances of bokeh.figure.circle() calls in the JWQL codebase. In some cases there was no keyword indicating glyph size. In other cases, 'radius' was used, which is now deprecated.

Local testing showed everything looks ok, with the exception of the bad pixel monitor and the EDB telemetry monitor, which failed due to bokeh version mismatches. (Actually, we may need to run the monitors to see updated plots in these cases, since the plots are not produced on the fly).


bhilbert4:
@mfixstsci coming back to this PR, it might be easiest to review and merge, and then check the results on the test server? The pages that are failing to load are failing because of a bokeh version mismatch between the html file and the installed version. I think we need to re-run the monitors themselves (rather than just the data retrieval/plotting functions) in order to regenerate the html file with the current version of bokeh before we'll be able to load the page.

mfixstsci:
> @mfixstsci coming back to this PR, it might be easiest to review and merge, and then check the results on the test server? The pages that are failing to load are failing because of a bokeh version mismatch between the html file and the installed version. I think we need to re-run the monitors themselves (rather than just the data retrieval/plotting functions) in order to regenerate the html file with the current version of bokeh before we'll be able to load the page.

Assuming the monitor will rewrite the bokeh figures, this idea should work. Worst case, we just delete the file on disk and regenerate it.

mfixstsci (approved):
Changes look good, thanks @bhilbert4

PR #1647, opened 2024-10-07
Remove bokeh templating code

bhilbert4:
Remove bokeh_templating code from the repository.
This code is no longer used to create figures anywhere in the code base. It also has not been maintained for quite a while.


bhilbert4:
@mfixstsci this is ready for review. There was only one place in JWQL where bokeh templating was still being imported, but it wasn't actually being used. So I removed the import, and removed the bokeh_templating code.

mfixstsci (approved):
This is great @bhilbert4! We have been meaning to do this for a while.

PR #1648, opened 2024-10-14
retrieve version from package

zacharyburnett:
attempt to fix issue in the build workflow where the `dunamai` version retrieval action wasn't able to find the `micromamba` environment (and so attempted to install `dunamai` to the system Python installation, which failed)

upon further reflection on this we don't even need to use `dunamai` at all, we can just get the version from `jwql.__version__`, no need to make this complicated


.github/workflows/build.yml
```diff
@@ -43,9 +43,8 @@ jobs:
       - run: pip install .
       - run: pip list
       - id: version
-        uses: mtkennerly/dunamai-action@v1
-        with:
-          args: --strict --pattern "default-unprefixed" --style semver
+        name: retrieve package version
+        run: echo version=$(python -c "import jwql; print(jwql.__version__)") >> $GITHUB_OUTPUT
```
zacharyburnett, line 47: we don't even have to use `dunamai` here, we can just use the pre-populated `__version__`

jwql/__init__.py
```diff
@@ -10,4 +12,4 @@
               f"while JWQL is using {__version__}")
 
 except FileNotFoundError:
-    print('Could not determine jwql config version')
+    warnings.warn('Could not determine jwql config version')
```
zacharyburnett, line 15: when this was `print(`, it was polluting the output of `import jwql; print(jwql.__version__)` in the workflow

mfixstsci (approved):
@zacharyburnett thanks for this, sorry for the delay.

PR #1651, opened 2024-10-16
Exclude source-specific WFSS files from observation page

bhilbert4:
WFSS source specific files have started showing up on JWQL's observation level pages. I'm not sure how this happened. The filenames come in from the MAST query. Maybe the metadata examined by MAST has changed?  This was leading to a lot of extra thumbnail placeholders being added to WFSS programs on the observation level pages. I don't think we want to keep track of these files, even though they are an output of the spec2 pipeline, and JWQL generally supports stage 2 outputs. In WFSS there is one file for each extracted source. By default the pipeline now keeps only the top 100 brightest sources per exposure. I think adding 100*n_exposures files to the page could quickly become overwhelming. Maybe if we come up with a better way to organize these, we can add them back in. They are 2D images of extracted spectra.

This is not tested yet.

### DIFF
--- a/jwql/website/apps/jwql/data_containers.py
+++ b/jwql/website/apps/jwql/data_containers.py
@@ -2144,7 +2144,7 @@ def thumbnails_ajax(inst, proposal, obs_num=None):
     ----------
     inst : str
         Name of JWST instrument
-    proposal : str (optional)
+    proposal : str
         Number of APT proposal to filter
     obs_num : str (optional)
         Observation number
@@ -2193,6 +2193,12 @@ def thumbnails_ajax(inst, proposal, obs_num=None):
     # Gather data for each rootname, and construct a list of all observations
     # in the proposal
     for rootname in rootnames:
+        # Skip over unsupported filenames
+        # e.g. jw02279-o001_s000... are spec2 products for WFSS with one file per source
+        # Any filename with a dash after the proposal number is either this spec2 product
+        # or a level 3 product
+        if f'jw{proposal}-' in rootname:
+            continue
 
         # Parse filename
         filename_dict = filename_parser(rootname)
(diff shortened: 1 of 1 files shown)


bhilbert4:
I think the (now commented out) lines that created a dictionary when the filename_parser doesn't recognize the filename doesn't make any sense to keep. It assumes a filename structure, except that structure is recognized by the filename_parser. So if the parser fails, this assumed dictionary probably won't contain correct information either.

bhilbert4:
@mfixstsci I think this is ready for review. The update at line 2200 solves the problem with the WFSS files that were getting through. Local testing looks good. There are a few MIRI proposals (e.g. 1219) where the observation level page is failing, but that looks like it's due to missing filter names in my local rootfileinfo db. We should certainly confirm this once on the test server.

With the fix at line 2200, we technically could leave the lines in place that create the dummy dictionary in cases where the filename parser fails. But this strikes me as asking for more trouble down the line. If the filename parser fails, there won't be rootfileinfo entries for the files, leading to more problems. Do you agree?

jwql/website/apps/jwql/data_containers.py
```diff
@@ -2193,27 +2193,22 @@ def thumbnails_ajax(inst, proposal, obs_num=None):
     # Gather data for each rootname, and construct a list of all observations
     # in the proposal
     for rootname in rootnames:
+        # Skip over unsupported filenames
```
mfixstsci, line 2196: Man, good catch here. I know some of these files have made their way into our filesystem/have thumbnails and preview images generated now. Is there a way to remove them now that we know this will filter them out in the future?

mfixstsci (approved):
This looks good to me @bhilbert4 I am going to do some testing and will let you know what I think.

jwql/website/apps/jwql/data_containers.py
bhilbert4: I haven't seen any thumbnails or preview images for these files. Only placeholder entries on the observation level pages. But it should be possible, although maybe a bit tedious, to find any existing thumbnails for these files and remove them.

mfixstsci: Right, that makes sense since we are getting 404s for these files that only placeholders exist.

PR #1658, opened 2024-11-25
[WIP] Jdaviz integration

york-stsci:
This adds Jdaviz integration to the archived exposure pages, including both imaging and spectroscopic data. Note that Jdaviz is currently not perfectly optimized, so there may be long loading times in some cases. At this point, it isn't necessarily clear whether that will improve with further optimization.


jwql/utils/constants.py
```diff
@@ -357,7 +357,7 @@
     "psfsub",
     "amiavg",
     "aminorm",
-    "ami",
+#    "ami",
```
bhilbert4: Might as well remove this completely.

jwql/utils/utils.py
```diff
@@ -381,6 +381,7 @@ def filename_parser(filename):
     ValueError
         When the provided file does not follow naming conventions
     """
+    logging.debug(f"Running filename_parser() on {filename}")
```
bhilbert4, line 384: This is going to generate a lot of logging statements. We call the filename parser in many different places. Keeping it at debug level should be ok though.

jwql/website/apps/jwql/data_containers.py
```diff
@@ -584,8 +586,9 @@ def get_available_suffixes(all_suffixes, return_untracked=True):
     # that specifies the order to use, add them to the end of the
     # suffixes list. Their order will be random since they are not in
     # EXPOSURE_PAGE_SUFFIX_ORDER.
-    if len(untracked_suffixes) > 0:
-        suffixes.extend(untracked_suffixes)
+    for suffix in untracked_suffixes:
```
bhilbert4, line 590: Good catch.

jwql/website/apps/jwql/data_containers.py
```diff
@@ -1401,17 +1406,20 @@ def get_image_info(file_root):
         os.path.join(FILESYSTEM_DIR, 'proprietary', proposal_dir,
                      observation_dir, '{}*.fits'.format(file_root))))
 
+    logging.info(f"Files before filtering: {filenames}")
```
bhilbert4: This logging statement along with the two others in this function will also potentially add a lot of lines to a typical log file. But I guess if they're going into the django log file, we ideally won't need to sift through them very often.

jwql/website/apps/jwql/static/js/jwql.js
```diff
@@ -130,20 +138,11 @@
          'https://mast.stsci.edu/api/v0.1/Download/file?uri=mast%3AJWST%2Fproduct%2F' +
          fits_filename + '.fits');
 
-    // Show the appropriate image
```
bhilbert4, line 133: My JS is pretty weak. But based on the testing we did, I'm assuming it's ok that these lines/variables are being removed.

jwql/website/apps/jwql/templates/view_image.html
```diff
@@ -149,23 +167,128 @@ <h5>Comments</h5>
             </div>
         </div>
 
+        <script>
+        function handle_change(type, file_root, inst) {
```
bhilbert4: Can you add some documentation here describing what this function does? Also, is there an advantage to having this function here rather than in jwql.js?

jwql/website/apps/jwql/templates/view_image.html
```diff
@@ -149,23 +167,128 @@ <h5>Comments</h5>
...
+            var get_string = "?suffix=" + type + "&preview=" + preview_type;
+            console.log("Get String: " + get_string);
+            
+            var file_root_list = {{ file_root_list | tojson | safe }};
+            var index = Number({{ index }});
+            document.getElementById("prev_button").href = '/' + inst + '/' + file_root_list[index - 1] + get_string;
+            document.getElementById("next_button").href = '/' + inst + '/' + file_root_list[index + 1] + get_string;
+        }
+        
+        function set_view(view) {
```
bhilbert4: Can you add some documentation here describing what this function does?

jwql/website/apps/jwql/templates/view_image.html
```diff
@@ -65,52 +65,70 @@ <h3>{{ file_root }}</h3>
...
         <p>
             <a id="view_header" class="btn btn-primary mx-2" role="button" href="#">View Header</a>
-            <a id="explore_image" class="btn btn-primary mx-2" role="button" href="#">Explore Image</a>
+            <a id="view_exposure" class="btn btn-primary mx-2" role="button" href="#">View Exposure</a>
             <!-- <a id="download_fits" class="btn btn-primary my-2 mx-2" role="button" href='{{ static("") }}preview_images/{{ file_root[:7] }}/{{ file_root }}' download>Download FITS</a>
             <a id="download_jpg" class="btn btn-primary my-2 mx-2" role="button" href='{{ static("") }}preview_images/{{ file_root[:7] }}/{{ jpg }}' download>Download JPEG</a> -->
         </p>
 
+    	<!-- Allow the user to change the preview format that is being displayed -->
+    	<a id="view_preview_type" data-current-preview="jpeg">View Preview As:</a>
```
bhilbert4, line 82: I like the idea that the user is able to set a preference via cookie.

jwql/website/apps/jwql/urls.py
```diff
@@ -45,6 +45,7 @@
 
 from django.urls import path
 from django.urls import re_path
+from django.views.generic import TemplateView
```
bhilbert4: Is this being used anywhere?

jwql/website/apps/jwql/views.py
```diff
@@ -57,12 +58,14 @@
 from django.core.paginator import Paginator
 from django.http import HttpResponse, JsonResponse
 from django.shortcuts import redirect, render
+from pathlib import Path
```
bhilbert4: Repeated import from above.

bhilbert4 (approved):
Looks good. Just a few mostly-trivial comments. We've done enough testing that I'm confident things are working as expected.

jwql/website/apps/jwql/data_containers.py
```diff
@@ -1647,6 +1656,34 @@ def get_preview_images_by_rootname(rootname):
...
+
+    Returns
+    -------
+    detector_list : list
+        A list of images that are part of the same exposure but with all detectors.
+    """
+    detector_list = []
+    search_rootname = rootname[:25]
+    filenames = get_filenames_by_rootname(search_rootname)
+    for filename in filesnames:
```
mfixstsci: Caught a typo here: should be `for filename in filenames:` instead of `filesnames`

jwql/website/apps/jwql/data_containers.py
```diff
@@ -1647,6 +1656,34 @@ def get_preview_images_by_rootname(rootname):
     return preview_images
 
 
+def get_detectors_by_rootname(rootname):
```
mfixstsci, line 1660: If you don't mind @york-stsci when you are able to make updates, could you write a test or two for this new function.

jwql/website/jwql_proj/settings.py
```diff
@@ -52,6 +56,7 @@
     'django.contrib.sessions',
     'django.contrib.messages',
     'django.contrib.staticfiles',
+    'revproxy.apps.RevProxyConfig'
```
mfixstsci: @york-stsci `revproxy` is giving me some issues locally when I try to run jwql/import methods from the website modules. It's originating from the django `setup()` call.

We should add `revproxy` to our dependency list, but even after adding revproxy I am getting a ModuleNotFoundError:
`ModuleNotFoundError: No module named 'revproxy.apps'`

jwql/utils/constants.py
york-stsci: Done.

jwql/utils/utils.py
york-stsci: That was my thought. I had the debug statements because I was trying to track down what the code flow was for deciding whether or not specific extensions would be displayed as options. My current thought is that, unless it causes problems for us, it can hopefully stay.

jwql/website/apps/jwql/data_containers.py
york-stsci: I've edited all of these into debug statements, because it's something we'll probably only want to turn on if we're trying to track down a problem.

jwql/website/apps/jwql/static/js/jwql.js
york-stsci: Yes, my memory is that the structural changes and the changes to the function above make those lines redundant, and because javascript (in a web context) fails silently (i.e. if there's an exception then it just stops running at that point), having code that's not doing anything, but could still fail, is to be avoided.

jwql/website/apps/jwql/templates/view_image.html
york-stsci: Done, and no that was a thing I was doing while actively writing the functions. I've moved them to jwql.js, and put the documentation there.

jwql/website/apps/jwql/templates/view_image.html
york-stsci: See above

jwql/website/apps/jwql/views.py
york-stsci: Removed duplicate import

jwql/website/apps/jwql/data_containers.py
york-stsci: Fixed

jwql/website/jwql_proj/settings.py
york-stsci: Oops. That's a way that I was trying to get the preview to load, and it's not being used at all anymore, but I forgot to delete it. I'm going to do my best to eliminate it entirely in my next update to this PR, but please check to make sure, because my own environment isn't seeing it because I actually installed revproxy.

york-stsci:
I've now pushed my first-pass response to the review.

jwql/website/jwql_proj/settings.py
mfixstsci: Thank you!

jwql/website/jwql_proj/urls.py
```diff
@@ -47,9 +47,13 @@
 """
 
 from django.contrib import admin
-from django.urls import include, path
+from django.urls import include, path, re_path
```
mfixstsci, line 50: Doesn't look like we use `re_path` in this module 😎

jwql/website/apps/jwql/urls.py
york-stsci: No, removed.

jwql/website/apps/jwql/data_containers.py
york-stsci: I've added one. I parametrized it so that we can add additional test cases, but even the basic one should be okay as a quick test.

mfixstsci (approved):
Did virtual code review and all tasks are marked as completed. Here we go!

PR #1662, opened 2024-12-18
Switch URL for prog info scraping to use the OPO site

### LINKED ISSUE #1660: Update proposal information page URL
_Issue [JWQL-260](https://jira.stsci.edu/browse/JWQL-260) was created on JIRA by [Bryan Hilbert](https://jira.stsci.edu/secure/ViewProfile.jspa?name=hilbert):_

The proposal information webpage that JWQL uses to get proposal info (in text_scrape in data_containers.py) went down last week. ITSD wasn't able to get the server back up and working, and so they updated the URL for that page. 

So now, when JWQL tries to load any Observation level page (e.g. <https://jwql.stsci.edu/NIRSpec/archive/3073/obs005/)> everything crashes because it's not able to retrieve any program info.

Update the URL that JWQL scrapes to get the program info. The new URL is <https://www.stsci.edu/jwst/science-execution/program-information>

Another option may be <https://www.stsci.edu/jwst-program-info/program/?program=3073> which I think is the OPO version of the program information page. This might be an easier URL to work with than the other one above.

bhilbert4:
The site that JWQL had been using to scrape for proposal information went down last week, and is not yet back up. JWQL did not check for the internal server error returned for those pages, which was causing JWQL to crash when trying to load any observation-level page. 

This PR switches over to use the OPO version of the proposal information pages. It also adds a check for an internal server error when scraping the page.

Resolves #1660

### DIFF
--- a/jwql/website/apps/jwql/data_containers.py
+++ b/jwql/website/apps/jwql/data_containers.py
@@ -2057,12 +2057,15 @@ def text_scrape(prop_id):
     program_meta : dict
         Dictionary containing information about program
     """
+    # Ensure prop_id is a 5-digit string
+    prop_id = str(prop_id).zfill(5)
 
     # Generate url
-    url = 'http://www.stsci.edu/cgi-bin/get-proposal-info?id=' + str(prop_id) + '&submit=Go&observatory=JWST'
+    url = f'https://www.stsci.edu/jwst-program-info/program/?program={prop_id}'
     html = BeautifulSoup(requests.get(url).text, 'lxml')
     not_available = "not available via this interface" in html.text
     not_available |= "temporarily unable" in html.text
+    not_available |= "internal error" in html.text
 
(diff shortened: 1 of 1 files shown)


bhilbert4:
@mfixstsci, local testing shows that this is working well, so I think it's ready for review.

mfixstsci (approved):

mfixstsci:
@bhilbert4 this is a good fix, thank you for submitting. I wonder if there is a more reliable API to use to obtain this same information instead of scraping websites 🤔

bhilbert4:
I started looking into getting the info from pyvo or mast, but there wasn't anything obvious in my quick check, so I moved on. I agree that if we can get the info that way, it would be better. I opened a JIRA ticket for that last night.

sosey:
add me to the ticket, I have a likely internal method, depends on what you want

PR #1666, opened 2025-01-13
Add wisp finder monitor

bhilbert4:
Add an automated wisp finder monitor to JWQL. This tool finds new data, creates a scaled image of it, and uses a neural network to predict whether there is a wisp in the image. If a wisp is found, it sets the wisp flag in the Anomalies instance for the appropriate RootFileInfo instance.

This model is trained and currently works only for the wisps on NIRCam's B4 detector.


bhilbert4:
Here's a good test case for both test and ops servers:
python wisp_finder.py -s 59964 -e 59965
This should find a handful of PID 1837 obs 25 files, where some have wisps and some don't, and none are currently flagged in the anomaly database.

mfixstsci (approved):
@bhilbert4 this is awesome! I am approving, with a few tasks:

- [x] Add the django model
- [x] Updates to the default config file.

bhilbert4:
Django model and migration file are both present in the PR now.

mfixstsci (approved):

bhilbert4:
I've updated the config files on all three servers to include the new entry.

PR #1667, opened 2025-01-15
Remove support for preview image and thumbnail listfiles

### LINKED ISSUE #1029: Remove preview image and thumbnail inventory files
These files are difficult to work with, and don't seem to offer much value. They were originally implemented as a time-saving feature (over using glob), but the amount of time saved as the number of files becomes very large isn't clear, and any extra time might be worth it in return for easier maintenance.

bhilbert4:
Resolves #1029 

Before launch, we implemented a set of text files that contained lists of preview images and thumbnails for each instrument. Originally these were designed to speed up calls to generate_preview_images, in that it could check the text file to see if a preview image already existed, and skip the creation of a new preview image if so.

But after launch, once the data volume increased enough, this method was no longer saving any time. So we removed the functionality that had the preview image generator open and read in the list files. However, we left in place the code that was adding names of new preview images to the list files at the end of each run of generate_preview_images.py. So the files continued to grow, but were not being used for anything. For NIRCam on production, the listfile is now over 400MB. 

This PR removes all support for the listfiles. The files are no longer opened and have new entries added. This should speed up generate_preview_image.


bhilbert4:
@mfixstsci this is ready for review

mfixstsci (approved):
Thanks for the history lesson and the updates @bhilbert4 -- this was added before my start on the project.

PR #1668, opened 2025-01-16
Update data_containers and bokeh_dashboard to use Django models

### LINKED ISSUE #1486: Update data_containers to use django
_Issue [JWQL-184](https://jira.stsci.edu/browse/JWQL-184) was created on JIRA by [Brian York](https://jira.stsci.edu/secure/ViewProfile.jspa?name=york):_

jwql.website.apps.jwql.data_containers.py

bhilbert4:
Resolves #1486 

Replace sqlalchemy queries with Django model queries in several functions/classes.

In order to make build_tables() work with an arbitrary model, I had to make adjustments so that monitor models are imported at the time of app initialization. This wasn't happening previously because those models are not defined in models.py. By adding the apps.py file with the imports, and then adding a reference to that into settings.py, the monitor models are now imported and available at app initialization. 

Local testing showed everything working.


bhilbert4:
@mfixstsci this is also ready for review

mfixstsci (approved):
@bhilbert4 thanks for the updates! Looks great

PR #1686, opened 2025-03-31
Fix scaling in WATA and MIRI TA preview images

### LINKED ISSUE #1587: Better preview images for WATA files
_Issue [JWQL-216](https://jira.stsci.edu/browse/JWQL-216) was created on JIRA by [Bryan Hilbert](https://jira.stsci.edu/secure/ViewProfile.jspa?name=hilbert):_

Originally reported by [Melanie Clarke](https://jira.stsci.edu/secure/ViewProfile.jspa?name=mclarke) in [JWQL-1302](https://github.com/spacetelescope/jwql/issues/1302), this looks like another issue that the Jira-github bot did not mirror onto Jira.

NIRSpec WATA files have an unusual raw data format, which makes the rate files produced by the stage 1 pipeline useless to look at. The TA source can sometimes be seen in the preview image for the uncal files, but sometimes it has to be checked in the explore image link, by differencing the uncal group 0-1 and 1-2. This makes reviewing the WATA images inefficient, when usually all that's needed is a glance to determine that the TA image looks okay.

It would be nice to generate a better default preview image for WATA files, so that they can be reviewed at a glance.

bhilbert4:
Adjust scaling when creating preview images for NIRSpec WATA and MIRI non-full-frame TA images. 

Trial and error with historical data shows that the WATA images look better when linearly scaled (rather than log). For MIRI non-full-frame TA images, it turns out that all the pixels in the subarrays were flagged as non-science. This meant there were no pixels left to use for scaling, and so scaling was falling back to a default scaling of 0 to 1. This fix instead checks for the case where all pixels are flagged as non-science, and when that happens, it declares all pixels as good for the purposes of scaling. This results in images that look decent, especially compared to the solid yellow squares we had previously.

I've also added an if statement to call the AutoMinorLocator function only when the scaling is linear. This should suppress all of the warnings that AutoMinorLocator does not work with log scaling.

### DIFF
--- a/jwql/jwql_monitors/generate_preview_images.py
+++ b/jwql/jwql_monitors/generate_preview_images.py
@@ -800,7 +800,7 @@ def process_program(program, overwrite):
         try:
             im = PreviewImage(filename, "SCI")
             im.clip_percent = 0.01
-            im.scaling = 'log'
+            im.set_scaling()
             im.cmap = 'viridis'
             im.output_format = 'jpg'
             im.preview_output_directory = preview_output_directory
--- a/jwql/utils/preview_image.py
+++ b/jwql/utils/preview_image.py
@@ -37,7 +37,6 @@ Use:
 import logging
 import os
 import socket
-import warnings
 
 from astropy.io import fits
 import numpy as np
@@ -672,7 +671,6 @@ class PreviewImage():
             True if saving a thumbnail image, false for the full
             preview image.
         """
-        warnings.filterwarnings("ignore", message="AutoMinorLocator does not work with logarithmic scale")
         plt.savefig(fname, bbox_inches='tight', pad_inches=0)
         permissions.set_permissions(fname)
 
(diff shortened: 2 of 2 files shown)


bhilbert4:
At the moment these changes are looking good. Just need to test on a greater diversity of subarray shapes and sizes.

bhilbert4:
Fixed the colorbar tick marks for the linear scaling case, and tested on a few other types of files. I'm happy with this now. I think it's ready for review, merge, and then a more comprehensive test on the test server. @mfixstsci

mfixstsci (approved):
Looks good to me @bhilbert4 the inline comments made it very easy to follow. Since we are dealing with preview images here, will they need to be regenerated with the new scaling?

bhilbert4:
> Looks good to me @bhilbert4 the inline comments made it very easy to follow. Since we are dealing with preview images here, will they need to be regenerated with the new scaling?

Yep. A manual run on WATA data is needed, or I think we can simply delete the existing preview images once this is released, and the next run of the preview image generator should re-create them.

PR #1690, opened 2025-07-02
Hotfix for changed pyVO URL

york-stsci:
This changes the pyVO URL to match the new requirement. This is required both by at least two monitors and by the website as a whole.


mfixstsci (approved):
Thanks @york-stsci! Looks good

PR #1691, opened 2025-07-15
Hotfix for changed pyVO URL

york-stsci:
Merging the hot fix into develop.


york-stsci:
This hot fix changes the pyvo URL, because STScI changed the service that they were using, and thus we need to access the new URL. As a separate change, we may want to move this URL into constants.py and have all the places that use it use that constant, rather than defining it as a URL in four different functions.

mfixstsci (approved):
LGTM

PR #1692, opened 2025-07-18
Aliveness Test Script

york-stsci:
This change

- Creates a "URLS" constant in constants.py
- Adds to that the URLs that are called by active code (i.e. not URLs that are only accessible in links from webpages)
- Changes existing places that use those URLs (except for MAST, at the moment) to use the entry in constants.py
- Adds a try/except block to constants.py accessing the STScI schema pages to load instrument schemas, because otherwise if that page is down then the aliveness test won't be able to import constants.py (because it will fail to access that URL) and then won't be able to run its check to tell us that the URL is down.
- Added an aliveness test script that goes through the URLS constant and attempts to access every site therein

### DIFF
--- a/jwql/utils/constants.py
+++ b/jwql/utils/constants.py
@@ -32,6 +32,16 @@ import asdf
 import inflection
 import os
 
+# URLs
+URLS = {
+    "CRDS": "https://jwst-crds.stsci.edu",
+    "DATAMODEL_SCHEMA": "http://stsci.edu/schemas/jwst_datamodel/subarray.schema",
+    "STSCI_JWST_PHASE2": "https://www.stsci.edu/jwst/phase2-public",
+    "STSCI_JWST_PROGRAM_INFO": "https://www.stsci.edu/jwst-program-info",
+    "STSCI_MAST": "https://mast.stsci.edu/",
+    "STSCI_VO": "https://mast.stsci.edu/vo-tap/api/v0.1/caom",
+}
+
 # Each amplifier is represented by 2 tuples, the first for x coordinates
 # and the second for y coordinates. Within each tuple are value for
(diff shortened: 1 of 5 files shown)


jwql/utils/constants.py
```diff
@@ -32,6 +32,16 @@
 import inflection
 import os
 
+# URLs
+URLS = {
+    "CRDS": "https://jwst-crds.stsci.edu",
+    "DATAMODEL_SCHEMA": "http://stsci.edu/schemas/jwst_datamodel/subarray.schema",
+    "STSCI_JWST_PHASE2": "https://www.stsci.edu/jwst/phase2-public",
```
bhilbert4, line 39: Looks like this URL is used in data_containers.py. Would be good to switch that hardcoded URL to this entry from constants.py

jwql/utils/constants.py
```diff
@@ -32,6 +32,16 @@
...
+    "STSCI_JWST_PROGRAM_INFO": "https://www.stsci.edu/jwst-program-info",
```
bhilbert4, line 40: Looks like this URL is used in data_containers.py (line 2089). Would be good to switch that hardcoded URL to this entry from constants.py

jwql/utils/constants.py
```diff
@@ -32,6 +32,16 @@
...
+    "STSCI_MAST": "https://mast.stsci.edu/",
```
bhilbert4, line 41: Looks like this URL is used in utils.py (line 320) and in claw_monitor.py. Would be good to switch those hardcoded URLs to this entry from constants.py

bhilbert4 (approved):
Looks good to me. There are a couple hardcoded instances of the URLs in the repo that can be replaced with the new variables from constants.py

PR #1693, opened 2025-07-24
Add level3 filename parser support

bhilbert4:
First steps towards support for level 3 files.


mfixstsci:
@bhilbert4 nice, I am excited about this one.

bhilbert4:
This PR allows the filename_parser to parse level 3 files, and allows archive_database_update to create RootFileInfo instances for those level 3 files.

jwql/tests/test_utils.py
```diff
@@ -225,7 +247,86 @@
...
+         'suffix': 's3d',
+         'recognized_filename': True,
+         'filename_type': 'stage_3_nirspec_ifu_mos',
+         'file_root': 'jw09230-o004_t001_nirspec_g395h_f290lp',
+         'detector': 'Unknown',
+         'group_root': 'jw09230-o004_t001_nirspec_g395h_f290lp',
+         'observation': '004'}),
+
+    # Test stag3, nirspec source based
+    ('jw04735-o005_s000013418_nirspec_f100lp-g140h_x1d.fits',
```
mfixstsci, line 302: @bhilbert4 as always, thank you for writing tests. So each part of this filename has an associated "key/value" pair that the file parser can recognize and can construct a valid level 3 filename?

One question I have, I see this has a specific source ID, I am assuming this x1d is for a MSA or IFU exposure. Will the source id's always correspond to the same source when the pipeline generates all of these individual x1d images? Will source `s000013418` be the same source in this version of the pipeline, older and newer?

jwql/utils/utils.py
```diff
@@ -458,6 +479,59 @@ def filename_parser(filename):
         r"_(?P<instrument>(nircam|niriss|nirspec|miri|fgs))"\
         r"_(?P<optical_elements>((?!_)[\w-])+)"
 
+    # Stage 3 WFSS filenames
+    # e.g. jw01076-o124_s000001220_nircam_f322w2-grismr_x1d.fits cal.fits c1d.fits
+    #This should not be used anymore...
```
mfixstsci, line 484: As in, these files are no longer produced by the pipeline?

jwql/utils/utils.py
```diff
@@ -458,6 +479,59 @@ def filename_parser(filename):
...
+    stage_3_wfss_source_id = \
```
mfixstsci, line 485: Wow -- this is impressive. Lot's of regex here!

jwql/website/apps/jwql/archive_database_update.py
```diff
@@ -124,25 +126,22 @@ def get_updates(update_database):
...
+            all_filenames = []
+            indices = []
 
-            # Filter source-based level 2 files out of the rootnames and filenames
-            all_rootnames = filter_rootnames(all_rootnames)
-            filenames = filter_filenames(filenames, all_rootnames)
+            for i, item in enumerate(filenames):
+                rname = '_'.join(item.split('/')[-1].split('_')[:-1])
+                if rname not in seen:
+                    seen.add(rname)
```
mfixstsci, line 138: This is cool, overriding the `set()` method to make it make sense for our context.

jwql/website/apps/jwql/archive_database_update.py
```diff
@@ -266,11 +267,11 @@ def files_in_filesystem(files, permission_type):
             full_filepath = os.path.join(FILESYSTEM, permission_type, relative_filepath)
             filenames.append(full_filepath)
         except ValueError:
-            print('Unable to determine filepath for {}'.format(filename))
+            logging.warning('Unable to determine filepath for {}'.format(filename))
     return filenames
 
 
-def update_database_table(update, instrument, prop, obs, thumbnail, obsfiles, types, startdate, enddate, proposal_category):
+def update_database_table(update, instrument, prop, obs, thumbnail, obsfiles, types, startdate, enddate, proposal_category, obsfilenames=[]):
```
mfixstsci, line 274: Please update the docstring here as well 😎

mfixstsci (approved):
Hey @bhilbert4 this looks great. Thank you for your work on this and I have a couple of questions here.

* How does this look on your local version of JWQL, were you able to update the dbs and then show level3 products in the app?
* Outside of all the regex for the filename parser, I see we add some new functions and removed some older ones. Does adding the level 3 files make the solutions/algorithm to detect and add new files easier over all?

jwql/tests/test_utils.py
bhilbert4: My experience with WFSS says no, the source number may change from one pipeline reduction to the next. In that case, the source number comes, I think, from the source ID assigned in the source_catalog step. So the numbers will depend on source finding parameters. I'm not sure if that's the same for MSA or IFU though. 

In this case, the filename_parser is looking for the -o and then the s, followed by 9 digits

jwql/utils/utils.py
bhilbert4: Exactly. They will be present in MAST until the new file naming scheme is in place and reprocessing is done and, I'm assuming, the MAST folks will have to delete them manually. But, once the initial reprocessing is done with the new file naming scheme, all the info will be available in the new files and we won't need to think about the old ones.

jwql/utils/utils.py
bhilbert4: So many file naming options to cover....

jwql/tests/test_utils.py
mfixstsci: Cool, that's what's important. As long as the test isn't dependent on the value of the source, but that the source ID is there.

jwql/utils/utils.py
mfixstsci: Cool, we will keep the comment when this day eventually comes.

jwql/website/apps/jwql/archive_database_update.py
bhilbert4: "seen" is just an empty set to start with. Then entries are added to complete a set of filenames

jwql/website/apps/jwql/archive_database_update.py
bhilbert4: Done.

mfixstsci (approved):

PR #1700, opened 2025-09-30
wisp finder - nans in png, close figures

bhilbert4:
Some fixes to the wisp_finder to speed up execution and make the prediction more accurate.

NaNs in the data were being transformed into black boxes that covered much of the field of view. This PR introduces an interpolation scheme to overwrite NaN pixels.

Also, the matplotlib figures are now explicitly closed, so as not to continue to take up memory after they are used in the prediction.


york-stsci (approved):
Looks good!

york-stsci (approved):
Nice

mfixstsci (approved):

PR #1703, opened 2025-10-07
Fix < sign in log statement of wisp finder

bhilbert4:
Correct the logging statement that compares the wisp probability and the threshold probability when a wisp is found.


york-stsci (approved):
LGTM

PR #1708, opened 2025-10-23
Fix Row for NRC Exposure Level View

mfixstsci:
Having an extra id in this tag was creating artificial space for a fall back image. Typically, we show 4 rows of NRC exposures per row, this extra id created a blank space for the fallback image with display set to `none`. This fixes the issue and spans the images to their proper size.


bhilbert4 (approved):
Good catch! Subtle error.

PR #1710, opened 2025-11-12
Fix new calendar widget for default date range

bhilbert4:
In the new calendar widget on the query page, the default dates were appearing on the webpage, but were not being populated into the actual variable used by the widget. Once a user changed the dates, the variable was being populated. So in cases where the user kept the default date range, the widget was failing with Null dates.

This PR updates the widget so that the variable does contain the default date range upon page load. This way if the user does not change the dates, they are still defined.

We also simplify the code a bit by setting the calendar widget to use 'to' as delimiter between the starting and ending dates, rather than manually switching to a dash.


mfixstsci (approved):
Nice -- LGTM! I tested this on the test server and they worked great.

PR #1731, opened 2025-12-15
Dependency Updates

mfixstsci:
updating upper and lower pins to bump jwst pipeline minor version up.
* crds
* asdf


pyproject.toml
```diff
@@ -19,32 +19,33 @@ classifiers = [
     "Programming Language :: Python",
 ]
 dependencies = [
-    "asdf>=3.1.0,<4",
-    "astropy>=6.0,<7",
-    "astroquery>=0.4.7,<0.5.0",
+    "asdf>=5.0.0,<6",
```
zacharyburnett: why bump the lower pin to `>=5.0.0`? That will severely limit the pipeline versions you'll be able to use. Unless `jwql` is now using features of `asdf` that were only introduced in ASDF 5, in which case ignore this comment

mfixstsci: I can bump it back!

bhilbert4 (approved):
LGTM!

PR #1732, opened 2025-12-15
Fix nirspec fixedslit filename parsing.

bhilbert4:
Nirspec fixed slit files had more options for filenaming than I realized earlier. This update allows the filename_parser to recognize all of them. 

Also updated the service used to connect to the engineering database so that it goes through MAST. 

Finally, in preview_image.py, return the level 2 preview image generation back to where it was before the level 3 work started. We can update the level 2 preview_image files to follow the same logic as the level 3 i2d files in a future PR.

The changes in this branch are currently being tested via a generate_preview_images.py run on the test server.


mfixstsci (approved):
LGTM! Thanks for catching this!

PR #1734, opened 2025-12-16
test on Python 3.12, 3.13, and 3.14

zacharyburnett:
update built and tested Python versions to `3.12`, `3.13,` and `3.14`


mfixstsci (commented):
Looks like there is test that is failing because of the regular expression library.

mfixstsci (approved):

PR #1747, opened 2026-02-02
Fix filename_parser and null read_patt_num

bhilbert4:
Fixes for the filename_parser and archive_database_update that were revealed through testing with level 3 files.

Make one of the filename_parser regular expressions more specific, in order to prevent matching with files it shouldn't.
Also, in some MAST queries for specific filenames, the results come back with patt_num set to None. However, read_patt_num in the RootFileInfo instances is non-nullable. Add a check for this, and set read_patt_num to 1 in cases where it is None in MAST.


mfixstsci (approved):
Looks good to me @bhilbert4. I've never heard of the `numbers` library before this so I learned something too, thanks!

PR #1750, opened 2026-02-18
Fixes column name change in STScI VO

york-stsci:
Fixes a votable column name change from "observationID" to "observationid"


mfixstsci (approved):
All tests are passing, thank you @york-stsci this looks great!

PR #1753, opened 2026-02-20
Add Level 3 Products to Web Application

mfixstsci:
This PR covers work that will include thumbnails and pngs of level three products and display them in the web application.


jwql/website/apps/jwql/data_containers.py
```diff
@@ -2022,7 +2035,7 @@ def get_thumbnail_by_rootname(rootname):
     thumbnail_basename = 'none'
 
     if len(thumbnails) > 0:
-        preferred = [thumb for thumb in thumbnails if 'rate' in thumb]
+        preferred = [thumb for thumb in thumbnails]
```
bhilbert4, line 2129: Playing with this today before I realized you had already looked into it. Below is what I came up with to try and make sure we use rate files for level 2 thumbnails, and "more interesting" thumbnails for the level 3 files. Since the list of thumbnail files is sorted, for level 2 rootnames, what you have should work, because it'll find the cal files first, which should work just as well as rate files. But for level 3, it would find cal or crf first, which I guess would be ok, but it wouldn't highlight that the file is a level 3 file. I know what's below is ugly, but with something along these lines, we could be sure to use a thumbnail from an x1d if it exists. And the nice thing here is that this is totally independent of what suffixes we will eventually choose to make thumbnails for, so maintenance in the future should be zero.

    if len(thumbnails) > 0:
        preferred = [thumb for thumb in thumbnails if 'rate' in thumb]
        if len(preferred) == 0:
            preferred = [thumb for thumb in thumbnails if 'dark' in thumb]
        if len(preferred) == 0:
            level3_suffixes = ['x1d', 'x1dints', 'phot', 'whtlt', 's2d', 's3d', 'i2d', 'crf', 'cal', 'psfsub', 'psfstack', 'ami-oi', 'amimulti-oi', 'aminorm-oi']
            for suffix in level3_suffixes:
                preferred = [thumb for thumb in thumbnails if suffix in thumb]
                print(suffix, preferred)
                if len(preferred) > 0:
                    break
        if len(preferred) > 0:
            thumbnail_basename = os.path.basename(preferred[0])

jwql/website/apps/jwql/data_containers.py
```diff
@@ -1283,14 +1283,27 @@ def get_filesystem_filenames(proposal=None, rootname=None,
...
-            os.path.join(FILESYSTEM_DIR, 'proprietary', proposal_dir,
-                         observation_dir, '{}*'.format(rootname))))
+        # Check for level 3 rootname
+        level_3_pattern = r'jw\d{5}-[a-zA-Z]\d{3}'
+        if re.search(level_3_pattern, rootname):
+            # Get level 3 rootname data
+            proposal_dir, ac_id = rootname.split('_')[0].split("-")
+            filenames = glob.glob(
+                os.path.join(FILESYSTEM_DIR, 'public', proposal_dir, "L3",
+                             "t", ac_id,'{}*'.format(rootname)))
```
bhilbert4: This line won't find all the L3 files, because there are other subdir options other than the current "t". There can also be "s". I ended up making this same fix in the PR for the exposure level pages, although not as cleverly as you have here with the regex.

jwql/website/apps/jwql/data_containers.py
```diff
@@ -1283,14 +1283,27 @@ def get_filesystem_filenames(proposal=None, rootname=None,
...
+            filenames.extend(glob.glob(
+                os.path.join(FILESYSTEM_DIR, 'proprietary', proposal_dir, "L3",
+                             "t", ac_id, '{}*'.format(rootname))))
```
bhilbert4: This line won't find all the L3 files, because there are other subdir options other than the current "t". There can also be "s". I ended up making this same fix in the PR for the exposure level pages, although not as cleverly as you have here with the regex.

jwql/website/apps/jwql/data_containers.py
```diff
@@ -2445,18 +2445,18 @@ def thumbnails_ajax(inst, proposal, obs_num=None):
         # e.g. jw02279-o001_s000... are spec2 products for WFSS with one file per source
         # Any filename with a dash after the proposal number is either this spec2 product
         # or a level 3 product
-        if f'jw{proposal}-' in rootname:
-            continue
+        # if f'jw{proposal}-' in rootname:
```
bhilbert4, line 2448: We need to add back in here something that will filter out the source-based files. e.g. jw05893-o014_s000005759_... These files are no longer supported, and just haven't been pulled out of MAST yet. There are TONS of them though, so we don't want to deal with them because we'd end up with hundreds or thousands of these files for each observation. The substring to look for is the "s" followed by a 9 digit number. e.g. "s000001234"

PR #1759, opened 2026-03-26
Show level 3 files on exposure and detector level pages

bhilbert4:
Tweak the code so that we can create and serve exposure-level pages that show level 3 files. For the moment these changes are intended only to create separate exposure-level pages for level 3 files, rather than adding level 3 suffixes to the exposure-level pages for level 2 files. While we may end up going that route in the future, that will take more work since we will have to map level 2 rootnames to associated level 3 rootnames.


bhilbert4:
Looking good so far. I think the exposure and detector level pages are working.

bhilbert4:
@mfix this one is also ready for review. The branch is also available on the test server. All the pages I've checked on there have appeared as expected, including jdaviz integration.

mfixstsci (approved):
Thanks @bhilbert4 this looks good!

PR #1760, opened 2026-04-01
Make only 1 thumbnail per level3 rootname

bhilbert4:
We're currently making thumbnail images for every suffix of every level 3 rootname. But we only need one thumbail per rootname. This PR enables that.


bhilbert4:
@mfixstsci this is ready for review. Branch is available on the test server. My checks show that it seems to be working well.

mfixstsci (approved):
This will be huge for my level 2 testing!

PR #1766, opened 2026-04-24
Turn off bad pixel search in Dark Monitor

bhilbert4:
With the dark monitor running on each individual full frame dark exposure, the SNR of the dark rate images is too low for a bad pixel search to be useful. In fact, the search finds that ~50% of the pixels are dead or hot. This causes the database interactions to take hours, and does not produce any useful information. So this PR turns off the bad pixel search. Lines are commented out rather than removed, in case we decide to have a more sophisticated method to determine whether to run the bad pixel check in the future.

I also shortened the list of NIRCam subarrays that the monitor will run on.


bhilbert4:
@mfixstsci this is ready for review. I ran it on the test server (on nircam only) and everything looks good.

mfixstsci (approved):
LGTM

PR #1768, opened 2026-04-28
Add code to check and remove stale lock files

### LINKED ISSUE #1767: Create script to check job status using lock files
_Issue [JWQL-308](https://jira.stsci.edu/browse/JWQL-308) was created on JIRA by [Bryan Hilbert](https://jira.stsci.edu/secure/ViewProfile.jspa?name=hilbert):_

There are times where for whatever reason, we get a lock file from one of the JWQL processes that remains in the directory if the process was ended in some non-nominal way (e.g. server reboot/update/etc). In this case, the lock file for the stale process then prevents any more instances of that process from starting. This can prevent any new data from being ingested.

The lock files contain the process id. For example, the archive_database_update instance that just started created a lock file with the contents: "`{}Process Id = 2522030. {`}So we should be able to create a script that will check for lock files, gather the process ID from those files, and check if the process is still running or not. If it is not, then the lock file should be deleted.

So for the job, we can run cron with:

`find [path] -type f -mtime +0 -exec [SCRIPT] \;`

where the last bit tells find to call the script for every file found. [path] can be "*.lock" and just run it on the home directory

bhilbert4:
Resolves #1767 

Sometimes when a process (like archive_database_update) is interrupted, the lock file is not deleted. When this happens, the script cannot run again until the lock file is manually deleted. If the lock file isn't noticed, this can lead to a long pause in running our periodic scripts. This PR adds code that will search for lock files and for any that are more than 24 hours old, it will check to see if the referenced process ID is still running. If it is not, the lock file is deleted. By running this via cron job, we shouldn't have to worry about stale lock files preventing code execution.

### DIFF
--- a/jwql/cleanup_locks.sh
+++ b/jwql/cleanup_locks.sh
@@ -1,0 +1,44 @@
+#!/usr/bin/env bash
+# cleanup_locks.sh - Find all .lock files in a directory tree and remove stale ones.
+#
+# Usage: ./cleanup_locks.sh [search_path]
+#
+# Defaults:
+#   search_path : directory containing this script
+
+set -euo pipefail
+
+SCRIPT_DIR="$(cd "$(dirname "${BASH_SOURCE[0]}")" && pwd)"
+SEARCH_PATH="${1:-$SCRIPT_DIR}"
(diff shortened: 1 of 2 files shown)


york-stsci (approved):
Looks like a simple script

PR #1769, opened 2026-05-07
Add level 3 file to Observation page

bhilbert4:
Began from the changes in the add-level3-radio-button branch from @mfixstsci, but I was unable to push my changes to his branch for some reason. So I'm trying as a new branch in my fork.

Adds level 3 file thumbnails to the observation-level pages.


jwql/jwql_monitors/generate_preview_images.py
york-stsci, line 202: Should file_type be given an entry here?

jwql/utils/utils.py
```diff
@@ -872,7 +872,7 @@ def get_rootnames_for_instrument_proposal(instrument, proposal):
         List of rootnames for the given instrument and proposal number
     """
     tap_service = vo.dal.TAPService(STSCI_VO_URL)
-    tap_results = tap_service.search(f"select observationID from dbo.CaomObservation where collection='JWST' and insName like '{instrument.lower()}%' and prpID='{int(proposal)}'")
+    tap_results = tap_service.search(f"select observationID from dbo.CaomObservation where collection='JWST' and insName like '{instrument.lower()}%' and prpID='{int(proposal)}'", maxrec=100000)
```
york-stsci, line 875: Is maxrec now a required field, or are we expecting to exceed this? Either way, is there a reason to choose this particular value?

jwql/website/apps/jwql/static/js/jwql.js
```diff
@@ -1589,63 +1637,99 @@ function update_thumbnail_array_all_obs(data) {
...
+                content += '<div class="thumbnail-info">';
+
+                if (stage == 'stage_3') {
+                    content += `Stage 3 <br>`;
+                }
+
+                content += `Proposal: ${filename_dict.program_id} <br>`;
+                content += `Observation: ${filename_dict.observation} <br>`;
+
+                if (stage == 'stage_2') {
```
york-stsci, line 1703: Is there a reason this isn't an if/else? Are there options other than stage 2/stage 3?

jwql/website/apps/jwql/data_containers.py
```diff
@@ -1957,7 +1957,7 @@ def get_rootnames_for_proposal(proposal):
     """
     tap_service = vo.dal.TAPService(STSCI_VO_URL)
     tap_results = tap_service.search(f"""select observationID from dbo.CaomObservation where
-                                     collection='JWST' and maxLevel=2 and prpID='{int(proposal)}'""")
+                                     collection='JWST' and prpID='{int(proposal)}'""", maxrec=100000)
```
york-stsci, line 1960: Pinging this as second instance of maxrec

york-stsci (commented):
Given that I've seen it running, this looks like a good set of code (modulo minor comments above). Assuming that the preview image generator succeeds as modified, I think we're pretty much good.

york-stsci (commented):
These changes look good. Remainder of review pending.

jwql/jwql_monitors/generate_preview_images.py
bhilbert4: Weird. I don't remember why I added file_type here. It's not used in the function at all. I'll remove it and update the calls to check_existence.

jwql/utils/utils.py
bhilbert4: This came from @mfixstsci 's early work on this. If I recall correctly, there was an issue with not providing the argument? And so he chose a value large enough that we would never have a query that returned more than the limit? @mfixstsci does that sound right?

jwql/website/apps/jwql/static/js/jwql.js
bhilbert4: No options other than 'stage_2' and 'stage_3'. I was just being explicit for the sake of readability.

jwql/utils/preview_image.py
```diff
@@ -670,6 +695,166 @@ def make_image(self, max_img_size=8.0, create_thumbnail=False):
...
+
+        Returns
+        -------
+        intgration_range : range
+            Range object of integration steps
+        """
+        if nint <= 5:
+            integration_range = range(0, nint)
+        else:
+            integration_range = range(0, nint, round(nint / 10))
```
york-stsci: Is it worth doing a linspace here? Like

`max_preview_integrations = 25`
`integration_range = np.linspace(0, nint-1, num=min(max_preview_integrations, nint), dtype=int)`

Note that with that you could also just drop the if/else, because no matter how many you have you will get either every integration or an even spacing of the maximum number.

bhilbert4: Sure that works. It's not quite as pleasing because when the number of integrations is only slightly higher than the max number of ints, it doesn't produce regular spacing between each selected integration. e.g. 

max_preview_integrations = 10
nint = 16 
produces: array([ 0,  1,  3,  5,  6,  8, 10, 11, 13, 15])

but that's a minor quibble.

york-stsci: Yeah. Linspace produces the closest you can do to an even spacing, but integers don't evenly space in circumstances like that, and we can't just do fractional integrations.

bhilbert4: The code as it is above produces even spacing. It achieves that by not having a strict max number of entries.

bhilbert4: Ok, updated to use linspace.

york-stsci: One last question (and it's totally okay if the answer is "we don't care") – do you want to put an upper bound on the number of integrations, like 100 or something?

york-stsci (approved):
I have one question that I asked, but it's absolutely not a thing where either answer should delay the merge. Code looks good, let's get this in.

jwql/utils/preview_image.py
bhilbert4: I think that producing preview images across the entire length of the integrations makes sense. For exposures with very large numbers of integrations (time series data), being able to get a quick look at the data across the entire exposure time is a good thing, since the first chunk of the exposure can be taken prior to any occultation. In that case, you'd have to get fairly far into the exposure before you reached integrations where anything is changing. So I think setting a max number of preview images (I set it to 10 in the code) makes sense, but those should be spread over however many integrations there are. (I've seen TSO data with up to something like 35000 integrations.)

york-stsci (approved):

york-stsci (approved):

PR #1772, opened 2026-07-27
Fix bias monitor issues related to new jwst pipeline version

bhilbert4:
This PR fixes a couple of issues with the bias monitor first seen after switching to version >2.0 of the jwst pipeline.

1) It fixes the extraction of the 0th group into its own file (which is then sent through the pipeline). It seems that the newer pipeline versions require the presence of the asdf extension, which wasn't previously present.

2) When the bias monitor calls the pipeline, the only steps that actually need to run are dq_init and reference pixel correction. However, at the moment all the steps through ramp fitting are run. This PR uses a parameter dictionary to turn off the unnecessary steps, which will save compute time.

### DIFF
--- a/jwql/instrument_monitors/common_monitors/bias_monitor.py
+++ b/jwql/instrument_monitors/common_monitors/bias_monitor.py
@@ -377,9 +377,8 @@ class Bias():
         logging.info("Creating calibration tasks")
 
         # Run only required pipeline steps
-        steps = {'dq_init': {'skip': False},
-                 'refpix': {'skip': False,
-                            'save_results': True},
+        steps = {'dq_init': {},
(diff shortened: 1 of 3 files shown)


bhilbert4:
Tested on the test server and it seems to be working. The pipeline runs complete successfully, and the correct steps are run. I added in the names of steps that are new since the bias monitor was initially created. I also updated the list of apertures that the monitor runs over, excluding a long list of apertures that exist only to support different reference locations, but are actually full frame. There will never be darks that are taken using these apertures, so the monitor was wasting time by looping over them and querying MAST for darks.
@mfixstsci @york-stsci this is ready for review.

york-stsci (approved):

PR #1776, opened 2026-07-29
Fix Bias Monitor Page

mfixstsci:
This adds lines to `lines_to_remove` variable that strips extra styling from our bokeh figures. @bhilbert4 found that the bad pixel monitor figures were stripping the same lines so I am adding it here to strip the lines out the bias monitor plots as well.


bhilbert4 (approved):
Hurray!

PR #1794, opened 2026-09-02
Fix links to multiple instruments to use full program number

bhilbert4:
Currently, when you use the query box on the front page and query for a program number, if that program uses multiple instruments, you get back a text box with a list of instruments, where each instrument name is a link to the first observation for that instrument. However, these links currently only work for 4 digit program numbers. The initial number is stripped from the string. This is a relic from before we reached program numbers > 9999. This PR fixes that by converting the program string (which may contain a leading zero) to an integer before using it in the construction of the link address.


bhilbert4:
Not yet tested.

bhilbert4:
Tested successfully on the test server. This is ready for review. @mfixstsci @york-stsci

york-stsci (approved):
Short and sweet. Looks good.

PR #1796, opened 2026-09-04
Bug leading to missing rate prev images

### LINKED ISSUE #1795: Preview image generator missing rate files
_Issue [JWQL-322](https://jira.stsci.edu/browse/JWQL-322) was created on JIRA by [Bryan Hilbert](https://jira.stsci.edu/secure/ViewProfile.jspa?name=hilbert):_

Reecently, the preview image generator has had several instances where, for a given rootname, it has found and created preview images for all suffixes except rate.fits. Usually, re-running will then allow the rate files to be found. It's not clear whether this is because the rate files are really not present in the directory (seems unlikely) or the rate files are somehow being removed from the list of files to process. We need to figure out what exactly is going on and try to stop this from happening.

bhilbert4:
Resolves #1795 

The preview image generator was skipping over rate files in cases where it had already made preview images for the rateints version of the file. This was due to a tweak to the glob search string used to check for existing preview image files when modifying the code to support level 3 files. This PR corrects that mistake by sorting the list of files such that rate files (and cal files) are done first, before the rateints and calints files.

It also introduces a new keyword such that users can specify which suffixes to create preview images for. This will greatly speed up recovery from this bug since it won't need to create new preview images for all suffixes.


york-stsci (approved):
Looks good, given that it's been run on test.

PR #1797, opened 2026-09-14
Set maxtasksperchild to 1 for prev image generator

bhilbert4:
In an attempt to stop the silent failures of the preview image generator, set maxtasksperchild to one when calling multiprocessing. This means that a given worker will create the preview images for a single program, and then it will be removed and replaced with a new worker for the next program. This should help prevent what may be a memory leak from silently stopping workers one at a time until there are none left.


mfixstsci (approved):
Looks good to me 👍

